PR #36, opened 2016-07-21
Added unlink profile to the methods exported to Objective-C

sebacancinos:


Auth0/Management/ObjectiveC/_ObjectiveManagementAPI.swift
```diff
@@ -68,6 +68,24 @@ public class _ObjectiveManagementAPI: NSObject {
                 }
         }
     }
+    
+    @objc(unlinkUserWithIdentifier:withProvider:fromUserId:callback:)
```
hzalaz: @sebacancinos let's rename the objc method to `unlinkIdentityWithIdentifier:provider:fromUserId:callback`

hzalaz:
@sebacancinos also change the value [here](https://github.com/inaka/Auth0.swift/blob/9ad492d865a433576e567bd8f2df2f9a6bd47ff8/script/.env#L3) from `0.16.2` to `0.17.2`

sebacancinos:
updated @hzalaz

PR #62, opened 2016-11-29
Support id_token response type

cocojoe:
Chaining method for response types
Auto set implicit grant
Credentials support for id_token only
Chaining method for nonce
Nonce required for id_token response type
Tests added

### DIFF
--- a/Auth0/Credentials.swift
+++ b/Auth0/Credentials.swift
@@ -33,14 +33,14 @@ public class Credentials: NSObject, JSONObjectPayload {
     public let idToken: String?
     public let refreshToken: String?
 
-    required public init(accessToken: String? = nil, tokenType: String? = nil, idToken: String? = nil, refreshToken: String? = nil) {
+    init(accessToken: String? = nil, tokenType: String? = nil, idToken: String? = nil, refreshToken: String? = nil) {
         self.accessToken = accessToken
         self.tokenType = tokenType
         self.idToken = idToken
         self.refreshToken = refreshToken
     }
 
-    convenience required public init?(json: [String: Any]) {
+    convenience required public init(json: [String: Any]) {
         self.init(accessToken: json["access_token"] as? String, tokenType: json["token_type"] as? String, idToken: json["id_token"] as? String, refreshToken: json["refresh_token"] as? String)
     }
 
--- a/Auth0/Credentials.swift
+++ b/Auth0/Credentials.swift
@@ -32,10 +32,6 @@ public class Credentials: NSObject, JSONObjectPayload {
     public let tokenType: String?
     public let idToken: String?
     public let refreshToken: String?
-    public var hasToken: Bool {
-        if self.idToken != nil || ( self.accessToken != nil && self.tokenType != nil) { return true }
-        return false
-    }
 
     required public init(accessToken: String? = nil, tokenType: String? = nil, idToken: String? = nil, refreshToken: String? = nil) {
         self.accessToken = accessToken
--- a/Auth0/OAuth2Grant.swift
+++ b/Auth0/OAuth2Grant.swift
@@ -34,18 +34,12 @@ struct ImplicitGrant: OAuth2Grant {
     let defaults: [String : String]
     let response: [ResponseOptions]
 
-    // Expected default behaviour in Unit Tests
-    init() {
-        self.init(response: [.token])
-    }
-
-    init(response: [ResponseOptions], nonce: String? = nil) {
+    init(response: [ResponseOptions] = [.token], nonce: String? = nil) {
         self.response = response
         if let nonce = nonce {
-            self.defaults = [ "response_type" : self.response.map { $0.label! }.joined(separator: " "),
-                              "nonce" : nonce]
+            self.defaults = [ "nonce" : nonce]
         } else {
-            self.defaults = [ "response_type" : self.response.map { $0.label! }.joined(separator: " ") ]
+            self.defaults = [:]
         }
     }
 
--- a/Auth0/OAuth2Grant.swift
+++ b/Auth0/OAuth2Grant.swift
@@ -32,10 +32,10 @@ protocol OAuth2Grant {
 struct ImplicitGrant: OAuth2Grant {
 
     let defaults: [String : String]
-    let response: [ResponseOptions]
+    let responseType: [ResponseOptions]
 
-    init(response: [ResponseOptions] = [.token], nonce: String? = nil) {
-        self.response = response
+    init(responseType: [ResponseOptions] = [.token], nonce: String? = nil) {
+        self.responseType = responseType
         if let nonce = nonce {
             self.defaults = ["nonce" : nonce]
         } else {
@@ -45,7 +45,7 @@ struct ImplicitGrant: OAuth2Grant {
 
     func credentials(from values: [String : String], callback: @escaping (Result<Credentials>) -> ()) {
         // id token reponse expectations and JWT validation, nonce validation
-        if response.contains(.id_token) {
+        if responseType.contains(.id_token) {
             guard let id_token = values["id_token"], let nonce = self.defaults["nonce"],
             validate(token: id_token, nonce: nonce) else {
                 return callback(.failure(error: WebAuthError.idTokenValidationFailed))
@@ -53,7 +53,7 @@ struct ImplicitGrant: OAuth2Grant {
         }
 
         // token response expectations
-        if response.contains(.token) {
+        if responseType.contains(.token) {
             guard values["access_token"] != nil && values["token_type"] != nil else {
                 return callback(.failure(error: WebAuthError.tokenValidationFailed))
             }
@@ -126,14 +126,7 @@ struct PKCE: OAuth2Grant {
 }
 
 private func validate(token: String, nonce: String) -> Bool {
-    do {
-        let jwt = try decode(jwt: token)
-        guard let nonceClaim = jwt.claim(name: "nonce").string,
-            nonceClaim == nonce else {
-                return false
-        }
-    } catch {
-        return false
-    }
-    return true
+    guard let jwt = try? decode(jwt: token) else { return false }
+    let tokenNonce = jwt.claim(name: "nonce").string
+    return tokenNonce == nonce
 }
--- a/Auth0/OAuth2Grant.swift
+++ b/Auth0/OAuth2Grant.swift
@@ -33,7 +33,7 @@ struct ImplicitGrant: OAuth2Grant {
     let defaults: [String : String] = ["response_type": "token"]
 
     func credentials(from values: [String : String], callback: @escaping (Result<Credentials>) -> ()) {
-        guard let credentials = Credentials(json: values as [String : Any]) else {
+        guard let credentials = Credentials(json: values as [String : Any]), credentials.hasToken else {
             let data = try! JSONSerialization.data(withJSONObject: values, options: [])
             let string = String(data: data, encoding: .utf8)
             callback(.failure(error: AuthenticationError(string: string)))
--- a/Auth0/OAuth2Grant.swift
+++ b/Auth0/OAuth2Grant.swift
@@ -37,7 +37,7 @@ struct ImplicitGrant: OAuth2Grant {
     init(response: [ResponseOptions] = [.token], nonce: String? = nil) {
         self.response = response
         if let nonce = nonce {
-            self.defaults = [ "nonce" : nonce]
+            self.defaults = ["nonce" : nonce]
         } else {
             self.defaults = [:]
         }
@@ -46,25 +46,16 @@ struct ImplicitGrant: OAuth2Grant {
     func credentials(from values: [String : String], callback: @escaping (Result<Credentials>) -> ()) {
         // id token reponse expectations and JWT validation, nonce validation
         if response.contains(.id_token) {
-            guard let id_token = values["id_token"] else {
-                return callback(.failure(error: ResponseError.idTokenMissing))
-            }
-
-            do {
-                let jwt = try decode(jwt: id_token)
-                guard let nonceClaim = jwt.claim(name: "nonce").string, let nonce = self.defaults["nonce"],
-                    nonceClaim == nonce else {
-                        return callback(.failure(error: ResponseError.nonceDoesNotMatch))
-                }
-            } catch {
-                return callback(.failure(error: ResponseError.tokenDecodeFailed))
+            guard let id_token = values["id_token"], let nonce = self.defaults["nonce"],
+            validate(token: id_token, nonce: nonce) else {
+                return callback(.failure(error: WebAuthError.idTokenValidationFailed))
             }
         }
 
         // token response expectations
         if response.contains(.token) {
             guard values["access_token"] != nil && values["token_type"] != nil else {
-                return callback(.failure(error: ResponseError.tokenIssue))
+                return callback(.failure(error: WebAuthError.tokenValidationFailed))
             }
         }
 
@@ -132,4 +123,17 @@ struct PKCE: OAuth2Grant {
         components.a0_queryValues.forEach { items[$0] = $1 }
         return items
     }
+}
+
+private func validate(token: String, nonce: String) -> Bool {
+    do {
+        let jwt = try decode(jwt: token)
+        guard let nonceClaim = jwt.claim(name: "nonce").string,
+            nonceClaim == nonce else {
+                return false
+        }
+    } catch {
+        return false
+    }
+    return true
 }
--- a/Auth0/OAuth2Grant.swift
+++ b/Auth0/OAuth2Grant.swift
@@ -32,9 +32,9 @@ protocol OAuth2Grant {
 struct ImplicitGrant: OAuth2Grant {
 
     let defaults: [String : String]
-    let responseType: [ResponseOptions]
+    let responseType: [ResponseType]
 
-    init(responseType: [ResponseOptions] = [.token], nonce: String? = nil) {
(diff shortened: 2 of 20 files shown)


Auth0/ResponseType.swift
```diff
@@ -0,0 +1,29 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import Foundation
+
+public enum ResponseType: String {
```
hzalaz: The OptionType is a best choice for this

Auth0/SafariWebAuth.swift
```diff
@@ -76,6 +76,20 @@ class SafariWebAuth: WebAuth {
         return self
     }
 
+    func responseType(_ responseType: [ResponseType]) -> Self {
+        if !responseType.contains(.code) {
+            _ = self.usingImplicitGrant()
```
hzalaz: The default was code grant and it should still be now

Auth0/SafariWebAuth.swift
```diff
@@ -76,6 +76,20 @@ class SafariWebAuth: WebAuth {
...
+        }
+        let response = responseType.reduce(String()) { $0 + $1.rawValue + " "}
```
hzalaz: Try to use literals whenever possible not `String()`

hzalaz: Also you could use 

```swift
responseType.joined(separator: " ")
```

and avoid the `dropLast()`

Auth0/SafariWebAuth.swift
```diff
@@ -88,6 +102,11 @@ class SafariWebAuth: WebAuth {
             else {
                 return callback(Result.failure(error: WebAuthError.noBundleIdentifierFound))
         }
+        if let response = self.parameters["response_type"], response.contains(ResponseType.id_token.rawValue) {
```
hzalaz: `response_type` is mandatory and we should combine the if and guard in a single guard, it serves no purpose having them separated

Auth0/WebAuthError.swift
```diff
@@ -63,6 +65,11 @@ public enum WebAuthError: CustomNSError {
                 NSLocalizedDescriptionKey: message,
                 WebAuthError.infoKey: self
             ]
+        case .noNonceProvided:
+            return [
+                NSLocalizedDescriptionKey: "You must provide a nonce value for the id_token response type",
```
hzalaz: `A nonce value must be supplied when response_type includes id_token in order to prevent replay attacks`

Auth0/Credentials.swift
```diff
@@ -28,24 +28,24 @@ import Foundation
 @objc(A0Credentials)
 public class Credentials: NSObject, JSONObjectPayload {
 
-    public let accessToken: String
-    public let tokenType: String
+    public let accessToken: String?
+    public let tokenType: String?
     public let idToken: String?
     public let refreshToken: String?
+    public var hasToken: Bool {
```
hzalaz: Don't see the why we need this method, the `hasToken` can only be known to the one who starts the auth or if we pass along the `response_type` and validate. Thing we should do before yielding the credentials to the caller

Auth0/Credentials.swift
```diff
@@ -28,24 +28,24 @@ import Foundation
...
+        if self.idToken != nil || ( self.accessToken != nil && self.tokenType != nil) { return true }
+        return false
+    }
 
-    required public init(accessToken: String, tokenType: String, idToken: String? = nil, refreshToken: String? = nil) {
+    required public init(accessToken: String? = nil, tokenType: String? = nil, idToken: String? = nil, refreshToken: String? = nil) {
```
hzalaz: This is sort of a breaking change so I'd need to figure out a way to communicate this, not a big one but still a breaking one

Auth0/OAuth2Grant.swift
```diff
@@ -33,7 +33,7 @@ struct ImplicitGrant: OAuth2Grant {
     let defaults: [String : String] = ["response_type": "token"]
 
     func credentials(from values: [String : String], callback: @escaping (Result<Credentials>) -> ()) {
-        guard let credentials = Credentials(json: values as [String : Any]) else {
+        guard let credentials = Credentials(json: values as [String : Any]), credentials.hasToken else {
```
hzalaz: We need to validate that was requested was returned otherwise we need to fail.

Auth0/SafariWebAuth.swift
```diff
@@ -76,6 +76,20 @@ class SafariWebAuth: WebAuth {
         return self
     }
 
+    func responseType(_ responseType: [ResponseType]) -> Self {
```
hzalaz: We need to enforce that a response_type is picked here and by default should include `code`

Auth0/SafariWebAuth.swift
cocojoe: It's not as easy as that due to it being an enum and requiring rawValue, perhaps with the suggestion of using `OptionSet` it will flow more easily. I'll look into it.

cocojoe: 👍

Auth0/OAuth2Grant.swift
```diff
@@ -30,9 +31,34 @@ protocol OAuth2Grant {
...
+        if let nonce = nonce {
+            self.defaults = ["nonce" : nonce]
+        } else {
+            self.defaults = [:]
+        }
+    }
 
     func credentials(from values: [String : String], callback: @escaping (Result<Credentials>) -> ()) {
+        // id token reponse expectations and JWT validation, nonce validation
+        if response.contains(.id_token) {
```
hzalaz: why the if-guard? just use `guard`

Auth0/OAuth2Grant.swift
```diff
@@ -30,9 +31,34 @@ protocol OAuth2Grant {
...
+            guard let id_token = values["id_token"], let nonce = self.defaults["nonce"],
+            validate(token: id_token, nonce: nonce) else {
+                return callback(.failure(error: WebAuthError.idTokenValidationFailed))
+            }
+        }
+
+        // token response expectations
+        if response.contains(.token) {
```
hzalaz: Same as above

Auth0/OAuth2Grant.swift
```diff
@@ -97,3 +124,16 @@ struct PKCE: OAuth2Grant {
         return items
     }
 }
+
+private func validate(token: String, nonce: String) -> Bool {
+    do {
```
hzalaz:
```swift
guard let jwt = try? decode(jwt: token), let nonce = jwt.claim(name: "nonce") else { return false }
return nonce.string == nonce
```

If we are not handling the error is useless to catch it

Auth0/SafariWebAuth.swift
```diff
@@ -76,18 +78,31 @@ class SafariWebAuth: WebAuth {
         return self
     }
 
-    func usingImplicitGrant() -> Self {
-        self.usePKCE = false
+    func response(_ response: [ResponseOptions]) -> Self {
+        if !response.contains(.code) { self.usePKCE = false }
```
hzalaz: do we need the if?

Auth0/SafariWebAuth.swift
```diff
@@ -76,18 +78,31 @@ class SafariWebAuth: WebAuth {
...
+    }
+
     func start(_ callback: @escaping (Result<Credentials>) -> ()) {
         guard
             let redirectURL = self.redirectURL
             , !redirectURL.absoluteString.hasPrefix(SafariWebAuth.NoBundleIdentifier)
             else {
                 return callback(Result.failure(error: WebAuthError.noBundleIdentifierFound))
         }
+        if self.response.contains(.id_token) {
```
hzalaz: if-guard again, it's better to have just guard

Auth0/ResponseOptions.swift
```diff
@@ -0,0 +1,43 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import Foundation
+
+public struct ResponseOptions: OptionSet {
+    public let rawValue: Int
+    public let label: String?
```
hzalaz: I assume this is part of the hack to use label right?

hzalaz: If we use an enum? it may be better, we might need to provide a computed property name but it could be less hackish

Auth0/WebAuth.swift
```diff
@@ -138,11 +138,25 @@ public protocol WebAuth: Trackable, Loggable {
      */
     func parameters(_ parameters: [String: String]) -> Self
 
+    /// Setup the response types to be used for authentcation
+    ///
+    /// - Parameter response: Array of ResponseOptions
+    /// - Returns: the same OAuth2 instance to allow method chaining
```
hzalaz: `OAuth2` -> `WebAuth`

Auth0/WebAuth.swift
```diff
@@ -138,11 +138,25 @@ public protocol WebAuth: Trackable, Loggable {
...
+    func response(_ response: [ResponseOptions]) -> Self
```
hzalaz: Let's stick with `responseType`

Auth0/WebAuthError.swift
```diff
@@ -29,12 +29,17 @@ import Foundation
  - CannotDismissWebAuthController: When trying to dismiss WebAuth controller, no presenter controller could be found.
  - UserCancelled:                  User cancelled the web-based authentication, e.g. tapped the "Done" button in SFSafariViewController
  - PKCENotAllowed:                 PKCE for the supplied Auth0 ClientId was not allowed. You need to set the `Token Endpoint Authentication Method` to `None` in your Auth0 Dashboard
+ - noNonceProvided:                A nonce value must be provided to use the response option of id_token
```
hzalaz: missing the last errors here

Auth0Tests/CredentialsSpec.swift
```diff
@@ -34,15 +34,15 @@ class CredentialsSpec: QuickSpec {
...
-            it("should return nil when missing access_token") {
-                let credentials = Credentials(json: ["token_type": Bearer])
-                expect(credentials).to(beNil())
-            }
-
-            it("should return nil when missing toke_type") {
-                let credentials = Credentials(json: ["access_token": AccessToken])
-                expect(credentials).to(beNil())
-            }
+//            it("should return nil when missing access_token") {
```
hzalaz: what about these?

Auth0/SafariWebAuth.swift
cocojoe: We can probably retire self.userPKCE and replace with a check to see if response contains .code

Auth0/ResponseOptions.swift
cocojoe: We can do this anyway, originally I had this but removed it for some reason.  However in retrospect I think it's fine.

```
var label: String? {
        switch self.rawValue {
        case ResponseOptions.token.rawValue:
            return "token"
        case ResponseOptions.id_token.rawValue:
            return "id_token"
        case ResponseOptions.code.rawValue:
            return "code"
        default:
            return nil
        }
    }
```

Auth0/WebAuth.swift
cocojoe: All existing documentation in this file references 'OAuth2', I'll rename them all then.

Auth0/OAuth2Grant.swift
cocojoe: jwt.claim is non optional,  agree it can be streamlined.

Auth0/OAuth2Grant.swift
cocojoe: Multiple *responseType* are processed in this `func`, `guard` would exit if the *responseType* was not `.id_token`, even `.id_token` would then fail in the next proposed guard as it wouldn't contain `.token`
This applies to all my usage of `if` then `guard`.

Auth0/OAuth2Grant.swift
```diff
@@ -30,9 +31,34 @@ protocol OAuth2Grant {
...
+        if responseType.contains(.id_token) {
+            guard let id_token = values["id_token"], let nonce = self.defaults["nonce"],
+            validate(token: id_token, nonce: nonce) else {
+                return callback(.failure(error: WebAuthError.idTokenValidationFailed))
+            }
+        }
+
+        // token response expectations
+        if responseType.contains(.token) {
+            guard values["access_token"] != nil && values["token_type"] != nil else {
```
hzalaz:
```swift
guard !responseType.contains(.token) || values["access_token"] != nil else {...}
```

hzalaz: No need to account for tokenType

Auth0/OAuth2Grant.swift
```diff
@@ -30,9 +31,34 @@ protocol OAuth2Grant {
...
+        if let nonce = nonce {
+            self.defaults = ["nonce" : nonce]
+        } else {
+            self.defaults = [:]
+        }
+    }
 
     func credentials(from values: [String : String], callback: @escaping (Result<Credentials>) -> ()) {
+        // id token reponse expectations and JWT validation, nonce validation
+        if responseType.contains(.id_token) {
```
hzalaz: Move all logic of nonce validation to he validate method and then you can join guard and if into a guard.

Auth0/Credentials.swift
```diff
@@ -28,24 +28,20 @@ import Foundation
...
         self.accessToken = accessToken
         self.tokenType = tokenType
         self.idToken = idToken
         self.refreshToken = refreshToken
     }
 
     convenience required public init?(json: [String: Any]) {
```
hzalaz: We are no longer returning nil so the init is not required to be a failable one

Auth0/ResponseOptions.swift
```diff
@@ -0,0 +1,50 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import Foundation
+
+public struct ResponseOptions: OptionSet {
```
hzalaz: Let's make it an enum so we can avoid the dev adding strange values as response types

hzalaz: Also `ResponseType` might be more suitable as a name (plus missing docs)

Auth0/ResponseOptions.swift
```diff
@@ -0,0 +1,50 @@
...
+    public let rawValue: Int
+
+    public init(rawValue: Int) {
+        self.rawValue = rawValue
+    }
+
+    public static let token     = ResponseOptions(rawValue: 1 << 0)
+    public static let id_token  = ResponseOptions(rawValue: 1 << 1)
```
hzalaz: Let's keep with camelCase

Auth0/WebAuthError.swift
```diff
@@ -29,12 +29,19 @@ import Foundation
...
  */
 public enum WebAuthError: CustomNSError {
     case noBundleIdentifierFound
     case cannotDismissWebAuthController
     case userCancelled
     case pkceNotAllowed(String)
+    case noNonceProvided
+    case missingResponseParam(String)
+    case idTokenValidationFailed
+    case tokenValidationFailed
```
hzalaz: `missingAccessToken`

Auth0/WebAuthError.swift
```diff
@@ -29,12 +29,19 @@ import Foundation
...
+ - tokenValidationFailed:          Response did not return the necessary token params
  */
 public enum WebAuthError: CustomNSError {
     case noBundleIdentifierFound
     case cannotDismissWebAuthController
     case userCancelled
     case pkceNotAllowed(String)
+    case noNonceProvided
+    case missingResponseParam(String)
+    case idTokenValidationFailed
```
hzalaz: `invalidIdTokenNonce`

Auth0/ResponseOptions.swift
cocojoe: So after originally being an enum and suggeting to change it to OptionSet, you want to go back to enum again?

hzalaz: forgot that enum cannot conform to OptionType protocol

Auth0/WebAuthError.swift
cocojoe: These were added.

PR #65, opened 2016-12-01
Multiple respone_type support

cocojoe:


Auth0/OAuth2Grant.swift
```diff
@@ -67,20 +67,28 @@ struct PKCE: OAuth2Grant {
...
+        self.verifier = verifier
+        self.responseType = responseType
+
+        var newDefaults: [String: String] = [
             "code_challenge": challenge,
             "code_challenge_method": method,
         ]
-        self.verifier = verifier
+
+        if let nonce = nonce {
```
hzalaz: I think if you set nil value the dict will remove the entry so the if might not be necessary

PR #67, opened 2016-12-02
Support for audience parameter for WebAuth

cocojoe:
Added tests


Auth0/WebAuth.swift
```diff
@@ -151,6 +151,12 @@ public protocol WebAuth: Trackable, Loggable {
     /// - Returns: the same WebAuth instance to allow method chaining
     func nonce(_ nonce: String) -> Self
 
+    /// The application that is the intended target of the token
```
hzalaz:
```
///  Audience name of the API that your application will call using the `access_token` returned after Auth.
///  This value must match the one defined in Auth0 Dashboard [APIs Section](https://manage.auth0.com/#/apis)
```

PR #72, opened 2016-12-13
expires_in supported in Credentials

cocojoe:
Tests added

### DIFF
--- a/Auth0/Credentials.swift
+++ b/Auth0/Credentials.swift
@@ -32,16 +32,22 @@ public class Credentials: NSObject, JSONObjectPayload {
     public let tokenType: String?
     public let idToken: String?
     public let refreshToken: String?
+    public let expiresIn: Date?
 
-    init(accessToken: String? = nil, tokenType: String? = nil, idToken: String? = nil, refreshToken: String? = nil) {
+    init(accessToken: String? = nil, tokenType: String? = nil, idToken: String? = nil, refreshToken: String? = nil, expiresIn: Date? = nil) {
         self.accessToken = accessToken
         self.tokenType = tokenType
         self.idToken = idToken
         self.refreshToken = refreshToken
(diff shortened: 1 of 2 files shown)


Auth0Tests/CredentialsSpec.swift
```diff
@@ -61,6 +64,30 @@ class CredentialsSpec: QuickSpec {
                 expect(credentials.refreshToken) == RefreshToken
             }
 
+            context("expires_in responses") {
+
+                it("should have valid exiresIn from string") {
+                    let credentials = Credentials(json: ["expires_in": String(expiresIn)])
```
hzalaz: use string interpolation for the value

Auth0/Credentials.swift
```diff
@@ -32,16 +32,22 @@ public class Credentials: NSObject, JSONObjectPayload {
...
         self.accessToken = accessToken
         self.tokenType = tokenType
         self.idToken = idToken
         self.refreshToken = refreshToken
+        self.expiresIn = expiresIn
     }
 
     convenience required public init(json: [String: Any]) {
-        self.init(accessToken: json["access_token"] as? String, tokenType: json["token_type"] as? String, idToken: json["id_token"] as? String, refreshToken: json["refresh_token"] as? String)
+        var expiresIn: Date? = nil
```
hzalaz: Something like this will be a bit clearer and swiftier
```swift
let expiresIn: Date?
switch json["expires_in"] {
case let string as String:
    guard let double = Double(string) else { break }
    expiresIn = Date(timeIntervalSince1970: double)
case let int as Int:
    expiresIn = Date(timeIntervalSince1970: Double(int))
case let double as Double:
    expiresIn = Date(timeIntervalSince1970: double)
default:
    expiresIn = nil
}
```

cocojoe: 👍

PR #86, opened 2017-01-30
Native Authentication support

cocojoe:
Adds support for native authentication with a third party IdP using their iOS SDK and then performing the Auth against Auth0.

### DIFF
--- a/Auth0.podspec
+++ b/Auth0.podspec
@@ -21,9 +21,9 @@ Pod::Spec.new do |s|
   s.ios.source_files = 'Auth0/*.{swift,h,m}'
   s.ios.frameworks = 'UIKit', 'SafariServices'
   s.osx.source_files = 'Auth0/*.swift'
-  s.osx.exclude_files = 'Auth0/_ObjectiveWebAuth.swift', 'Auth0/ControllerModalPresenter.swift', 'Auth0/OAuth2Grant.swift', 'Auth0/OAuth2Session.swift', 'Auth0/SessionStorage.swift', 'Auth0/WebAuth.swift', 'Auth0/WebAuthError.swift', 'Auth0/SafariWebAuth.swift'
+  s.osx.exclude_files = 'Auth0/_ObjectiveWebAuth.swift', 'Auth0/ControllerModalPresenter.swift', 'Auth0/OAuth2Grant.swift', 'Auth0/AuthTransaction.swift', 'Auth0/TransactionStore.swift', 'Auth0/WebAuth.swift', 'Auth0/WebAuthError.swift', 'Auth0/SafariWebAuth.swift'
   s.watchos.source_files = 'Auth0/*.swift'
-  s.watchos.exclude_files = 'Auth0/_ObjectiveWebAuth.swift', 'Auth0/ControllerModalPresenter.swift', 'Auth0/OAuth2Grant.swift', 'Auth0/OAuth2Session.swift', 'Auth0/SessionStorage.swift', 'Auth0/WebAuth.swift', 'Auth0/WebAuthError.swift', 'Auth0/SafariWebAuth.swift'
+  s.watchos.exclude_files = 'Auth0/_ObjectiveWebAuth.swift', 'Auth0/ControllerModalPresenter.swift', 'Auth0/OAuth2Grant.swift', 'Auth0/AuthTransaction.swift', 'Auth0/TransactionStore.swift', 'Auth0/WebAuth.swift', 'Auth0/WebAuthError.swift', 'Auth0/SafariWebAuth.swift'
   s.tvos.source_files = 'Auth0/*.swift'
-  s.tvos.exclude_files = 'Auth0/_ObjectiveWebAuth.swift', 'Auth0/ControllerModalPresenter.swift', 'Auth0/OAuth2Grant.swift', 'Auth0/OAuth2Session.swift', 'Auth0/SessionStorage.swift', 'Auth0/WebAuth.swift', 'Auth0/WebAuthError.swift', 'Auth0/SafariWebAuth.swift'
+  s.tvos.exclude_files = 'Auth0/_ObjectiveWebAuth.swift', 'Auth0/ControllerModalPresenter.swift', 'Auth0/OAuth2Grant.swift', 'Auth0/AuthTransaction.swift', 'Auth0/TransactionStore.swift', 'Auth0/WebAuth.swift', 'Auth0/WebAuthError.swift', 'Auth0/SafariWebAuth.swift'
 end
--- a/Auth0.podspec
+++ b/Auth0.podspec
@@ -21,9 +21,9 @@ Pod::Spec.new do |s|
   s.ios.source_files = 'Auth0/*.{swift,h,m}'
   s.ios.frameworks = 'UIKit', 'SafariServices'
   s.osx.source_files = 'Auth0/*.swift'
-  s.osx.exclude_files = 'Auth0/_ObjectiveWebAuth.swift', 'Auth0/ControllerModalPresenter.swift', 'Auth0/OAuth2Grant.swift', 'Auth0/AuthTransaction.swift', 'Auth0/TransactionStore.swift', 'Auth0/WebAuth.swift', 'Auth0/WebAuthError.swift', 'Auth0/SafariWebAuth.swift'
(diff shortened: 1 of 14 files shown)


Auth0.podspec
```diff
@@ -21,9 +21,9 @@ Pod::Spec.new do |s|
   s.ios.source_files = 'Auth0/*.{swift,h,m}'
   s.ios.frameworks = 'UIKit', 'SafariServices'
   s.osx.source_files = 'Auth0/*.swift'
-  s.osx.exclude_files = 'Auth0/_ObjectiveWebAuth.swift', 'Auth0/ControllerModalPresenter.swift', 'Auth0/OAuth2Grant.swift', 'Auth0/OAuth2Session.swift', 'Auth0/SessionStorage.swift', 'Auth0/WebAuth.swift', 'Auth0/WebAuthError.swift', 'Auth0/SafariWebAuth.swift'
+  s.osx.exclude_files = 'Auth0/_ObjectiveWebAuth.swift', 'Auth0/ControllerModalPresenter.swift', 'Auth0/OAuth2Grant.swift', 'Auth0/AuthTransaction.swift', 'Auth0/TransactionStore.swift', 'Auth0/WebAuth.swift', 'Auth0/WebAuthError.swift', 'Auth0/SafariWebAuth.swift'
```
hzalaz: Shouldn't we exclude native auth file?

Auth0/AuthTransaction.swift
```diff
@@ -0,0 +1,43 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import UIKit
+import SafariServices
+
+/**
+ Represents an on going AuthTransaction with Auth0.
```
hzalaz: `Represents an on going Auth transaction with an Identity Provider (Auth0 or a third party).`

Auth0/AuthTransaction.swift
```diff
@@ -0,0 +1,43 @@
...
+
+ It will handle result from the redirect URL configured when the AuthTransaction flow was started,
```
hzalaz: Since the Auth will be done outside the application control, Safari or third party IdP App, the only way to communicate the results back is using an url with a  registered custom scheme in our application so iOS can open it on success/failure. When that happens iOS will call a method in your `AppDelegate` and there is where we need to handle the result.

<USE SAME CODE SNIPPET>

Auth0/NativeAuth.swift
```diff
@@ -0,0 +1,74 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import UIKit
+
+public protocol AuthProvider {
```
hzalaz: docs?

Auth0/NativeAuth.swift
```diff
@@ -0,0 +1,74 @@
...
+    func login(withConnection connection: String, scope: String, parameters: [String: Any]) -> NativeAuthTransaction
+}
+
+public struct NativeAuthCredentials {
+    let token: String
+    let extras: [String: Any]
+
+    public init(token: String, extras: [String: Any]) {
```
hzalaz: Make public the attributes instead of creating an `init`

Auth0/NativeAuth.swift
```diff
@@ -0,0 +1,74 @@
...
+    var state: String? {
+        return self.connection
+    }
+
+    public func start(callback: @escaping (Result<Credentials>) -> ()) {
+        TransactionStore.sharedInstance.store(self)
+        self.auth { result in
+            switch result {
+            case .success(let nativeAuthCredentials):
+                self.socialAuthentication(nativeAuthCredentials, callback: callback)
```
hzalaz: A bit silly but `credentials` is enough for the name.

hzalaz: Also inline the method, there is no gain on having it, it just makes you perform an extra jump to see where the logic is

Auth0/NativeAuth.swift
```diff
@@ -0,0 +1,74 @@
...
+            case .failure(let error):
+                callback(.failure(error: error))
+            }
+        }
+    }
+
+    private func socialAuthentication(_ nativeAuthCredentials: NativeAuthCredentials, callback: @escaping (Result<Credentials>) -> ()) {
+        var parameters: [String: Any] = self.parameters
+        nativeAuthCredentials.extras.forEach { key, value in parameters[key] = value }
```
hzalaz: a matter of style: there is no need to name the parameters here in the closure

Auth0/NativeAuth.swift
```diff
@@ -0,0 +1,74 @@
...
+
+        authentication().loginSocial(token: nativeAuthCredentials.token, connection: connection, scope: scope, parameters: parameters)
+            .start { callback($0) }
```
hzalaz: Just pass the callback as the parameter why wrapping it up in another one?

Auth0/TransactionStore.swift
```diff
@@ -22,11 +22,11 @@
 
 import UIKit
 
-/// Keeps track of current Auth session (e.g. OAuth2)
-class SessionStorage {
-    static let sharedInstance = SessionStorage()
+/// Keeps track of current Auth Transaction
+class TransactionStore {
+    static let sharedInstance = TransactionStore()
```
hzalaz: Now that we are modifying it let's use the `shared` name instead

Auth0/NativeAuth.swift
```diff
@@ -0,0 +1,74 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import UIKit
+
+public protocol AuthProvider {
+    func login(withConnection connection: String, scope: String, parameters: [String: Any]) -> NativeAuthTransaction
+}
+
+public struct NativeAuthCredentials {
```
hzalaz: docs?

Auth0/NativeAuth.swift
```diff
@@ -0,0 +1,74 @@
...
+public protocol NativeAuthTransaction: AuthTransaction {
+    var scope: String { get }
+    var connection: String { get }
+    var parameters: [String: Any] { get }
+
+    typealias Callback = (Result<NativeAuthCredentials>) -> ()
+    func auth(callback: @escaping Callback) -> ()
+}
+
+public extension NativeAuthTransaction {
```
hzalaz: doc?

Auth0/NativeAuth.swift
```diff
@@ -0,0 +1,74 @@
+// NativeAuth.swift
```
hzalaz: I believe we might need to split the file in parts

Auth0/NativeAuth.swift
```diff
@@ -22,20 +22,54 @@
...
+/**
+ The NativeAuthTransaction expands on the AuthTransaction protocol to add additional functionality
+ required for handling Native Authentication scenarios.
+ 
+ The auth func handles the result from the IdP SDK authentication callback.
+ 
+ Scope, connnection and parameters are used in the final stage of Auth Transaction during the Auth0 Social Authentication.
+ 
+ ```
+ class SocialNativeTransaction: NativeAuthTransaction {
```
hzalaz: I am not buying this example, it does not explain what each method does and what the parameter are for, what are the properties. Also it is not showing any real case usage.

We need to document each protocol property and method explaining what it does

Auth0/NativeAuth.swift
cocojoe: I broke out AuthProvider, everything left is NativeAuth*

Auth0/NativeAuth.swift
cocojoe: Have fleshed this out, let's see how it feels.

Auth0/NativeAuth.swift
cocojoe: Thought I had did this for a reason,  if you want a public structure type to be initializable with a memberwise initializer when used in another module, you must provide a public memberwise initializer yourself.

hzalaz: unless both attributes are public so it should infer the constructor

cocojoe: It will infer for the project but not when imported. [Source](http://stackoverflow.com/questions/26224693/how-can-i-make-public-by-default-the-member-wise-initialiser-for-structs-in-swif/26224873#26224873)

PR #87, opened 2017-02-01
Adding ns coding conformance

DarkSatyr:
In Lock library UserProfile conforms to NSCoding what allow to archive an unarchive profile easily, when transited to Auth0 we found that there is no conformance to NSCoding.
Of course, we can convert to json -> serialize json -> deserialize -> create Profile from json, but we would like to have simple way of serialization/deserialization of Profile objects since we are going to use Auth0 authentication permanently


hzalaz:
@DarkSatyr thanks for the PR but I am not too convinced in coupling the profile to `NSSecureCoding` also we'd make some changes in the response of userinfo so we would need to review it again.

I am a bit inclined to return a raw map and let the user of the lib handle the storage whatever they want

hzalaz:
Closing this one, added the feature request and will consider the best way to provide a support for NSSecureCoding

PR #90, opened 2017-02-07
Added support for resetting parameters in WebAuth

cocojoe:
Fix Ensure parameters are cleared after URL build

### DIFF
--- a/Auth0/SafariWebAuth.swift
+++ b/Auth0/SafariWebAuth.swift
@@ -100,6 +100,11 @@ class SafariWebAuth: WebAuth {
         return self
     }
 
+    func reset() -> Self {
+        self.parameters = [:]
+        return self
+    }
+
     func start(_ callback: @escaping (Result<Credentials>) -> Void) {
         guard
             let redirectURL = self.redirectURL, !redirectURL.absoluteString.hasPrefix(SafariWebAuth.NoBundleIdentifier)
--- a/Auth0/WebAuth.swift
+++ b/Auth0/WebAuth.swift
@@ -200,4 +200,9 @@ public protocol WebAuth: Trackable, Loggable {
      - parameter callback: callback called with the result of the WebAuth flow
      */
     func start(_ callback: @escaping (Result<Credentials>) -> Void)
+
+    /*
+     Reset the WebAuth parameters dictionary.
+     */
+    func reset() -> Self
 }
--- a/Auth0Tests/WebAuthSpec.swift
+++ b/Auth0Tests/WebAuthSpec.swift
@@ -289,6 +289,18 @@ class WebAuthSpec: QuickSpec {
 
         }
 
+        describe("reset") {
+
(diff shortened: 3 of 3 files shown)


Auth0/SafariWebAuth.swift
```diff
@@ -153,6 +158,7 @@ class SafariWebAuth: WebAuth {
             entries["nonce"] = self.nonce
         }
         self.parameters.forEach { entries[$0] = $1 }
+        self.parameters = [:]
```
hzalaz: Is this needed or there is some side effect if we don't clean it?

Auth0/WebAuth.swift
```diff
@@ -118,6 +118,15 @@ public protocol WebAuth: Trackable, Loggable {
     func scope(_ scope: String) -> Self
 
     /**
+      Provider scopes for oauth2/social connections. e.g. Facebook, Google etc
+
+     - parameter connectionScope: oauth2/social scope list: `user_friends email`
```
hzalaz: I think they are comma-separated but I might be wrong

cocojoe: It works with either, I went with spaces as it was consistent with setting scope.

Auth0/SafariWebAuth.swift
cocojoe: In a WebAuth instance I noticed if you cancel a connection then try another connection the parameters will persist.

Auth0/WebAuth.swift
hzalaz: The server expects comma separated so let's honor this

Auth0/SafariWebAuth.swift
hzalaz: This is intended since `WebAuth` should not be reusable in that way since its the same instance so I'd not make this hidden side-effect here. Better to properly handle this in Lock

cocojoe: Will remove and refactor into Lock

Auth0/WebAuth.swift
cocojoe: @hzalaz Can I suggest there should be a method to reset parameters, if you are using a common instance of `WebAuth`, parameters are cumulative with no way to remove them. 
This is most relevant with the requirement of connection specific params like `connectionScopes` and `login_hint`.

You can of course create a new instance each time but it is less expensive to `[:]` than it is init a new object. 

Thoughts?

Auth0/SafariWebAuth.swift
cocojoe: @hzalaz Can I suggest there should be a method to reset parameters, if you are using a common instance of `WebAuth`, parameters are cumulative with no way to remove them. 
This is most relevant with the requirement of connection specific params like `connectionScopes` and `login_hint`.

You can of course create a new instance each time but it is less expensive to `[:]` than it is init a new object. 

Thoughts?

PR #93, opened 2017-02-28
Fixes for SwiftLint 0.16.1

cocojoe:
Added Lint Pass to Default Lane
Added SwiftLint version check in build phase

### DIFF
--- a/Auth0.xcodeproj/project.pbxproj
+++ b/Auth0.xcodeproj/project.pbxproj
@@ -1157,7 +1157,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "if which swiftlint >/dev/null && [[ `swiftlint version` =~ \"0.16\" ]] && [[ \"${CONFIGURATION}\" = \"Debug\" ]]; then\nswiftlint\nfi";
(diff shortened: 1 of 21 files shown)


Auth0.xcodeproj/project.pbxproj
```diff
@@ -1157,7 +1157,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "if which swiftlint >/dev/null; then\nswiftlint\nfi";
+			shellScript = "if which swiftlint >/dev/null && [[ `swiftlint version` =~ \"0.16\" ]] && [[ \"${CONFIGURATION}\" = \"Debug\" ]]; then\nswiftlint\nfi";
```
hzalaz: If we check the configuration first there is no need to try calling the command.

hzalaz:
Also we need to avoid running it in CIRCLE_CI since time consumed matters

cocojoe:
So I can remove the `lint` pass the `ios ci` build.  However the build script pass will still run, not sure how to get around that.  However in reality we are talking `0m0.134s` to run a lint pass.

cocojoe:
Well actually I can probably check a env to stop build phase script under CI however I still think we should have the official lint lane pass in Circle CI as it helps us catch it when swiftlint gets updated and the run time cost is near 0.

PR #95, opened 2017-03-03
Avoid WebAuth to retain UIApplication root ViewController

cocojoe:
If initialised before `viewDidAppear` the rootView would be stored as nil and any subsequent request to present WebAuth will fail.

### Example
```swift
class ViewController: UIViewController {

    var webAuth: WebAuth!

    override func viewDidLoad() {

        webAuth = Auth0.webAuth()
```

### DIFF
--- a/Auth0/ControllerModalPresenter.swift
+++ b/Auth0/ControllerModalPresenter.swift
@@ -24,27 +24,31 @@ import UIKit
 
 struct ControllerModalPresenter {
 
-    func present(_ controller: UIViewController) {
+    var rootViewController: UIViewController? {
+        return UIApplication.shared.keyWindow?.rootViewController
+    }
+
+    func present(controller: UIViewController) {
         topViewController?.present(controller, animated: true, completion: nil)
     }
 
     var topViewController: UIViewController? {
-        guard let root = UIApplication.shared.keyWindow?.rootViewController else { return nil }
-        return findTopViewController(root)
+        guard let root = self.rootViewController else { return nil }
+        return findTopViewController(from: root)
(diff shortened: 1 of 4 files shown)


Auth0/ControllerModalPresenter.swift
```diff
@@ -24,18 +24,12 @@ import UIKit
...
-        self.rootViewController = rootViewController
-    }
-
     func present(_ controller: UIViewController) {
         topViewController?.present(controller, animated: true, completion: nil)
     }
 
     var topViewController: UIViewController? {
-        guard let root = rootViewController else { return nil }
+        guard let root = UIApplication.shared.keyWindow?.rootViewController else { return nil }
```
hzalaz: I'd keep the old behaviour of allowing it to pass a root controller and if its nil, default to the one from `UIApplication` when there is the need to present a VC

Auth0/OAuth2Grant.swift
```diff
@@ -144,7 +144,7 @@ private func decode(jwt: String) -> [String: Any]? {
     let paddingLength = requiredLength - length
     if paddingLength > 0 {
         let padding = "".padding(toLength: Int(paddingLength), withPad: "=", startingAt: 0)
-        base64 = base64 + padding
+        base64 += padding
```
hzalaz: I am not a big fan of this, I'd rather use string interpolation so its clearer

cocojoe: Was really just to please a SwiftLint error, I'll make an exception. I was a bit surprised that this was an `error`

PR #104, opened 2017-03-22
Method to check native auth availability for provider in the device

cocojoe:
This is required to ensure the correct auth method can be selected for Native Plugins in situations where the native pathway becomes unavailable.


Auth0/AuthProvider.swift
```diff
@@ -35,8 +35,27 @@ import UIKit
...
     func login(withConnection connection: String, scope: String, parameters: [String: Any]) -> NativeAuthTransaction
+
+    /**
+     Some IdP authenticators like Auth0Twitter rely upon accessibility of on device social accounts.
+     If these are not accessible/available this pre-login check lets you know you should provide
+     alternative authentication means such as Auth0.WebAuth.
+     
+     - returns: Bool
+     */
+    func canUseNative() -> Bool
```
hzalaz: why not something more common like `isAvailable()` or `available` prop both static. Also the comment is a bit misleading since we are dealing Auth providers so something like

```
Determine if the Auth method used by the Provider is available in the device. e.g. if using a Twitter Auth provider it should check the presence of a Twitter account in the device.

If a Auth is performed on a provider that returns `false` the transaction will fail with an error

- returns: Bool if the AuthProvider is available on the device
```

cocojoe: Thx, amusingly I did call it `isAvailable()` originally then got conflicted looking at naming in v1 twitter.  Should have stuck with original name 👍

Auth0/AuthProvider.swift
```diff
@@ -35,8 +35,27 @@ import UIKit
...
+}
+
+/**
+ Extension to provide default(s)
+ */
+public extension AuthProvider {
```
cocojoe: Originally I didn't have a default as felt it's important for developers implementing to understand what it did.  Then I thought the most common use case (based on our use case) would be `true` so perhaps this is a convenience.  So i'll go back to original thoughts 👍

cocojoe:
@hzalaz SwiftLint now at 0.17, I've updated in this PR as was failing on Circle CI.  Only area of question here is https://github.com/auth0/Auth0.swift/blob/added-nativeauth-available/Auth0/Authentication.swift#L437

Not sure why SwiftLint is only mentioning this now that enum should be lower case, although the enum is actually mixed case as we have`.iOSLink`. This should really be updated to all lower case however it's public API and hence a breaking change, albeit it a rather small one and compiler will suggest the fix.... Thoughts?

PR #110, opened 2017-04-19
Added Serializable Protocol and Credentials conformance

cocojoe:
### Public API
Serialization Protocol
```swift
func serialize() -> String
init?(withSerialized string: String)
```

### Notes

After discussion, have not serialized a decoded idToken or any other form of profile. (Can change of course)
KeyChain storage should probably be in Lock, and QuickStarts should ideally be refactored to use this Serialization.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -93,6 +93,7 @@ fastlane/report.xml
 fastlane/Preview.html
 fastlane/screenshots
 fastlane/test_output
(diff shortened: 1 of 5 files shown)


cocojoe:
Don't feel this is required here after discussion, will provide appropriate utility functionality in Lock.Swift

PR #111, opened 2017-04-19
Added method to revoke refresh tokens

cocojoe:
Public API `func revoke(refreshToken: String)`
Added tests

### DIFF
--- a/Auth0/Auth0Authentication.swift
+++ b/Auth0/Auth0Authentication.swift
@@ -178,7 +178,6 @@ struct Auth0Authentication: Authentication {
 
     func revoke(refreshToken: String) -> Request<Void, AuthenticationError> {
         let payload: [String: Any] = [
-            "tokenEndpointAuthMethod": "none",
             "token": refreshToken,
             "client_id": self.clientId
         ]
--- a/Auth0/Auth0Authentication.swift
+++ b/Auth0/Auth0Authentication.swift
@@ -176,6 +176,16 @@ struct Auth0Authentication: Authentication {
         return Request(session: session, url: oauthToken, method: "POST", handle: authenticationObject, payload: payload, logger: self.logger, telemetry: self.telemetry)
     }
 
+    func revoke(refreshToken: String) -> Request<Void, AuthenticationError> {
+        let payload: [String: Any] = [
+            "tokenEndpointAuthMethod": "none",
+            "token": refreshToken,
+            "client_id": self.clientId
+        ]
+        let oauthToken = URL(string: "/oauth/revoke", relativeTo: self.url)!
+        return Request(session: session, url: oauthToken, method: "POST", handle: noBody, payload: payload, logger: self.logger, telemetry: self.telemetry)
+    }
+
     func delegation(withParameters parameters: [String : Any]) -> Request<[String : Any], AuthenticationError> {
         var payload: [String: Any] = [
             "client_id": self.clientId,
--- a/Auth0/Response.swift
+++ b/Auth0/Response.swift
@@ -57,7 +57,7 @@ struct Response<E: Auth0Error> {
             return json
         }
         // This piece of code is dedicated to our friends the backend devs :)
-        if response.url?.lastPathComponent == "change_password" {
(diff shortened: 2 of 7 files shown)


Auth0/Auth0Authentication.swift
```diff
@@ -176,6 +176,16 @@ struct Auth0Authentication: Authentication {
         return Request(session: session, url: oauthToken, method: "POST", handle: authenticationObject, payload: payload, logger: self.logger, telemetry: self.telemetry)
     }
 
+    func revoke(refreshToken: String) -> Request<Void, AuthenticationError> {
+        let payload: [String: Any] = [
+            "tokenEndpointAuthMethod": "none",
```
hzalaz: This is not needed

Auth0/Response.swift
```diff
@@ -57,7 +57,7 @@ struct Response<E: Auth0Error> {
             return json
         }
         // This piece of code is dedicated to our friends the backend devs :)
-        if response.url?.lastPathComponent == "change_password" {
+        if response.url?.lastPathComponent == "change_password" || response.url?.lastPathComponent == "revoke" {
```
hzalaz: Why should we need this?

Auth0Tests/AuthenticationSpec.swift
```diff
@@ -728,27 +759,27 @@ class AuthenticationSpec: QuickSpec {
                     }
                 }
             }
-
+
```
hzalaz: Check that you are adding spaces that should not be there

Auth0/Auth0Authentication.swift
hzalaz: Also this is missing the grant_type AFAIK, did it work without it?

Auth0Tests/AuthenticationSpec.swift
cocojoe: Odd...

Auth0/Auth0Authentication.swift
cocojoe: I based on https://auth0.com/docs/tokens/preview/refresh-token#revoke-a-refresh-token
grant_type not mentioned either.

Auth0/Response.swift
cocojoe: Well ideally we shouldn't, no reason API shouldn't return JSON eg { 'success': 1 }.  Both endpoints `change_password` and `revoke` return no body (HTTP 2OO only) success, so paired it up with the original hack/workaround (otherwise it will throw as there is no content).  

In general code is set up to expect JSON, you could change this `if/else` block to return nil. The `Handlers.swift` should certainly be able to deal with it, but no doubt other potential issues?

Auth0Tests/AuthenticationSpec.swift
cocojoe: Only change recently in editor was upgrading to 8.3, generally with tests I run  `Structure/Re-Indent`

hzalaz: Yes on update the config get messed up most of the times.

Auth0/Response.swift
hzalaz: That if is a hack to avoid blowing up for change_password since it does not follow the rules of content negotiation. I'd avoid adding extra hacks like this, I though we have a way to tell there is no response or I will ignore the response (Maybe in Guardian.swift we did something like it)

Auth0/Auth0Authentication.swift
hzalaz: But you can ask the rest of the team who already implemented the feature

hzalaz: I believe this is already done in Android right @lbalmaceda ?

Auth0/Response.swift
cocojoe: It's deal with in Guardian, not so transferrable https://github.com/auth0/Guardian.swift/blob/master/Guardian/Request.swift#L88

I looked at the `changed_password` response, forgot it returned plain text.  
I've made a change, feel free to tell me if it sucks :).

hzalaz:
@cocojoe wait a sec to merge

PR #112, opened 2017-04-27
Add Mac platform to Fastlane, multiple devices iOS

cocojoe:
Added Multiple devices for iOS
Removed Scanfile (No multi platform support)
Skip Docs fastlane
Update fastlane


cocojoe:
tvOS is not supported in fastlane, it could be added in a more manual fashion with xcodebuilds in another PR.

cocojoe:
Going to close this for now as discussed to do something based on this, will leave branch for reference.

PR #113, opened 2017-05-02
Credentials support NSSecureCoding, CredentialsManager Utility, KeyChain Storage

cocojoe:
Although I was very tempted to add SimpleKeyChain to this repo, I felt we should leave it up to the user.  NS*Coding is how you generally expect to serialize information in iOS.  I don't feel we need to add any convenience wrappers here to `NSKeyedArchiver/NSKeyedUnArchiver` as they are essentially one liners and is yet another public API.

### DIFF
--- a/Auth0/Authentication.swift
+++ b/Auth0/Authentication.swift
@@ -753,4 +753,44 @@ public extension Authentication {
     func renew(withRefreshToken refreshToken: String, scope: String? = nil) -> Request<Credentials, AuthenticationError> {
         return self.renew(withRefreshToken: refreshToken, scope: scope)
     }
+
+    /**
+     Renew expired user's credentials using user's current credentials
+
+     ```
+     Auth0.renewExpired(credentials, scope: "openid profile offline_access") {
+         guard $0 == nil else { return }
+         print($1)
+     }
+     ```
+
+     or asking the same scopes requested when the refresh token was issued
+
+     ```
+     Auth0.renewExpired(credentials) {
+         guard $0 == nil else { return }
+         print($1)
+     }
+     ```
+
+     - parameter credentials: the client's credentials obtained on auth
+     - parameter scope: scopes to request for the new tokens. By default is nil which will ask for the same ones requested during Auth
+     - parameter callback: callback with the user's credentials or the cause of the error.
+     - important: This method only works for a refresh token obtained after auth with OAuth 2.0 API Authorization.
+     */
+    func renewExpired(_ credentials: Credentials, scope: String? = nil, callback: @escaping (Error?, Credentials?) -> Void) {
+        guard let refreshToken = credentials.refreshToken else {
+            return callback(AuthenticationError(string: "missing refresh_token", statusCode: 400), nil)
+        }
+        guard let expiresIn = credentials.expiresIn, expiresIn < Date() else { return callback(nil, credentials) }
+
+        self.renew(withRefreshToken: refreshToken, scope: scope).start {
+            switch $0 {
+            case .success(let credentials):
+                callback(nil, credentials)
+            case .failure(let error):
+                callback(error, nil)
+            }
+        }
+    }
 }
--- a/Auth0/Authentication.swift
+++ b/Auth0/Authentication.swift
@@ -753,44 +753,4 @@ public extension Authentication {
     func renew(withRefreshToken refreshToken: String, scope: String? = nil) -> Request<Credentials, AuthenticationError> {
         return self.renew(withRefreshToken: refreshToken, scope: scope)
     }
-
-    /**
-     Renew expired user's credentials using user's current credentials
-
-     ```
-     Auth0.renewExpired(credentials, scope: "openid profile offline_access") {
-         guard $0 == nil else { return }
-         print($1)
-     }
-     ```
-
-     or asking the same scopes requested when the refresh token was issued
-
-     ```
-     Auth0.renewExpired(credentials) {
-         guard $0 == nil else { return }
-         print($1)
-     }
-     ```
-
-     - parameter credentials: the client's credentials obtained on auth
-     - parameter scope: scopes to request for the new tokens. By default is nil which will ask for the same ones requested during Auth
-     - parameter callback: callback with the user's credentials or the cause of the error.
-     - important: This method only works for a refresh token obtained after auth with OAuth 2.0 API Authorization.
-     */
-    func renewExpired(_ credentials: Credentials, scope: String? = nil, callback: @escaping (Error?, Credentials?) -> Void) {
-        guard let refreshToken = credentials.refreshToken else {
-            return callback(AuthenticationError(string: "missing refresh_token", statusCode: 400), nil)
-        }
-        guard let expiresIn = credentials.expiresIn, expiresIn < Date() else { return callback(nil, credentials) }
-
-        self.renew(withRefreshToken: refreshToken, scope: scope).start {
-            switch $0 {
-            case .success(let credentials):
-                callback(nil, credentials)
-            case .failure(let error):
-                callback(error, nil)
-            }
-        }
-    }
 }
--- a/Auth0/CredentialsManager.swift
+++ b/Auth0/CredentialsManager.swift
@@ -46,10 +46,9 @@ public struct CredentialsManager {
         return self.storage.setData(NSKeyedArchiver.archivedData(withRootObject: credentials), forKey: storeKey)
     }
 
-    /// Retrieve credentials from keychain and yield new credentials if the accessToken has expired
+    /// Retrieve credentials from keychain and yield new credentials using refreshToken if accessToken has expired
     /// otherwise the retrieved credentails will be returned as they have not expired.
     ///
-    /// - seeAlso: [Auth0 Refresh Tokens Docs](https://auth0.com/docs/tokens/refresh-token)
     ///
     /// ```
     /// credentialsManager.credentials {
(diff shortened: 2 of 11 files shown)


Auth0/Authentication.swift
```diff
@@ -774,15 +774,15 @@ public extension Authentication {
...
      */
     func renewExpired(_ credentials: Credentials, scope: String? = nil, callback: @escaping (Error?, Credentials?) -> Void) {
-        guard
-            let refreshToken = credentials.refreshToken,
-            let expiresIn = credentials.expiresIn, expiresIn < Date(timeIntervalSinceNow: -60)
-            else { return callback(AuthenticationError(string: "missing refresh_token or have not expired", statusCode: 400), nil) }
+        guard let refreshToken = credentials.refreshToken else {
+            return callback(AuthenticationError(string: "missing refresh_token", statusCode: 400), nil)
+        }
+        guard let expiresIn = credentials.expiresIn, expiresIn < Date(timeIntervalSinceNow: -60) else { return callback(nil, credentials) }
```
lbalmaceda: What does the `-60` means? Now that I see this, it reminds me that some tokens might be deemed invalid wrongly depending on the device's clock. Remember the leeway thing we support in our jwt libs? Maybe we should open a method to customize it too for this case. It should not be an issue for most of the users but some of them might have it.

cocojoe: I added the 60 although it should perhaps be longer based like 5 minutes.  As it would be if a token was not renewed as it was close to expiry and then immediately expires when the user goes to use it a few seconds later.
I've not looked at this in the JWT Libs, although I feel if a user's clock is wrong this isn't really our issue.

lbalmaceda: The token should be used right after the user asks for it, so it's up to him if he takes too long and the server rejects it.. Maybe not 5 minutes, but 2 min already is too much. Remember that the problem here is the server, because we can still decode a jwt payload if needed even if expired.. Maybe it's best to check the server "allowed leeway" and pick a smaller value just to be sure. Anyway, it's not too hard to log in again 😛. I'll approve this pr and let you decide if a change it's needed.

hzalaz: We should have a leeway eventually just in case to avoid issues with customers. Having 5 minutes or a big one is not that a big deal since the RS will properly validate the AT and reject it on API call

Auth0/Authentication.swift
```diff
@@ -753,4 +753,44 @@ public extension Authentication {
...
+         print($1)
+     }
+     ```
+
+     - parameter credentials: the client's credentials obtained on auth
+     - parameter scope: scopes to request for the new tokens. By default is nil which will ask for the same ones requested during Auth
+     - parameter callback: callback with the user's credentials or the cause of the error.
+     - important: This method only works for a refresh token obtained after auth with OAuth 2.0 API Authorization.
+     */
+    func renewExpired(_ credentials: Credentials, scope: String? = nil, callback: @escaping (Error?, Credentials?) -> Void) {
```
hzalaz: This sort of breaks all the convention of API with the request that the rest of the methods have. If we want to avoid that we can move this to a different type

Auth0/Authentication.swift
```diff
@@ -753,4 +753,44 @@ public extension Authentication {
...
+        guard let refreshToken = credentials.refreshToken else {
+            return callback(AuthenticationError(string: "missing refresh_token", statusCode: 400), nil)
```
hzalaz: We should use a different type of error for this case

Auth0/Authentication.swift
hzalaz: Also the naming is a bit odd, since it won't renew expired all the time

hzalaz: If we add the keychain storing how the whole story will look like?

Auth0/Authentication.swift
cocojoe: I wanted to avoid creating another error type.  `WebAuthError` is closer but as it deals with missing token cases or can create a new error type but it's pretty much one case.

Auth0/Authentication.swift
cocojoe: Regarding naming, well `renewExpired` implies only renew when expired, you will get credentials back if they have not expired they are just not renewed as they aren't expired.

Yes, I was conflicted over where to put it.  I put it here as it requires `Authentication` so rather than passing an `Authentication` instance it, but agree the breaking of convention makes it not quite feel right from an API purity point of view.

I ran originally ran through a few scenarios, for example I liked the idea of a static method in `Credentials` or in a `Storage` utility class.

cocojoe:
```
public class Storage: A0SimpleKeychain {

    public func archive(object: NSSecureCoding, forKey key: String) -> Bool {
        return self.setData(NSKeyedArchiver.archivedData(withRootObject: object), forKey: key)
    }

    public func unarchive(objectWithKey key: String) -> Any? {
        guard let data = self.data(forKey:key) else { return nil }
        return NSKeyedUnarchiver.unarchiveObject(with: data)
    }

   public func renewExpired(_ credentials, authentication...) { }
}
```

cocojoe: One thought:
```
let storedCredentials = storage.unarchive("credentials") as? Credentials
// Class method in Credentials
Credentials.renewExpired(credentials, authentication: authentication) { error, credentials in }
```

You could of course add a 1 call wrapper as well to unarchive and renew from storage, which people would use.

hzalaz: The inheritance there is not needed and tbh it exposes way too much info and API. Also you could have achieved the same thing with an extension.

I'd say we iterate the idea a bit more with a separate entity to handle the Auth state.

If I read it `renewExpired` seems to be imply that it might be called when I know the token is expired not to ensure I have a fresh/valid/nonexpired token

Auth0/Authentication.swift
hzalaz: Ok but put status code 0 not 400

cocojoe:
@hzalaz after todays talks, I've added `SimpleKeychain` as it makes sense to have this here now. I've used my `SimpleKeychain` branch as it supports all platforms that `Auth0.Swift` does.

Turned this into a Credentials utility for easy, store, retrieve, renew.  Sort of thing be nice to see in the Centralized QS.
I've not enforced expiry checking as I feel this is an education issue.

I can update tests and add docheaders in morning, wanted to check it was the right direction to take. Not 100% sure on name though.

cocojoe:
I am still not sure on `renewExpired`, I like the idea of a best practice (as user probably not bother otherwise) method to `retrieve -> check expiry -> (return or renew)`. 
In last iteration this was an exercise for the user, `retrieve -> check expiry yourself -> use or call renew`.
Perhaps really it just needs a better name to imply retrieval + optional renew if expired.
`retrieveCredentialsAutoRenew`

Auth0/CredentialsManager.swift
```diff
@@ -0,0 +1,89 @@
...
+    /// - Parameter credentials: credentials instance to store
+    /// - Returns: Bool outcome of success
+    public func store(_ credentials: Credentials) -> Bool {
+        return self.storage.setData(NSKeyedArchiver.archivedData(withRootObject: credentials), forKey: storeKey)
+    }
+
+    /// Retrieve stored credentials instance from keychain
+    ///
+    /// - Returns: Optional Credentials instance
+    public func retrieve() -> Credentials? {
```
hzalaz: Lets remove this one in favor of the next one

Auth0/CredentialsManager.swift
```diff
@@ -0,0 +1,89 @@
...
+    ///   - authentication: Auth0 authentication instance
+    public init(authentication: Authentication) {
+        self.authentication = authentication
+    }
+
+    /// Store credentials instance in keychain
+    ///
+    /// - Parameter credentials: credentials instance to store
+    /// - Returns: Bool outcome of success
+    public func store(_ credentials: Credentials) -> Bool {
```
hzalaz: Lets not use `_`

Auth0/CredentialsManager.swift
```diff
@@ -0,0 +1,89 @@
...
+    ///    guard $0 == nil else { return }
+    ///    print($1)
+    /// }
+    /// ```
+    ///
+    /// - Parameters:
+    ///   - scope: scopes to request for the new tokens. By default is nil which will ask for the same ones requested during original Auth
+    ///   - callback: callback with the user's credentials or the cause of the error.
+    /// - Important: This method only works for a refresh token obtained after auth with OAuth 2.0 API Authorization.
+    public func retrieveCredentialsAutoRenew(withScope scope: String? = nil, callback: @escaping (CredentialsManagerError?, Credentials?) -> Void) {
```
hzalaz: Just call it either `credentials(withScope:, callback:)` or `withCredentials(scope:, callback)` the idea is to use this one to try to get credentials to be used. If it fails the user should login otherwise it will always yield a valid token

Auth0/CredentialsManagerError.swift
```diff
@@ -0,0 +1,30 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import Foundation
+
+public enum CredentialsManagerError: Error {
```
hzalaz: I am not sure about this list of errors, few of them are actionable for the developer

Auth0/CredentialsManager.swift
cocojoe: I don't mind either way, however isn't it clear inside a `CredentialsManager` that has a param type of `Credentials` that it is going to store a credentials instance?

Auth0/CredentialsManager.swift
cocojoe: Thx, sometimes naming is the hardest part.

Auth0/CredentialsManagerError.swift
cocojoe: I tend to add them for logging, could argue that expiresIn and refreshToken could be one guard and one error e.g. just `.noRefreshToken`

Auth0/CredentialsManager.swift
hzalaz: more for the flow of the api `manager.store(credentials: result)` then in the future we could have `manager.store(accessToken: token, expiresIn: date)` to avoid passing all the credential object.

Auth0/CredentialsManagerError.swift
hzalaz: There should be 3 cases

- Failed to refresh (api call or no expires In)
- No credentials (no creds)
- No refresh token (only raised when trying to renew)

PR #115, opened 2017-05-09
Added WebAuth Auth0 Session Clear

cocojoe:
This call is silent/hidden with no redirect as it's not a web app so the only requirement, there is a callback so the user can do something if they want to from the result.

### DIFF
--- a/Auth0.podspec
+++ b/Auth0.podspec
@@ -10,6 +10,7 @@ web_auth_files = [
   'Auth0/WebAuthError.swift',
   'Auth0/SafariWebAuth.swift',
   'Auth0/SafariSession.swift',
+  'Auth0/SilentSafariViewController.swift',
   'Auth0/NativeAuth.swift',
   'Auth0/AuthProvider.swift'
 ]
--- a/Auth0/SafariWebAuth.swift
+++ b/Auth0/SafariWebAuth.swift
@@ -182,6 +182,13 @@ class SafariWebAuth: WebAuth {
             .appendingPathComponent(bundleIdentifier)
             .appendingPathComponent("callback")
     }
+
+    func logout(useFederated federated: Bool = false, callback: @escaping (Bool) -> Void) {
+        let logoutURL = federated ? URL(string: "/v2/logout?federated", relativeTo: self.url)! : URL(string: "/v2/logout", relativeTo: self.url)!
+        let controller = SilentSFSafariViewController(url: logoutURL) { callback($0) }
+        logger?.trace(url: logoutURL, source: "Safari")
+        self.presenter.present(controller: controller)
+    }
 }
 
 private func generateDefaultState() -> String? {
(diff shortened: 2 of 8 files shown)


Auth0/SafariWebAuth.swift
```diff
@@ -194,3 +201,19 @@ private func generateDefaultState() -> String? {
     guard result == 0 else { return nil }
     return data.a0_encodeBase64URLSafe()
 }
+
+private class SilentSFSafariViewController: SFSafariViewController, SFSafariViewControllerDelegate {
```
hzalaz: Let's move this so a diff file

cocojoe:
@hzalaz I would say it's as good as transparent, you can see it faintly at the top.
logout

Auth0/WebAuth.swift
```diff
@@ -200,4 +200,20 @@ public protocol WebAuth: Trackable, Loggable {
...
+     ```
+     Auth0
+        .webAuth()
+        .clearSession(useFederated: false) { print($0) }
+     ```
+     
+     - parameter federated: Bool to enable signout of the IdP
+     - parameter callback: callback called with the result of the logout
+     */
+    func clearSession(useFederated federated: Bool, callback: @escaping (Bool) -> Void)
```
hzalaz: Just `federated` and probably make it an optional  parameter

Auth0/WebAuth.swift
```diff
@@ -200,4 +200,20 @@ public protocol WebAuth: Trackable, Loggable {
      - parameter callback: callback called with the result of the WebAuth flow
      */
     func start(_ callback: @escaping (Result<Credentials>) -> Void)
+
+    /**
+     Clear the SSO Cookie in Auth0 and optionally signout from IdP.
```
hzalaz: Removes Auth0 session and optionally allows to remove the Identity Provider session.

Auth0/WebAuth.swift
```diff
@@ -200,4 +200,20 @@ public protocol WebAuth: Trackable, Loggable {
...
+     More info: https://auth0.com/docs/logout
```
hzalaz: Use `note` or `seeAlso` like in the rest of the docs of the public API

Auth0/SilentSafariViewController.swift
```diff
@@ -0,0 +1,40 @@
...
+
+    required init(url URL: URL, callback: @escaping (Bool) -> Void) {
+        super.init(url: URL, entersReaderIfAvailable: false)
+        self.onResult = callback
+        self.delegate = self
+        self.view.alpha = 0.05 // Apple does not allow invisible SafariViews, this is the threshold.
+        self.modalPresentationStyle = .overCurrentContext
+    }
+
+    func safariViewController(_ controller: SFSafariViewController, didCompleteInitialLoad didLoadSuccessfully: Bool) {
```
hzalaz: What if it fails? should we handle that case?

Auth0.podspec
```diff
@@ -10,6 +10,7 @@ web_auth_files = [
   'Auth0/WebAuthError.swift',
   'Auth0/SafariWebAuth.swift',
   'Auth0/SafariSession.swift',
+  'SilentSafariViewController.swift',
```
hzalaz: Why its in the root and not with the rest ?

Auth0/WebAuth.swift
cocojoe: Annoyingly, `seeAlso` appears to be broken in Xcode, few bug reports out there.

Auth0/WebAuth.swift
cocojoe: Originally was going to default to `false` (not a fan of optional booleans) then thought the user should understand what method they are calling.

Auth0/SilentSafariViewController.swift
cocojoe: Not sure I understand, the controller is dismissed and the success status `didLoadSuccessfully` is returned in the callback.
Are you saying we should wrap it into an Error type for coding style?

hzalaz: If the url is invalid? if there is no internet connection? what will happened then?

Auth0/WebAuth.swift
hzalaz: All good making it mandatory then

Auth0/WebAuth.swift
hzalaz: Yeah, they changed the parser yet again

Auth0/SilentSafariViewController.swift
cocojoe: If the URL called in the SFSafariViewController responds with anything other than HTTP 200 OK the callback will return false. (Have also manually tested this).

PR #116, opened 2017-05-10
Added Touch ID Utility

cocojoe:


cocojoe:
@hzalaz To be helpful to the end user, have coupled to CredentialsManager as #1 use case, most likely to store / retrieve credentials.  There is a `requireTouch` for general purposes.
Will rebase ones other PR is merged.  
If we want a really simple TouchAuthentication, decoupled from credentials, easy enough to remove.
Will add tests, just wanted to open comments up.

cocojoe:
Have stripped this back, decoupled from `credentialsManager` easy enough to wrap up a `store` or `retrieve/renew` inside a touch callback.

Note: Moved base in PR for easier to read diff.

cocojoe:
Rebased to master

Auth0/TouchAuthentication.swift
```diff
@@ -0,0 +1,52 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import Foundation
+import LocalAuthentication
+
+public struct TouchAuthentication {
```
hzalaz: Should not be public

Auth0/TouchAuthentication.swift
```diff
@@ -0,0 +1,52 @@
...
+
+    public var available: Bool {
+        return self.authContext.canEvaluatePolicy(LAPolicy.deviceOwnerAuthenticationWithBiometrics, error: nil)
+    }
+
+    public func requireTouch(withTitle title: String, cancelTitle: String? = nil, callback: @escaping (Error?) -> Void) {
+        if #available(iOS 10, *) {
+            self.authContext.localizedCancelTitle = cancelTitle
+        }
+        self.authContext.localizedFallbackTitle = ""
```
hzalaz: This is to hide the `use passcode`?

Auth0/TouchAuthentication.swift
```diff
@@ -0,0 +1,52 @@
...
+
+        self.authContext.evaluatePolicy(LAPolicy.deviceOwnerAuthenticationWithBiometrics, localizedReason: title) {
```
hzalaz: Add param names since:

- there are 2
- they are used more than once

Auth0/TouchAuthentication.swift
```diff
@@ -0,0 +1,52 @@
...
+            guard $1 == nil else {
```
hzalaz: What about?

```swift 
if let error = error {
   return callback(error);
}
callback(success ? nil : LAError(LAError.authenticationFailed))
```

Auth0/CredentialsManager.swift
```diff
@@ -80,4 +82,11 @@ public struct CredentialsManager {
             }
         }
     }
+
+    public func credentialsOnTouch(withTitle title: String, scope: String? = nil, callback: @escaping (CredentialsManagerError?, Credentials?) -> Void) {
```
hzalaz: Can't we keep the other interface and make TouchID a configuration thing of the credential manager?

Auth0/TouchAuthentication.swift
cocojoe: oh, originally I added it so it could be used in Lock as a helper.

Auth0/TouchAuthentication.swift
cocojoe: Yes, I wanted to be strict. Could be left to the developer to decide?

Auth0/CredentialsManager.swift
cocojoe: We could now yes.

cocojoe: Originally I was thinking of a `useTouch` flag in `CredentialsManager`, will do something similar.

PR #120, opened 2017-06-08
Added scope property to Credentials

cocojoe:


cocojoe:
@lbalmaceda See if the `scope` property doc makes sense

hzalaz:
@lbalmaceda it will be `[String]` 😜 .

@cocojoe expanding what @lbalmaceda said we could have a method to help check that the returned scopes contains a list of scopes like

```swift
let requested = "openid profile email read:user post:ticket"
...
let missing: [String] = credentials.rejected(from scope: requested)
```

but it could be material of another PR so Android is in sync too

cocojoe:
Agree a helper would be nice, as this has been approved already will merge and raise another PR.

PR #122, opened 2017-06-09
Added OIDC Conformant UserInfo class and API Method

cocojoe:
Public API  Added (Open to name changes)
- `func userClaimInfo(token: String) -> Request<UserInfo, AuthenticationError>` 
-  `public class UserInfo: NSObject, JSONObjectPayload`

This adheres to the standard claims specification, although we may not support all claims. I feel we should support all properties for developer convenience.  For example Facebook `openid profile` will return.
```
{"sub":"facebook|1010000000000","name":"Martin Walsh","given_name":"Martin","family_name":"Walsh"
,"nickname":"martin.walsh"
,"picture":"https://scontent.xx.fbcdn.net/v/t1.0-1/p50x50/myawesomeprofile.jpg"
,"gender":"male","locale":"en-US","updated_at":"2017-06-09T14:49:21.310Z"}
```

### DIFF
--- a/Auth0/Authentication.swift
+++ b/Auth0/Authentication.swift
@@ -299,7 +299,7 @@ public protocol Authentication: Trackable, Loggable {
     func tokenInfo(token: String) -> Request<Profile, AuthenticationError>
 
     /**
-     Returns user information by performing a request to /userinfo endpoint
+     Returns user information by performing a request to /userinfo endpoint.
 
      ```
      Auth0
@@ -311,8 +311,27 @@ public protocol Authentication: Trackable, Loggable {
      - parameter token: token obtained by authenticating the user
 
      - returns: a request that will yield user information
+     - important: If you are using an OIDC Conformant client please see `userClaimInfo`
      */
     func userInfo(token: String) -> Request<Profile, AuthenticationError>
+
+    /**
+     Returns OIDC standard claims information by performing a request
+     to /userinfo endpoint.
+
+     ```
+     Auth0
+     .authentication(clientId, domain: "samples.auth0.com")
+     .userClaimInfo(token: token)
+     .start { print($0) }
+     ```
+
+     - parameter token: token obtained by authenticating the user
+
+     - returns: a request that will yield user information
+     - important: This method should be used for OIDC Conformant clients.
+     */
+    func userClaimInfo(token: String) -> Request<UserInfo, AuthenticationError>
 
     /**
      Logs in a user using a social Identity Provider token. e.g. Facebook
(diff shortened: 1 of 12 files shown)


cocojoe:
I need to add tests, just wanted to open up any further discussion.
Could possibly deprecate the old `userInfo`.

Auth0/UserInfo.swift
```diff
@@ -0,0 +1,140 @@
...
+        let phoneNumber = json["phone_number"] as? String
+        let phoneNumberVerified = json["phone_number_verified"] as? Bool
+        let address = json["address"] as? [String: String]
+
+        var updatedAt: Date?
+        if let dateString = json["updated_at"] as? String {
+            updatedAt = date(from: dateString)
+        }
+
+        let keys = Set(["sub", "name", "given_name", "family_name", "middle_name", "nickname", "preferred_username", "profile", "picture", "website", "email", "email_verified", "gender", "birthdate", "zoneinfo", "locale",
```
hzalaz: And if we do

```swift
let publicClaims = ["sub", "name", "given_name", "family_name", "middle_name", "nickname", "preferred_username", "profile", "picture", "website", "email", "email_verified", "gender", "birthdate", "zoneinfo", "locale","phone_number", "phone_number_verified", "address", "updated_at"] // This can even be a static thing outside of the class since it won't change
let customClaims = json.filter { !publicClaims.contains($0.0) }
```

?

cocojoe: I agree `publicClaims` should be static.
The issue with `filter` here is it will return an array of tuples, so it's not quite as elegant as this one liner.
In Swift 4 this works as a filtered dictionary will return a dictionary. (As it should really)

Auth0/Authentication.swift
```diff
@@ -311,10 +311,29 @@ public protocol Authentication: Trackable, Loggable {
...
+     .userClaimInfo(token: token)
+     .start { print($0) }
+     ```
+
+     - parameter token: token obtained by authenticating the user
+
+     - returns: a request that will yield user information
+     - important: This method should be used for OIDC Conformant clients.
+     */
+    func userClaimInfo(token: String) -> Request<UserInfo, AuthenticationError>
```
hzalaz: Let's try to keep the `userInfo` name

Auth0/Profile.swift
```diff
@@ -103,7 +103,7 @@ public class Profile: NSObject, JSONObjectPayload {
 
 }
 
-private func date(from string: String) -> Date? {
+func date(from string: String) -> Date? {
```
hzalaz: Try moving this to other file since its shared

PR #125, opened 2017-06-15
Use new SFSafariViewController init for iOS11

cocojoe:
Update Nimble dependency
Xcode8 and 9 compatible


PolyglotC:
Needing this :)

cocojoe:
@PolyglotC there is a swift-4 beta branch btw https://github.com/auth0/Auth0.swift/tree/swift-4.0.beta

PR #133, opened 2017-07-06
Added credential manager methods `clear` and `hasValid`

cocojoe:
Additional methods added to `CredentialsManager`

`clear()` -> Delete credentials from keychain storage
`hasValid()` -> Quick check that valid credentials are available & not expired


Auth0/CredentialsManager.swift
```diff
@@ -58,6 +58,26 @@ public struct CredentialsManager {
         return self.storage.setData(NSKeyedArchiver.archivedData(withRootObject: credentials), forKey: storeKey)
     }
 
+    /// Clear credentials stored in keychain
+    ///
+    /// - Returns: Bool outcome of removal success
```
hzalaz: no need to state the `Bool` type, just say true if credentials were removed, false if no credentials were found or it failed to remove`

Auth0/CredentialsManager.swift
```diff
@@ -58,6 +58,26 @@ public struct CredentialsManager {
...
+    public func clearCredentials() -> Bool {
+        return self.storage.deleteEntry(forKey: storeKey)
+    }
+
+    /// Checks if valid credentials are available
+    ///
+    /// - Returns: Bool outcome of result
```
hzalaz: no need to state the Bool type, just say if there are valid and non-expired credentials stored`

Auth0/CredentialsManager.swift
```diff
@@ -58,6 +58,26 @@ public struct CredentialsManager {
...
+    public func hasCredentials() -> Bool {
+        guard
+            let data = self.storage.data(forKey:self.storeKey),
+            let credentials = NSKeyedUnarchiver.unarchiveObject(with: data) as? Credentials,
+            credentials.accessToken != nil,
+            let expiresIn = credentials.expiresIn, expiresIn > Date()
```
hzalaz: I think it will be clear to have

```swift
let expiresIn = credentials.expiresIn else { return false }
return expiresIn > Date()
```

hzalaz:
Also I might be inclined to have just `clear` and `hasValid` without the `Credentials`

cocojoe:
@hzalaz These were my original thoughts on naming conventions as it's implied by the context of the class.  However the argument is consistency with Android although the Swift API guidelines would say **Omit Needless Words** 😄

Auth0/CredentialsManager.swift
```diff
@@ -53,11 +53,31 @@ public struct CredentialsManager {
     /// Store credentials instance in keychain
     ///
     /// - Parameter credentials: credentials instance to store
-    /// - Returns: Bool outcome of success
+    /// - Returns: if credentials were stored
```
lbalmaceda: how can this fail?

Auth0/CredentialsManager.swift
```diff
@@ -53,11 +53,31 @@ public struct CredentialsManager {
...
     public func store(credentials: Credentials) -> Bool {
         return self.storage.setData(NSKeyedArchiver.archivedData(withRootObject: credentials), forKey: storeKey)
     }
 
+    /// Clear credentials stored in keychain
+    ///
+    /// - Returns: if credentials were removed
+    public func clear() -> Bool {
+        return self.storage.deleteEntry(forKey: storeKey)
```
lbalmaceda: how can this fail?

Auth0/CredentialsManager.swift
```diff
@@ -53,11 +53,31 @@ public struct CredentialsManager {
...
+    ///
+    /// - Returns: if there are valid and non-expired credentials stored
+    public func hasValid() -> Bool {
+        guard
+            let data = self.storage.data(forKey:self.storeKey),
+            let credentials = NSKeyedUnarchiver.unarchiveObject(with: data) as? Credentials,
+            credentials.accessToken != nil,
+            let expiresIn = credentials.expiresIn
+            else { return false }
+        return expiresIn > Date()
```
lbalmaceda: This is one of the conditions. But if `expiresIn <= Date() && credentials.refreshToken != nil` you will be able to refresh it, so that's another case that returns "valid credentials". No network call required here.

Auth0/CredentialsManager.swift
cocojoe: The underlying method can fail, it's just unlikely to ever happen :)

Auth0/CredentialsManager.swift
cocojoe: The underlying method can fail, it's just unlikely to ever happen :)

Auth0/CredentialsManager.swift
cocojoe: I don't know if the refreshToken is valid or not, could be revoked.  This check tells me if the current accessToken has expired or not, if it has expired it is invalid and the user should request fresh credentials with `retrieveCredentials`.

lbalmaceda: Having the refresh_token revoked is an edge case. Admin has to do that either manually in the user's dashboard or calling the MGMT API with a well-scoped token. I rather pay the price of a 401 and make the user clear the credentials than saying "your token is no longer valid" when I know that I can refresh it. That it's also a valid credentials scenario IMO.

hzalaz: This is just checking the AT (credentials) not if the RT is valid or not

hzalaz: If it were me I'd have the following props:

- empty: if there is an AT stored
- expired: if the AT is expired
- canRefresh: if there is a RT

PR #138, opened 2017-08-04
Added changePassword to Authentication API

cocojoe:
Added tests


cocojoe:
Pending endpoint finalisation

PR #141, opened 2017-08-15
Added /oauth/token invalid password to AuthenticationError

cocojoe:


Auth0/AuthenticationError.swift
```diff
@@ -121,7 +121,7 @@ public class AuthenticationError: Auth0Error, CustomStringConvertible {
 
     /// When username and/or password used for authentication are invalid
     public var isInvalidCredentials: Bool {
-        return self.code == "invalid_user_password"
+        return self.code == "invalid_user_password" || (self.code == "invalid_grant" && self.description == "Wrong email or password.")
```
lbalmaceda: parentheses `( ... )` can be ignored because the `||` is separating the condition.

cocojoe: I like them, makes it clear

cocojoe:
Going to close this and do it in Lock now.

Issue #144, opened 2017-08-31
Refresh token is lost after refresh


danielphillips:
Note that logging in to retrieve credentials with Auth0 returns: `access_token`, `expires_in`, `id_token`, `refresh_token`, `scope` and `token_type`; however when refreshing the credentials, `refresh_token` is not returned in the response.

---

If using the `CredentialsManager`, after logging in one would call `store(credentials:)` to persist the credentials for later use.

When credentials are needed, one would call `credentials(withScope:callback:)` on a `CredentialsManager` instance, and if the access token is expired the credentials manager is kind enough to renew the token on our behalf, transparently returning brand new credentials, which we then save again using `store(credentials:)` again.

However, since the credentials are stored onto one key in the keychain, and given the fact that `refresh_token` is not returned in the response when refreshing the token, the new credentials have no refresh token. 

Inside `retrieveCredentials(withScope:callback:)` of `CredentialsManager` there is a check to make sure the credentials have a refresh token, if it isn't there then the credentials manager returns a `.noRefreshToken` error.

At this point in time I must log out and login again. 

We think this is an oversight on the `CredentialsManager`.

I did however want to raise this issue on the chance that I'm not using the SDK as intended. I would greatly appreciate any advice or information on this. 

Thank you

gscalzo:
Even if not ideal, a quick solution is in https://github.com/auth0/Auth0.swift/pull/145

cocojoe:
Thanks, I've tested this now and can confirm. I am also going to propose auto saving the new credentials as a convenience as I feel it's the most common scenario for the intended audience of this utility.

cocojoe:
Merged

PR #145, opened 2017-08-31
[fix] - #144 Refresh token is lost after refresh

gscalzo:
The refresh toke in saved into the credentials after being renewed. Not sure it's the best way to handle this situation, however it fixes our scenario.
Added a spec to cover the issue

### DIFF
--- a/Auth0/CredentialsManager.swift
+++ b/Auth0/CredentialsManager.swift
@@ -120,7 +120,13 @@ public struct CredentialsManager {
         self.authentication.renew(withRefreshToken: refreshToken, scope: scope).start {
             switch $0 {
             case .success(let credentials):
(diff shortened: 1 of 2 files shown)


danielphillips:
I guess the ideal solution would be to not serialise the Credentials alone to the keychain, it could maybe have a dictionary with the credentials and refresh token separate, serialising that means then you wouldn't need to reinfect a refresh token after refreshing.

cocojoe:
@gscalzo can I just check how you are calling renew that the refresh_token is being lost.  Thx

cocojoe:
Thanks for the contribution, I am going to close this PR and make some other changes. 👍 for adding a test 😄

PR #146, opened 2017-09-05
Fix - Ensure existing refreshToken returned in Credentials Manager

cocojoe:
Added - Renewed credentials are autosaved in Credentials Manager


cocojoe:
#144

Auth0/CredentialsManager.swift
```diff
@@ -120,7 +121,14 @@ public struct CredentialsManager {
...
             switch $0 {
             case .success(let credentials):
-                callback(nil, credentials)
+                let newCredentials = Credentials(accessToken: credentials.accessToken,
+                                                       tokenType: credentials.tokenType,
+                                                       idToken: credentials.idToken,
+                                                       refreshToken: refreshToken,
+                                                       expiresIn: credentials.expiresIn,
+                                                       scope: credentials.scope)
+                _ = self.store(credentials: newCredentials)
```
lbalmaceda: what does the `_` mean?? I can't find it as a declared variable in the class scope.

Auth0/_ObjectiveManagementAPI.swift
```diff
@@ -69,7 +69,6 @@ public class _ObjectiveManagementAPI: NSObject {
     }
 
     @objc(unlinkUserWithIdentifier:provider:fromUserId:callback:)
-    // swiftlint:disable:next function_parameter_count
```
lbalmaceda: Why are all this lint checks removed?

Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -175,18 +176,34 @@ class CredentialsManagerSpec: QuickSpec {
...
                 }
 
+                it("should store new credentials") {
+                    credentials = Credentials(accessToken: AccessToken, tokenType: TokenType, idToken: IdToken, refreshToken: RefreshToken, expiresIn: Date(timeIntervalSinceNow: -3600))
+                    _ = credentialsManager.store(credentials: credentials)
+                    waitUntil(timeout: 2) { done in
+                        credentialsManager.credentials { error = $0; newCredentials = $1
+                            expect(error).to(beNil())
+                            credentialsManager.credentials {
+                                expect($1!.accessToken) == NewAccessToken
```
lbalmaceda: here you use `$1!` but in the above test you refer to the same value as `newCredentials`. Why?

Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -175,18 +176,34 @@ class CredentialsManagerSpec: QuickSpec {
...
 
-                it("should yield new credentials") {
+                it("should yield new credentials, maintain refresh token") {
                     credentials = Credentials(accessToken: AccessToken, tokenType: TokenType, idToken: IdToken, refreshToken: RefreshToken, expiresIn: Date(timeIntervalSinceNow: -3600))
                     _ = credentialsManager.store(credentials: credentials)
                     waitUntil(timeout: 2) { done in
                         credentialsManager.credentials { error = $0; newCredentials = $1
                             expect(error).to(beNil())
-                            expect(newCredentials?.accessToken) == AccessToken
+                            expect(newCredentials?.accessToken) == NewAccessToken
```
lbalmaceda: `idToken` will change too, as the dates change. You should test that too.

Auth0/CredentialsManager.swift
cocojoe: Is to avoiding warnings, if I set it to a variable e.g. `foo = self...` Xcode would warn that it is not used. So you use `_ = `

Auth0/_ObjectiveManagementAPI.swift
cocojoe: In latest Swiftlint 0.22 they are no longer required.

```
 Line 47: SwiftLint rule 'function_parameter_count' did not trigger a violation in the disabled region. Please remove the disable command.
[03:09:56]: ▸ /Users/distiller/Auth0.swift/Auth0/OAuth2Grant.swift
```

Auth0/CredentialsManager.swift
lbalmaceda: Why set it anyway if it's not being used?

Auth0Tests/CredentialsManagerSpec.swift
cocojoe: It's not the same value, the second `credentialsManager.credentials {` is another closure so the $0,$1 are different from the first closure.

Auth0/CredentialsManager.swift
cocojoe: You get a warning if it's not set 😄

PR #147, opened 2017-09-07
Added invalid_credentials check for /oauth/token

cocojoe:


cocojoe:
I changed it back, although it's fine, I dunno it just looks better with latest 😄

PR #154, opened 2017-09-19
Added SFAuthenticationSession support in iOS 11

cocojoe:
- Adding SFAuthenticationSession support for iOS 11 (Default behaviour for iOS 11 Devices)
- Added clearSession using SFAuthenticationSession - Requires registration of callback schema in logouts URL in dashboard. (Added to README)


Auth0/SafariWebAuth.swift
```diff
@@ -100,6 +101,11 @@ class SafariWebAuth: WebAuth {
         return self
     }
 
+    func useLegacyAuthentication() -> Self {
```
hzalaz: Not a big fun but can't think of a better name 😄

cocojoe: Naming is the hardest part of development 😆

PR #163, opened 2017-10-17
Release 1.9.0

cocojoe:
**Added**
- Added SFAuthenticationSession support in iOS 11 [\#154](https://github.com/auth0/Auth0.swift/pull/154) ([cocojoe](https://github.com/cocojoe))

**Updgrade Notes**
If you are using the [clearSession](https://github.com/auth0/Auth0.swift/blob/master/Auth0/WebAuth.swift#L235) method in iOS 11, you will need to add your callback URL to **Allowed Logout URLs** in your [Auth0 Client Dashboard](https://manage.auth0.com/#/clients/)

### DIFF
--- a/App/Info.plist
+++ b/App/Info.plist
@@ -15,7 +15,7 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>1.8.0</string>
+	<string>1.9.0</string>
 	<key>CFBundleSignature</key>
 	<string>????</string>
 	<key>CFBundleURLTypes</key>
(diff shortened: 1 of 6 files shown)


lbalmaceda:
@cocojoe regarding the upgrading notes, shouldn't that be somewhere in the readme?

cocojoe:
Maybe or changelog?

lbalmaceda:
Honestly, I don't know how many people will check the changelog. I would if something gets broken from one release to the other, but not otherwise. Two thoughts:
- `returnTo` url should be customizable. If you use the same callback URL as in the login flow, how would you distinguish where is it coming from?
- I'd document that this url needs to be whitelisted in the "allowed logout url" section of the dashboard in the method docs like here https://github.com/auth0/Auth0.swift/blob/master/Auth0/WebAuth.swift#L232

cocojoe:
I'll make another PR with some extra docs, regarding the callback URL, it's like that for ease of setup, user just has to copy/paste in dashboard and it's done.
Code wise there is a store that stores the current callback session along with its callback, this decides the action, not the callback URL itself.

Issue #169, opened 2017-11-16
SafariWebAuth.clearSession ignores federated flag


srgray:
Running under iOS 11 (and Swift >= 3.2), SafariWebAuth.clearSession ignores the federated flag passed into the function. It does not add the 'federated' parameter to the logoutURL.

cocojoe:
Good spot. Thank you. #171

cocojoe:
Merged

PR #170, opened 2017-11-16
#169 Consider federated flag when building logoutURL for iOS 11

srgray:
Fix for Issue #169


cocojoe:
Thank you, I've optimised this into #171

PR #180, opened 2017-12-20
Updated Credentials Manager

cocojoe:
Renamed Touch Internals -> BioAuthentication
Deprecated enableTouchAuth method -> enableBiometrics for clarity
Expand README Docs
Update Circle CI Xcode Version

### DIFF
--- a/Auth0/CredentialsManager.swift
+++ b/Auth0/CredentialsManager.swift
@@ -46,18 +46,18 @@ public struct CredentialsManager {
     ///   - title: main message to display in TouchID prompt
     ///   - cancelTitle: cancel message to display in TouchID prompt (iOS 10+)
     ///   - fallbackTitle: fallback message to display in TouchID prompt after a failed match
-    @available(*, deprecated, message: "see enableBioAuth(withTitle title:, cancelTitle:, fallbackTitle:)")
+    @available(*, deprecated, message: "see enableBiometrics(withTitle title:, cancelTitle:, fallbackTitle:)")
     public mutating func enableTouchAuth(withTitle title: String, cancelTitle: String? = nil, fallbackTitle: String? = nil) {
-        self.enableBiometricAuth(withTitle: title, cancelTitle: cancelTitle, fallbackTitle: fallbackTitle)
+        self.enableBiometrics(withTitle: title, cancelTitle: cancelTitle, fallbackTitle: fallbackTitle)
     }
 
-    /// Enable Device Biometric Authentication for additional securtity during credentials retrieval.
+    /// Enable Biometric Authentication for additional securtity during credentials retrieval.
     ///
     /// - Parameters:
     ///   - title: main message to display when Touch ID is used
     ///   - cancelTitle: cancel message to display when Touch ID is used (iOS 10+)
     ///   - fallbackTitle: fallback message to display when Touch ID is used after a failed match
-    public mutating func enableBiometricAuth(withTitle title: String, cancelTitle: String? = nil, fallbackTitle: String? = nil) {
+    public mutating func enableBiometrics(withTitle title: String, cancelTitle: String? = nil, fallbackTitle: String? = nil) {
         self.bioAuth = BioAuthentication(authContext: LAContext(), title: title, cancelTitle: cancelTitle, fallbackTitle: fallbackTitle)
     }
 
--- a/Auth0/CredentialsManager.swift
+++ b/Auth0/CredentialsManager.swift
@@ -30,7 +30,7 @@ public struct CredentialsManager {
     private let storage = A0SimpleKeychain()
     private let storeKey = "credentials"
     private let authentication: Authentication
-    private var touchAuth: TouchAuthentication?
+    private var bioAuth: BioAuthentication?
 
     /// Creates a new CredentialsManager instance
     ///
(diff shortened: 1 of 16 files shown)


README.md
```diff
@@ -198,7 +200,10 @@ Store user credentials securely in the KeyChain.
 credentialsManager.store(credentials: credentials)
 ```
 
-#### Retrieve stored credentials and auto-renew if expired
+#### Retrieve stored credentials 
+
+Credentials will automatically be renewed (if expired) using the refresh token. The scope `offline_access` is required
+to ensure the refreshtToken is returned.
```
lbalmaceda: remove the new line after "required". Also typo. I would use "refresh token" instead.

README.md
```diff
@@ -207,6 +212,15 @@ credentialsManager.credentials { error, credentials in
 }
 ```
 
+#### Biometric authentication
+
+You can enable an additional level of user authentication before retrieving credentials using the biometric authentication
```
lbalmaceda: remove the new line after "biometric" and "authentication" (2nd)

Auth0/CredentialsManager.swift
```diff
@@ -40,14 +40,25 @@ public struct CredentialsManager {
...
+    @available(*, deprecated, message: "see enableBioAuth(withTitle title:, cancelTitle:, fallbackTitle:)")
     public mutating func enableTouchAuth(withTitle title: String, cancelTitle: String? = nil, fallbackTitle: String? = nil) {
-        self.touchAuth = TouchAuthentication(authContext: LAContext(), title: title, cancelTitle: cancelTitle, fallbackTitle: fallbackTitle)
+        self.bioAuth = BioAuthentication(authContext: LAContext(), title: title, cancelTitle: cancelTitle, fallbackTitle: fallbackTitle)
+    }
+
+    /// Enable Device Biometric Authentication for additional securtity during credentials retrieval.
+    ///
+    /// - Parameters:
+    ///   - title: main message to display in TouchID prompt
```
lbalmaceda: parameters talk about "touch" id. Should be changed to "bio" something

Auth0/CredentialsManager.swift
```diff
@@ -40,14 +40,25 @@ public struct CredentialsManager {
...
     }
 
-    /// Enable TouchID Authentication for additional securtity during credentials retrieval.
+    /// Enable Face ID / Touch ID Authentication for additional securtity during credentials retrieval.
     ///
     /// - Parameters:
     ///   - title: main message to display in TouchID prompt
     ///   - cancelTitle: cancel message to display in TouchID prompt (iOS 10+)
     ///   - fallbackTitle: fallback message to display in TouchID prompt after a failed match
+    @available(*, deprecated, message: "see enableBioAuth(withTitle title:, cancelTitle:, fallbackTitle:)")
```
lbalmaceda: so just deprecated because of the name change? No difference in the parameters??

Auth0/CredentialsManager.swift
cocojoe: Well it is only used during a touch prompt, it's not valid for face id. could be more verbose perhaps 'display when using touch id'

Auth0/CredentialsManager.swift
cocojoe: Yes, for the sake of clarity it should be the more generic `bio` as `touchid` might imply to some developers that it is ONLY for `touch`

cocojoe: Thoughts on this?

lbalmaceda: The new name is better but deprecating the method just because of the name doesn't sound right to me. Make this call 👍

lbalmaceda: Additional comment: Message says "see enableBioAuth" but the method is called "enableBiometricAuth". Also if these methods are going to have the same logic, make one call the other instead of copy-pasting the logic.

Auth0/CredentialsManager.swift
```diff
@@ -40,23 +40,23 @@ public struct CredentialsManager {
...
+        self.enableBiometricAuth(withTitle: title, cancelTitle: cancelTitle, fallbackTitle: fallbackTitle)
     }
 
     /// Enable Device Biometric Authentication for additional securtity during credentials retrieval.
     ///
     /// - Parameters:
-    ///   - title: main message to display in TouchID prompt
-    ///   - cancelTitle: cancel message to display in TouchID prompt (iOS 10+)
-    ///   - fallbackTitle: fallback message to display in TouchID prompt after a failed match
+    ///   - title: main message to display when Touch ID is used
```
lbalmaceda: aren't these messages also displayed when Face ID is used?

Auth0/CredentialsManager.swift
lbalmaceda: I don't know how this works as haven't tested myself. So when a face is expected no message is displayed, nothing that can be customized?

Auth0/CredentialsManager.swift
cocojoe: Made that change in the most recent commit.

Auth0/CredentialsManager.swift
cocojoe: No, it just comes up with this standard image / text.  https://support.apple.com/library/content/dam/edam/applecare/images/en_US/ios11-iphone-x-wallet-apple-pay-animation.gif

Issue #187, opened 2018-02-20
Client ID and redirect URL not present in clear session URL


Rypac:
When using the `WebAuth.clearSession` API to log out of an Auth0 session on iOS 11, the Client ID and redirect URL are not included in the clear session URL used to invoke the `SFAuthenticationSession`.

This issue is only present if the `federated` parameter is `false`.

cocojoe:
It should attempt to callback to the app, did you add the callback URL to the Allowed Logout URls?

https://github.com/auth0/auth0.swift#upgrade-notes

Rypac:
Thanks for the response @cocojoe.

Yes, the callback URL has been added to the Allowed Logout URLs. I believe I've found the issue as to why the redirect URL is not present in the clear session URL and have addressed it in #188.

In short, the change is:

``` diff
@@ -211,7 +211,8 @@ class SafariWebAuth: WebAuth {
    let returnTo = URLQueryItem(name: "returnTo", value: self.redirectURL?.absoluteString)
    let clientId = URLQueryItem(name: "client_id", value: self.clientId)
    var components = URLComponents(url: logoutURL, resolvingAgainstBaseURL: true)
-   components?.queryItems?.append(contentsOf: [returnTo, clientId])
+   let queryItems = components?.queryItems ?? []
+   components?.queryItems = queryItems + [returnTo, clientId]
    guard let clearSessionURL = components?.url, let redirectURL = returnTo.value else {
      return callback(false)
    }
```

When `federated` is not present in `logoutURL` the `components?.queryItems` is `nil`. Therefore appending the `returnTo` and `clientId` query items has no effect and will not be included in `clearSessionURL`.

---

I agree with your point about the `SFAuthenticationSession` being more intrusive. Another option I have been playing around with was removing the session data from the keychain and using the `prompt=login` query parameter that you mentioned to force a new login.

My only reservation about this approach is the session not being cleared from the Safari website data. Unless I am mistaken, will signing into the same and authenticating with Auth0 not just redirect you to the app using the session saved in Safari?

This would not be ideal if the user has previously logged out.

cocojoe:
You are right this is a bug, thanks. I'll raise a PR for this.

If you try the `prompt=login` the IdP should require re-authentication. So in the case of FB it will prompt for the password to be re-entered. You have to go with what works best for your use case.

Rypac:
[...]

PR #188, opened 2018-02-20
Fixed client ID and redirect URL query items not being passed in nonfederated clearSession()

Rypac:
As documented in issue #187, if no query items are present in the URL components when constructing
the clear session URL, the client ID and redirect URL query items are not added to the URL. This results in the web auth clear session never calling back successfully from the `SFAuthenticationSession`.

This change ensures that the client ID and redirect URL are always added to the clear session URL in the situation where the `federated` query item is not present.


cocojoe:
I couldn't find a good way to test either, which is a bit annoying now as would have caught this. One way would be to have a mock SFAuthenticationSession, however would need to open up the class to allow for passing of this object.

PR #201, opened 2018-05-31
Fix renewing credentials

mariancerny:
Credentials should be automatically be renewed (if expired) using the refresh token.

When credentials expired the `credentials()` method always returned
`.noCredentials` because `hasValid()` was `false`.

Fixed by replacing `hasValid()` with `renewableCredentials() != nil`.

### DIFF
--- a/Auth0/CredentialsManager.swift
+++ b/Auth0/CredentialsManager.swift
@@ -81,9 +81,7 @@ public struct CredentialsManager {
     /// - Returns: if there are valid and non-expired credentials stored
     public func hasValid() -> Bool {
         guard
-            let data = self.storage.data(forKey: self.storeKey),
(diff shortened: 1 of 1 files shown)


cocojoe:
I am not convinced, what is your test case?

In the `hasValid` method, you have the following logic.

https://github.com/auth0/Auth0.swift/blob/master/Auth0/CredentialsManager.swift#L89

So even if the AT has expired, as long as there is a RT available it will still return true as it can be renewed.

mariancerny:
You are right. This fix is bogus.

I was trying to fix an issue in our app and misread the condition

    return expiresIn > Date() || credentials.refreshToken != nil

with and (`&&`) instead of or (`||`). The problem was that our backend was misconfigured - it was not sending the refreshToken. When I later found out about the problem, I didn't reevaluate my "fix" in Auth0 library.

Sorry for wasting your time.

PR #203, opened 2018-06-15
Fix warnings if compiling as Swift 4.1 in Xcode 9.4.x. Fixes crash in tests

guykogus:
The unit tests don't all pass (at least on my computer) but they at least don't crash anymore.


cocojoe:
What problem does this address?  Please add some details of what you are trying to achieve?  

The only case I can see would be once Xcode 10 is gold and I would say after a year it's time to have a Swift 4.2+ version as it can't be long before Swift 3.x becomes retired as 2.x did.  This would result in a new Major version of the Auth0. library so as not to be a breaking change. 

This SDK is intentionally using Swift 3.2, this means I converted to Swift 4 then switched back to 3.2 to ensure it will work as a Framework in both Xcode 8.x and 9.x.

guykogus:
I like having no warnings in my projects. It's mainly complaining about the use of `flatMap`, which was deprecated for `compactMap`, so you need to update to Swift 4 for it.
There are also some warnings about deprecated methods but that's not something I'm gonna touch.

Libraries built in Swift 4 should be compatible with Swift 3, so it should be safe to upgrade, but I think it has to be 4.2 with 3.2 and 4.1 with 3.1, or something like that. I'm not really sure but I've definitely had problems with it in the past.

If you think this isn't a useful change that's fine, I'm happy to let it slide, just thought I'd offer it to you.

cocojoe:
How is your project setup? Would like to recreate out of interest.Please be detailed. Thx

guykogus:
I'm using cocoapods. It's very easy to reproduce.

1. Using the latest Xcode (9.4), create a new iOS Application project.
2. From the root directory run `pod init`
3. Add `pod 'Auth0'` to the `Podfile`
4. Run `pod install`
5. Build the project.

You'll see the following warnings:
auth0_errors

cocojoe:
Thanks I'm on 9.3, will download 9.4 and take a look. However, any "fix" still needs to be compatible in 3.2 for now. So may require wrapping code in swift version checks, will take look.

guykogus:
Good call, I just added a check to make sure that `compactMap` is available.

cocojoe:
Can you build/test in both 3.2 and 4.1 to test and then set back to 3.2 so not a breaking change.

guykogus:
They both work. Reverting.

guykogus:
I've updated the topic and description to match the current changes.

guykogus:
Btw, if you choose to approve I recommend doing a Squash and Merge, rather than a standard Merge... 😬

cocojoe:
What version of Cocoapods are you using?

I would have thought the swift-version of the library would be respected by now, as this came up a while ago in another repo. https://github.com/auth0/Lock.swift/issues/470  As it's a bit frustrating to be changed to a different version that the framework never said it was written 😄

guykogus:
I'm always using the latest (non-beta) tools, so I'm on cocoapods 1.5.3. I think the reason that it's not respecting the `.swift-version` file is because on Xcode 9.4 we have Swift 3.3, not 3.2. Also, I think that 4.1 is backwards-compatible with 3.2, as in it's not source-breaking, so there might be some warnings like the one I'm trying to fix here, but there's no reason for cocoapods to block the project from being imported.

cocojoe:
I'm running 9.4 now so will take a look today. However, looks fine at a glance 👍

cocojoe:
@guykogus I checked it out but one of the changes broke some tests so i moved to #207 If you want to take a look. Thx

guykogus:
What was the breaking change? The only things that we have different are the `// swiftlint:disable` calls and the `#if`, and I'm pretty sure they're both right.

cocojoe:
Broke all AuthenticationSpec tests https://github.com/auth0/Auth0.swift/pull/203/files#diff-61e1051d511212d6b24d4d4e6c70a53aR280

guykogus:
I was sure I added that for a reason. Maybe I tried to run the tests in `Release` to test something and it failed so I added that `#if`, but the tests don't include the `DEBUG` flag.

PR #206, opened 2018-07-16
Enable Credentials Manager for tvOS and Mac Platforms

cocojoe:
BioMetrics only available for iOS Platform Only
Remove restriction in CocoaPod Spec
Added tvOS, Mac Apps for Hosted Testing (Required for Keychain)

### DIFF
--- a/Auth0.podspec
+++ b/Auth0.podspec
@@ -19,6 +19,27 @@ web_auth_files = [
   'Auth0/BioAuthentication.swift'
 ]
 
+watchos_exclude_files = [
+'Auth0/_ObjectiveWebAuth.swift',
+'Auth0/ControllerModalPresenter.swift',
+'Auth0/OAuth2Grant.swift',
+'Auth0/AuthTransaction.swift',
+'Auth0/TransactionStore.swift',
+'Auth0/WebAuth.swift',
+'Auth0/WebAuthError.swift',
+'Auth0/SafariWebAuth.swift',
+'Auth0/AuthSession.swift',
+'Auth0/SafariSession.swift',
+'Auth0/SafariAuthenticationSession.swift',
+'Auth0/SafariAuthenticationCallback.swift',
+'Auth0/SilentSafariViewController.swift',
+'Auth0/NativeAuth.swift',
+'Auth0/AuthProvider.swift',
+'Auth0/BioAuthentication.swift',
+'Auth0/CredentialsManagerError.swift',
+'Auth0/CredentialsManager.swift'
+]
+
 Pod::Spec.new do |s|
   s.name             = 'Auth0'
   s.version          = version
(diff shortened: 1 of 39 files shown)


Auth0.podspec
```diff
@@ -16,11 +16,30 @@ web_auth_files = [
   'Auth0/SilentSafariViewController.swift',
   'Auth0/NativeAuth.swift',
   'Auth0/AuthProvider.swift',
-  'Auth0/CredentialsManager.swift',
```
lbalmaceda: this are referenced in `s.osx.exclude_files`. Why did you remove them?

Auth0.podspec
```diff
@@ -16,11 +16,30 @@ web_auth_files = [
...
-  'Auth0/CredentialsManagerError.swift',
   'Auth0/BioAuthentication.swift'
 ]
 
+watchos_exclude_files = [
```
lbalmaceda: this looks like `web_auth_files` plus 2 entries. Is there a nicer way to define this array? i.e. joining the already defined base array.

OAuth2TV/AppDelegate.swift
```diff
@@ -0,0 +1,46 @@
...
+    var window: UIWindow?
+
+
+    func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
+        // Override point for customization after application launch.
+        return true
+    }
+
+    func applicationWillResignActive(_ application: UIApplication) {
+        // Sent when the application is about to move from active to inactive state. This can occur for certain types of temporary interruptions (such as an incoming phone call or SMS message) or when the user quits the application and it begins the transition to the background state.
```
lbalmaceda: > (such as an incoming phone call or SMS message)

how can this happen on a TV?

OAuth2TV/ViewController.swift
```diff
@@ -0,0 +1,25 @@
...
+//  ViewController.swift
+//  OAuth2TV
+//
+//  Created by Martin on 16/07/2018.
+//  Copyright © 2018 Auth0. All rights reserved.
+//
+
+import UIKit
+
+class ViewController: UIViewController {
```
lbalmaceda: what's the point on keeping this class if it only overrides 2 methods that call the super? as opposed to  `OAuth2Mac/ViewController.swift` which does change one of the methods logic.

Auth0.podspec
cocojoe: Enabling it for tvOS also allows it for Mac

Auth0.podspec
cocojoe: Sadly not, i've tried joining arrays etc it will just throw out an error in cocoapods when linting the lib.

OAuth2TV/AppDelegate.swift
cocojoe: *shrug* this is default boiler plate apple app creation.

OAuth2TV/ViewController.swift
cocojoe: Again this is Apple's boilerplate application creation.

cocojoe:
@lbalmaceda these are pre-processor macros, not run-time.  There is no need to create a pointless no-op method, in the above the method will not be visible in any way when building a tvOS app. From the perspective of the developer they only see the methods available to that platform. 

I've tested locally for tvOS, fastlane doesn't support tvOS so I can't easily add to the current CI setup.
The tenant who requested this feature has also tested and confirmed all working for their use case.

PR #207, opened 2018-07-20
Fix Swift 4.1 Warning

cocojoe:
This fixes a compiler warning if the project is compiled as 4.1.

`flatMap` is Swift 3.x
`compactMap` is only available in 4.1, `flatMap` will produce a compiler warning in Swift 4.1

The ! to ? came up in the original PR, although I think this was caused by another change in the original PR that I reverted. However, the style in this PR is more consistent with the rest of the spec.

The project should be 3.x, seems to have accidentally been update in the last PR, so coincidentally this fixes it.


cocojoe:
Kind of strange 9.4.1 throws "Gem::FilePermissionError: You don't have write permissions for the"

guykogus:
Could you use `bundle install --development`?

cocojoe:
It should work anyway, feels more likely image is broken, will roll back to 9.4.0

PR #208, opened 2018-07-24
Added support for custom Keychain key in Credentials Manager

danielphillips:
Solution for support ticket 45174.

This allows us to use multiple `CredentialsManager` instances in order to support multiple accounts in one app.

### DIFF
--- a/Auth0/CredentialsManager.swift
+++ b/Auth0/CredentialsManager.swift
@@ -30,7 +30,7 @@ import LocalAuthentication
 public struct CredentialsManager {
 
     private let storage = A0SimpleKeychain()
-    private let storeKey = "credentials"
+    private let storeKey: String
     private let authentication: Authentication
     #if os(iOS)
     private var bioAuth: BioAuthentication?
@@ -40,7 +40,9 @@ public struct CredentialsManager {
     ///
     /// - Parameters:
     ///   - authentication: Auth0 authentication instance
-    public init(authentication: Authentication) {
+    ///   - storeKey: Key used to store user credentials in the keychain, defaults to "credentials"
+    public init(authentication: Authentication, storeKey: String = "credentials") {
+        self.storeKey = storeKey
         self.authentication = authentication
     }
 
--- a/Auth0Tests/CredentialsManagerSpec.swift
+++ b/Auth0Tests/CredentialsManagerSpec.swift
@@ -71,7 +71,83 @@ class CredentialsManagerSpec: QuickSpec {
             it("should fail to clear credentials") {
                 expect(credentialsManager.clear()).to(beFalse())
             }
+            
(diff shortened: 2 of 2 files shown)


Auth0/CredentialsManager.swift
```diff
@@ -40,7 +40,8 @@ public struct CredentialsManager {
     ///
     /// - Parameters:
     ///   - authentication: Auth0 authentication instance
-    public init(authentication: Authentication) {
+    public init(authentication: Authentication, storeKey: String = "credentials") {
```
cocojoe: Add Doc Header

danielphillips: Done.

danielphillips:
I've added two tests proving that:
1. Two instances using a different store key store to different places in the keychain and don't interfere with each other.
2. Two instances using the same store key experience shared credentials.

I'm not best pleased with how I've laid out the tests, any feedback on this would be nice.

Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -71,7 +71,63 @@ class CredentialsManagerSpec: QuickSpec {
             it("should fail to clear credentials") {
                 expect(credentialsManager.clear()).to(beFalse())
             }
+            
+        }
+        
+        describe("multi instances of credentials manager") {
```
cocojoe: I'm not as bothered about tests as long as they work. However, they would look cleaner to use closure shorthand like https://github.com/danielphillips/Auth0.swift/blob/a78a4018722c9202ee5d1a75c2608f5c00f855ac/Auth0Tests/CredentialsManagerSpec.swift#L263

```
credentialsManager.credentials { 
   expect($1?.accessToken) == AccessToken
   expect($1?.idToken) == IdToken
  done()
}
```

Cut down on the extra params.

cocojoe:
@danielphillips just merged another PR into master, can you rebase this please.

Auth0Tests/CredentialsManagerSpec.swift
danielphillips: Good point thank you. Done that now.

danielphillips:
I'm rebased and I've addressed the feedback.

Auth0/CredentialsManager.swift
```diff
@@ -40,7 +40,9 @@ public struct CredentialsManager {
     ///
     /// - Parameters:
     ///   - authentication: Auth0 authentication instance
-    public init(authentication: Authentication) {
+    ///   - storeKey: String key to store user credentials in the keychain, defaults to "credentials"
```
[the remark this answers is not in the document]
cocojoe: Just this to address and I'll approve, thx

PR #211, opened 2018-07-31
Update README.md

mmaddex:
The [Migration Guide](https://auth0.com/docs/migrations/guides/calling-api-with-idtokens) to discourage the use of ID tokens when authorizing some calls to API v2 mentions an as of now open-ended grace period that started March 31, 2018. However, it seems to me that we should update the documentation to encourage use of `accessToken` over `idToken` when calling management api.

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,7 +278,7 @@ Auth0
 
 ```swift
 Auth0
-    .users(token: idToken)
+    .users(token: accessToken)
     .get("user identifier", fields: ["user_metadata"], include: true)
     .start { result in
         switch result {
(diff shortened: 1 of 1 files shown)


cocojoe:
I think it may make more sense now to reference the QuickStart that deals with MetaData as now you must specify the correct scopes to be able to do this.

https://auth0.com/docs/quickstart/native/ios-swift/03-user-sessions#managing-metadata

cocojoe:
Have added a ticket in backlog to replace with Quickstart info on the subject.

PR #218, opened 2018-09-08
Fix sample code warning in README

kangaroo5383:
This way after the `guard`, `credentials` is properly unwrapped.


cocojoe:
Thanks

Issue #220, opened 2018-09-13
Compiler Issue in Xcode 10 Discussion


cocojoe:
There is a known issue when compiling in Xcode 10, the Swift 4.2 compiler introduced a bug (which has been recently patched but did not make it into Xcode 10 GM). 
This is a problem today but will be fixed in a later Xcode 10 release. Awkward...

The issue results in a **Segmentation Fault: 11** caused by when a `mutating function` returns `Self`.

Reference: 
https://github.com/apple/swift/pull/18937
https://github.com/apple/swift/pull/19038

Affected Code:
https://github.com/auth0/Auth0.swift/blob/master/Auth0/Loggable.swift#L40
```
mutating func using(logger: Logger) -> Self {...}
```

https://github.com/auth0/Auth0.swift/blob/master/Auth0/Loggable.swift#L52
```
 mutating func logging(enabled: Bool) -> Self {...}
```

Suggested `workarounds` include:

1. Changing _Self_ to _Loggable_, this breaks method chaining.

It is a Breaking Change as any chaining use is now broken (but it's easy to fix). However, the BC is really in the compiler so perhaps just clear notes would be okay in the release?

2. Can't remove mutating as inherited in protocols & structs.

cocojoe:
Have taken a deep breath now.  
Will roll out with creating a new instance, which may be an improvement for immutability. No BC either.

```
func using(logger: Logger) -> Self {
        var loggable = self
        loggable.logger = logger
        return loggable
    }
```

```
func logging(enabled: Bool) -> Self {
        var loggable = self
        if enabled {
            loggable.logger = DefaultLogger()
        } else {
            loggable.logger = nil
        }
        return loggable
    }
```

cocojoe:
I'll raise a PR for this tomorrow and update Lock.

OrestNazarewycz:
Thanks for being so responsive on this one, really appreciate it @cocojoe !!

cocojoe:
PR In Progress: https://github.com/auth0/Auth0.swift/pull/221

cocojoe:
Released

PR #221, opened 2018-09-13
Xcode 10 Update

cocojoe:
Updated Loggable Protocol & Extension to workaround 4.2 compiler issue
Removed test due to potential issue in Nimble
Updated Dependencies
Added support for compiling as 4.2

### DIFF
--- a/App/ViewController.swift
+++ b/App/ViewController.swift
@@ -35,14 +35,14 @@ class ViewController: UIViewController {
     }
 
     @IBAction func startOAuth2(_ sender: Any) {
-        var auth0 = Auth0.webAuth()
+        let auth0 = Auth0.webAuth()
         auth0
             .logging(enabled: true)
             .start(onAuth)
     }
 
     @IBAction func startGoogleOAuth2(_ sender: Any) {
-        var auth0 = Auth0.webAuth()
+        let auth0 = Auth0.webAuth()
         auth0
             .logging(enabled: true)
             .connection("google-oauth2")
@@ -50,7 +50,7 @@ class ViewController: UIViewController {
     }
 
     @IBAction func startTokenGoogleOAuth2(_ sender: Any) {
-        var auth0 = Auth0.webAuth()
+        let auth0 = Auth0.webAuth()
         auth0
             .logging(enabled: true)
             .connection("google-oauth2")
@@ -59,7 +59,7 @@ class ViewController: UIViewController {
     }
 
     @IBAction func startIDTokenGoogleOAuth2(_ sender: Any) {
-        var auth0 = Auth0.webAuth()
+        let auth0 = Auth0.webAuth()
         auth0
             .logging(enabled: true)
             .connection("google-oauth2")
@@ -69,7 +69,7 @@ class ViewController: UIViewController {
     }
 
     @IBAction func startTokenIDTokenGoogleOAuth2(_ sender: Any) {
-        var auth0 = Auth0.webAuth()
+        let auth0 = Auth0.webAuth()
         auth0
             .logging(enabled: true)
             .connection("google-oauth2")
@@ -79,7 +79,7 @@ class ViewController: UIViewController {
     }
 
     @IBAction func startCodeIDTokenGoogleOAuth2(_ sender: Any) {
-        var auth0 = Auth0.webAuth()
+        let auth0 = Auth0.webAuth()
         auth0
             .logging(enabled: true)
             .connection("google-oauth2")
--- a/App/ViewController.swift
+++ b/App/ViewController.swift
@@ -35,23 +35,20 @@ class ViewController: UIViewController {
     }
 
     @IBAction func startOAuth2(_ sender: Any) {
-        let auth0 = Auth0.webAuth()
-        auth0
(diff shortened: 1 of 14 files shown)


Auth0/SafariWebAuth.swift
```diff
@@ -23,6 +23,12 @@
 import UIKit
 import SafariServices
 
+#if swift(>=4.2)
+public typealias OpenURLOptionsKey = UIApplication.OpenURLOptionsKey
```
guykogus: I don't think setting a public `typealias` is a good idea as it can conflict with others. Perhaps it should be inside a `struct` or something, e.g.
```Swift
struct Auth0Application {
   #if swift
   ...
   #endif
}
```
Would that be possible? That way we can keep backwards compatibility without risking others conflicting, in case they try to use the same trick.

 It also doesn't make sense to define it in this class instead of in `Auth0.swift`, for example.

Thoughts?

App/ViewController.swift
```diff
@@ -35,22 +35,22 @@ class ViewController: UIViewController {
     }
 
     @IBAction func startOAuth2(_ sender: Any) {
-        var auth0 = Auth0.webAuth()
+        let auth0 = Auth0.webAuth()
```
guykogus: If you've made this non-variable you can skip the whole `let auth0 =` bit and just go straight into
```Swift
Auth0
   .webAuth()
   .logging...
```

Auth0/SafariWebAuth.swift
cocojoe: It has to be public due to some of the methods that include it as a param type being public. I could prefix it to avoid an unlikely collision e.g. `Auth0URLOptionsKey`

Auth0.swift doesn't include UIKit as it's used cross platform, UIKit is for iOS and the only areas that use this are in the files that are only included for iOS.

App/ViewController.swift
cocojoe: Yeah good point, the sample App isn't a priority. However, it will look cleaner 👍

Auth0Tests/ResponseSpec.swift
```diff
@@ -70,11 +70,11 @@ class ResponseSpec: QuickSpec {
                 expect(try? response.result()).to(beNil())
             }
 
-            it("should fail with invalid JSON") {
-                let data = "A simple String".data(using: .utf8)
-                let response = Response<AuthenticationError>(data: data, response: http(200), error: nil)
-                expect(try? response.result()).to(beNil())
-            }
+//            it("should fail with invalid JSON") {
```
guykogus: Why is this failing?

cocojoe: Well it's a strange error, it expects `nil` but gets `Optional(nil)` I feel it's most likely an issue in Quick/Nimble magic while libs mature for Xcode 10.

Auth0/WebAuth.swift
```diff
@@ -22,6 +22,12 @@
 
 import UIKit
 
+#if swift(>=4.2)
+public typealias A0URLOptionsKey = UIApplication.OpenURLOptionsKey
```
lbalmaceda: Fix indentation

Auth0/SafariAuthenticationCallback.swift
```diff
@@ -40,7 +40,7 @@ class SafariAuthenticationSessionCallback: AuthTransaction {
         self.authSession?.start()
     }
 
-    func resume(_ url: URL, options: [UIApplicationOpenURLOptionsKey: Any]) -> Bool {
+    func resume(_ url: URL, options: [A0URLOptionsKey: Any]) -> Bool {
```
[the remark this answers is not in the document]
cocojoe: It doesn't really affect the signature, it looks different in source but the compiler looks to the `typealias` to see what type it really is and will then use that.  So no changes for existing code.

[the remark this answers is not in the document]
cocojoe: I will also be testing Lock against this before any release.

Auth0Tests/ResponseSpec.swift
guykogus: Minor correction, `if let foo = Optional<Any>.none as? Any {` won't enter the block. It turns out that this issue is related to generic functions being casted to `Any`.

cocojoe: This is a change between 4.1 and 4.2? You have some source on this change. Thx

cocojoe: Thanks @guykogus  I don't feel 100% comfortable with 4.2 tbh. Let me reach out on this.

Auth0/SafariAuthenticationCallback.swift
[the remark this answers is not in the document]
cocojoe: @hzalaz do you feel that's a deal breaker for this change?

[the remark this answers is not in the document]
hzalaz: @cocojoe for me no, kinda forced by xcode so there is no other way than making a major but imho is not worth a major for this change alone.

cocojoe:
#220

PR #226, opened 2018-09-25
Added Custom Domains information to README

cocojoe:


README.md
```diff
@@ -274,6 +274,12 @@ Auth0
     }
 ```
 
+### Custom Domains
+
+If you are using the [Custom Domains](https://auth0.com/docs/custom-domains) feature, it's important you do not use the custom domain value when configuring the `.audience`. Please use the Auth0 domain specified for your Application in the [Auth0 Dashboard](https://auth0.com/docs/dashboard).
```
lbalmaceda: I find it confusing. After checking the [docs](https://auth0.com/docs/custom-domains#-service-not-found-) (which you can refer) I suggest:
  
If you are using the [Custom Domains](https://auth0.com/docs/custom-domains) feature and require to request the `/userinfo` or `/api/v2/` (I'm not 100% sure about this one) audience, please use the Auth0 domain specified for your Application in the dashboard.

Example:  `.audience("https://{YOUR_AUTH0_DOMAIN}/userinfo")`

cocojoe: good point, I was only thinking about Auth0 audiences.

README.md
```diff
@@ -274,6 +274,13 @@ Auth0
     }
 ```
 
+### Custom Domains
+
+If you are using the [Custom Domains](https://auth0.com/docs/custom-domains) feature and need to use an Auth0 endpoint
```
lbalmaceda: It's still confusing. Now I read: "If you need to call this endpoint, use the auth0 domain". So I understand: "If I need to call `acme.com/userinfo` I need to call instead `acme12.auth0.com/userinfo` and that's not true. AFAIK using a custom domain means using my own url namespace even for authorize and userinfo calls, as long as the AUDIENCE that it's being requested (IF ANY!) is still pointing to the auth0 domain, like the example line you had before.

Also, why the new line at the end?

cocojoe: It does say an 'Auth0' endpoint

PR #239, opened 2018-10-23
Use ASWebAuthenticationSession on iOS 12

### LINKED ISSUE #238: SFAuthenticationSession is now deprecated in iOS 12, ASWebAuthenticationSession should be used instead

martinoconnor:
Fixes #238

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ jobs:
 
     steps:
       - checkout
-      - run: | 
+      - run: |
           brew install swiftlint
           bundle install --without=developments
           grep -lR "shouldUseLaunchSchemeArgsEnv" *.* --null | xargs -0 sed -i '' -e 's/shouldUseLaunchSchemeArgsEnv = "YES"/shouldUseLaunchSchemeArgsEnv = "YES" codeCoverageEnabled = "YES"/g'
@@ -28,7 +28,7 @@ jobs:
           name: Pod Lib Lint
           command: bundle exec fastlane ios pod_lint
       - run: |
-          bash <(curl -s https://codecov.io/bash) -J 'Auth0' 
+          bash <(curl -s https://codecov.io/bash) -J 'Auth0'
       - save_cache:
           key: dependency-cache
           paths:
--- a/Auth0/SafariAuthenticationCallback.swift
+++ b/Auth0/SafariAuthenticationCallback.swift
@@ -30,7 +30,7 @@ class SafariAuthenticationSessionCallback: AuthTransaction {
 
     var state: String?
     var callback: (Bool) -> Void = { _ in }
-    
+
     private var authSession: NSObject?
 
     init(url: URL, schemeURL: String, callback: @escaping (Bool) -> Void) {
--- a/Auth0/SafariAuthenticationCallback.swift
+++ b/Auth0/SafariAuthenticationCallback.swift
@@ -21,6 +21,7 @@
 // THE SOFTWARE.
 
 import UIKit
+import AuthenticationServices
 import SafariServices
 
 #if swift(>=3.2)
@@ -28,16 +29,27 @@ import SafariServices
 class SafariAuthenticationSessionCallback: AuthTransaction {
 
     var state: String?
-    var authSession: SFAuthenticationSession?
     var callback: (Bool) -> Void = { _ in }
+    
+    private var authSession: NSObject?
 
     init(url: URL, schemeURL: String, callback: @escaping (Bool) -> Void) {
         self.callback = callback
-        self.authSession = SFAuthenticationSession(url: url, callbackURLScheme: schemeURL) { [unowned self] url, _ in
-            self.callback(url != nil)
-            TransactionStore.shared.clear()
+        if #available(iOS 12.0, *) {
+            let authSession = ASWebAuthenticationSession(url: url, callbackURLScheme: schemeURL) { [unowned self] url, _ in
+                self.callback(url != nil)
+                TransactionStore.shared.clear()
+            }
+            self.authSession = authSession
+            authSession.start()
+        } else {
+            let authSession = SFAuthenticationSession(url: url, callbackURLScheme: schemeURL) { [unowned self] url, _ in
+                self.callback(url != nil)
+                TransactionStore.shared.clear()
+            }
+            self.authSession = authSession
+            authSession.start()
         }
-        self.authSession?.start()
     }
 
     func resume(_ url: URL, options: [A0URLOptionsKey: Any]) -> Bool {
--- a/Auth0/SafariAuthenticationSession.swift
+++ b/Auth0/SafariAuthenticationSession.swift
@@ -22,31 +22,45 @@
 
 import UIKit
 import SafariServices
+import AuthenticationServices
 
 #if swift(>=3.2)
 @available(iOS 11.0, *)
 class SafariAuthenticationSession: AuthSession {
-
-    typealias FinishSession = (Result<Credentials>) -> Void
-
-    var authSession: SFAuthenticationSession?
-    let authorizeURL: URL
+    
+    private var authSession: NSObject?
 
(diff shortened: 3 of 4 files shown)


cocojoe:
Thanks for the PR @martinoconnor

Auth0/SafariWebAuth.swift
```diff
@@ -120,7 +120,11 @@ class SafariWebAuth: WebAuth {
         let authorizeURL = self.buildAuthorizeURL(withRedirectURL: redirectURL, defaults: handler.defaults, state: state)
 
         #if swift(>=3.2)
-        if #available(iOS 11.0, *), self.authenticationSession {
+        if #available(iOS 12.0, *), self.authenticationSession {
+            let session = ASAuthenticationSession(authorizeURL: authorizeURL, redirectURL: redirectURL, state: state, handler: handler, finish: callback, logger: logger)
+            logger?.trace(url: authorizeURL, source: "AuthenticationServicesAuthenticationSession")
+            self.storage.store(session)
+        } else if #available(iOS 11.0, *), self.authenticationSession {
```
cocojoe: For iOS 11 you still have to use SFAuthenticationSession as per Apple guidelines. So there are 3 tiers of support iOS >=12, iOS 11, iOS <11.0

Auth0/ASAuthenticationSession.swift
```diff
@@ -0,0 +1,52 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import UIKit
+import AuthenticationServices
+
+#if swift(>=3.2)
+@available(iOS 12.0, *)
+class ASAuthenticationSession: AuthSession {
```
cocojoe: Do we really need a whole new class or could we `typealias` the correct method based upon iOS?

martinoconnor: I think the class is probably justified as the ASWebAuthenticationSession or SFAuthenticationSession is stored in a member variable of the respective class. Then we have to worry about the error type in the handler callback which is related to the corresponding session type

martinoconnor:
Looks like the build failure is due to CocoaPods/CocoaPods#7708 so will have to wait until 1.6 of CocoaPods comes out of beta

Auth0/SafariWebAuth.swift
cocojoe: My mistake, saw the word `Safari`, naming these classes/methods is quite tricky 😆

cocojoe:
#240 This will fix the pod lint issue.

Please can you rebase this PR against master. Thx

Auth0/ASAuthenticationSession.swift
martinoconnor: I have moved the changes inside the original SafariAuthenticationSession class

cocojoe:
@martinoconnor what about https://github.com/auth0/Auth0.swift/blob/master/Auth0/WebAuth.swift#L244 ?

Auth0/SafariAuthenticationCallback.swift
```diff
@@ -22,22 +22,28 @@
...
 import UIKit
 import SafariServices
+import AuthenticationServices
 
 #if swift(>=3.2)
 @available(iOS 11.0, *)
 class SafariAuthenticationSessionCallback: AuthTransaction {
 
     var state: String?
-    var authSession: SFAuthenticationSession?
```
cocojoe: By removing this, there is nothing retaining the session alive, so as soon as you accept the `Auth0.com` dialog it will fail and never launch the browser.

`Attempting to load the view of a view controller while it is deallocating is not allowed and may result in undefined behavior (<SFAuthenticationViewController: 0x7f851c82b600>)`

Auth0/SafariAuthenticationCallback.swift
```diff
@@ -22,22 +22,28 @@
...
     var callback: (Bool) -> Void = { _ in }
 
     init(url: URL, schemeURL: String, callback: @escaping (Bool) -> Void) {
         self.callback = callback
-        self.authSession = SFAuthenticationSession(url: url, callbackURLScheme: schemeURL) { [unowned self] url, _ in
```
cocojoe: Related to previous comment.

Auth0/SafariAuthenticationSession.swift
```diff
@@ -22,31 +22,39 @@
...
 
-    typealias FinishSession = (Result<Credentials>) -> Void
-
-    var authSession: SFAuthenticationSession?
-    let authorizeURL: URL
-
     init(authorizeURL: URL, redirectURL: URL, state: String? = nil, handler: OAuth2Grant, finish: @escaping FinishSession, logger: Logger?) {
-        self.authorizeURL = authorizeURL
         super.init(redirectURL: redirectURL, state: state, handler: handler, finish: finish, logger: logger)
-        self.authSession = SFAuthenticationSession(url: self.authorizeURL, callbackURLScheme: self.redirectURL.absoluteString) { [unowned self] in
```
cocojoe: Retention doesn't appear to be an issue here, although I'm sure it was hence the use of the `authSession` property.
It's difficult to test these behaviour in unit tests.

martinoconnor: Looks like SFAuthenticationSession requires a strong reference, as per the apple docs

> Note
> 
> Ensure that there is a strong reference to the SFAuthenticationSession instance when the session is in progress.

I don't see a similar note for ASWebAuthenticationSession

martinoconnor: If ASWebAuthenticationSession is deallocated, the permissions dialog is dismissed

Auth0/SafariAuthenticationCallback.swift
cocojoe: AnyObject not work here? Doesn't look right pulling in ObjC

.circleci/config.yml
```diff
@@ -40,4 +40,4 @@ workflows:
   version: 2
   build-test-report:
     jobs:
-      - build-and-test
```
cocojoe: It would be good not to make unnecessary formatting changes thanks

Auth0/SafariAuthenticationCallback.swift
cocojoe: Same as previous `NSObject` comment

fastlane/Scanfile
```diff
@@ -1,5 +1,5 @@
 scheme "Auth0.iOS"
-devices ["iPhone 8"]
+devices ["iPhone 8 (11.2)", "iPhone 8 (12.0)", "iPhone 7 (10.3.1)"]
```
cocojoe: I see what you're doing for your own tests, but please don't include in this PR. At best it would be in a different PR, although I intend to replace with multiple platform tests, which i recently added to Lock.swift.

Auth0/SafariAuthenticationSession.swift
```diff
@@ -22,31 +22,45 @@
...
-            guard $1 == nil, let callbackURL = $0 else {
-                if case SFAuthenticationError.canceledLogin = $1! {
-                    self.finish(.failure(error: WebAuthError.userCancelled))
-                } else {
-                    self.finish(.failure(error: $1!))
+        if #available(iOS 12.0, *) {
+            let authSession = ASWebAuthenticationSession(url: authorizeURL, callbackURLScheme: self.redirectURL.absoluteString) { [unowned self] in
```
cocojoe: Seems like a lot of duplication of code in side these blocks, could we refactor this down to one block and a typealias?

cocojoe:
Also please rebase again master, thanks.

martinoconnor:
Apologies for the delay, work has been busy lately, but will make the changes when I'm free

cocojoe:
@martinoconnor no problem, let me know if you need a hand on finishing this up.

cocojoe:
Moved to #245

PR #242, opened 2018-11-09
Add Multiple Platform CI

cocojoe:
Update Code for 4.2

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,7 +97,6 @@ jobs:
           brew install swiftlint
           bundle install --without=development
           grep -lR "shouldUseLaunchSchemeArgsEnv" *.* --null | xargs -0 sed -i '' -e 's/shouldUseLaunchSchemeArgsEnv = "YES"/shouldUseLaunchSchemeArgsEnv = "YES" codeCoverageEnabled = "YES"/g'
-          security unlock-keychain -u   
       - run:
           name: Bootstrap
           command: carthage bootstrap --platform mac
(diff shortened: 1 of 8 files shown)


.circleci/config.yml
```diff
@@ -24,20 +22,127 @@ jobs:
             SCHEME: Auth0.iOS
             DEVICE: iPhone 8
             FASTLANE_EXPLICIT_OPEN_SIMULATOR: 2
-      - run:
-          name: Pod Lib Lint
-          command: bundle exec fastlane ios pod_lint
       - run: |
-          bash <(curl -s https://codecov.io/bash) -J 'Auth0' 
+          bash <(curl -s https://codecov.io/bash) -J 'Auth0'
```
lbalmaceda: Should we also upload the coverage for the rest of the build variants? Why do you need to pass the package name?

.circleci/config.yml
```diff
@@ -1,18 +1,16 @@
 version: 2
 jobs:
-  build-and-test:
+  build-and-test-iOS-swift-4.2:
```
lbalmaceda: I'd rename it to something shorter. They are all builds running tests. 
- `iOS-swift-4.2-build`
- `build-iOS-swift-4.2`

fastlane/Fastfile
```diff
@@ -9,7 +9,7 @@ platform :ios do
   desc "Installs dependencies using Carthage"
   lane :dependencies do |options|
     action = options[:action]
-    carthage(use_binaries: false, command: action, cache_builds: true, platform: 'all')
+    carthage(use_binaries: false, command: action, cache_builds: true, platform: 'iOS')
```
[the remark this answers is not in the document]
cocojoe: Limiting build to the platform in question, otherwise it will build every platform even though it's only using 1. The previous option of 'all' was wrong. I

.circleci/config.yml
cocojoe: I've ran into issues in the past with multiple codecov, I don't think it understands what to do with it for example. MacOS doesn't have WebAuth so if that report is sent after the iOS one codecov will report lots of missing codecov.

.circleci/config.yml
cocojoe: Will think about it ;)

PR #245, opened 2018-11-15
Added ASWebAuthenticationSession Support iOS12

cocojoe:
### Changes

Follow up to #239 

Internal changes to enable `ASWebAuthenticationSession` in iOS 12 and maintain support for Xcode 9.

### References

#238 

### Testing

### Checklist

### DIFF
--- a/Auth0/SafariAuthenticationCallback.swift
+++ b/Auth0/SafariAuthenticationCallback.swift
@@ -30,7 +30,7 @@ class SafariAuthenticationSessionCallback: AuthTransaction {
 
     var state: String?
     var callback: (Bool) -> Void = { _ in }
-    
+
     private var authSession: NSObject?
 
     init(url: URL, schemeURL: String, callback: @escaping (Bool) -> Void) {
--- a/Auth0/SafariAuthenticationCallback.swift
+++ b/Auth0/SafariAuthenticationCallback.swift
@@ -21,6 +21,7 @@
 // THE SOFTWARE.
 
 import UIKit
+import AuthenticationServices
 import SafariServices
 
 #if swift(>=3.2)
@@ -28,16 +29,27 @@ import SafariServices
 class SafariAuthenticationSessionCallback: AuthTransaction {
 
     var state: String?
-    var authSession: SFAuthenticationSession?
     var callback: (Bool) -> Void = { _ in }
+    
+    private var authSession: NSObject?
 
     init(url: URL, schemeURL: String, callback: @escaping (Bool) -> Void) {
         self.callback = callback
-        self.authSession = SFAuthenticationSession(url: url, callbackURLScheme: schemeURL) { [unowned self] url, _ in
-            self.callback(url != nil)
-            TransactionStore.shared.clear()
+        if #available(iOS 12.0, *) {
+            let authSession = ASWebAuthenticationSession(url: url, callbackURLScheme: schemeURL) { [unowned self] url, _ in
+                self.callback(url != nil)
+                TransactionStore.shared.clear()
+            }
+            self.authSession = authSession
+            authSession.start()
+        } else {
+            let authSession = SFAuthenticationSession(url: url, callbackURLScheme: schemeURL) { [unowned self] url, _ in
+                self.callback(url != nil)
+                TransactionStore.shared.clear()
+            }
+            self.authSession = authSession
+            authSession.start()
         }
-        self.authSession?.start()
     }
 
     func resume(_ url: URL, options: [A0URLOptionsKey: Any]) -> Bool {
--- a/Auth0/SafariAuthenticationCallback.swift
+++ b/Auth0/SafariAuthenticationCallback.swift
@@ -21,8 +21,10 @@
 // THE SOFTWARE.
 
 import UIKit
-import AuthenticationServices
 import SafariServices
+#if canImport(AuthenticationServices)
+import AuthenticationServices
+#endif
 
 #if swift(>=3.2)
 @available(iOS 11.0, *)
@@ -35,6 +37,7 @@ class SafariAuthenticationSessionCallback: AuthTransaction {
 
     init(url: URL, schemeURL: String, callback: @escaping (Bool) -> Void) {
         self.callback = callback
+        #if canImport(AuthenticationServices)
         if #available(iOS 12.0, *) {
             let authSession = ASWebAuthenticationSession(url: url, callbackURLScheme: schemeURL) { [unowned self] url, _ in
                 self.callback(url != nil)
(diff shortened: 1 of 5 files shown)


cocojoe:
@martinoconnor Was a little tricker to add backwards compatibility for Xcode 9. However, all good and passing all platform tests.

Auth0/SafariAuthenticationCallback.swift
```diff
@@ -22,22 +22,45 @@
...
     var state: String?
-    var authSession: SFAuthenticationSession?
     var callback: (Bool) -> Void = { _ in }
 
+    private var authSession: NSObject?
+
     init(url: URL, schemeURL: String, callback: @escaping (Bool) -> Void) {
         self.callback = callback
-        self.authSession = SFAuthenticationSession(url: url, callbackURLScheme: schemeURL) { [unowned self] url, _ in
+        #if canImport(AuthenticationServices)
```
cocojoe: So turns out there was this handy pre-processor added into Swift 4.1 that let's us check for a Framework.
As this Framework does not exist in Xcode 9, it can not compile without this.

Auth0/SafariAuthenticationCallback.swift
```diff
@@ -22,22 +22,45 @@
 
 import UIKit
 import SafariServices
+#if canImport(AuthenticationServices)
```
cocojoe: As this Framework does not exist in Xcode 9, it can not compile without this.

Auth0/SafariAuthenticationCallback.swift
```diff
@@ -22,22 +22,45 @@
...
+            authSession.start()
+        } else {
+            let authSession = SFAuthenticationSession(url: url, callbackURLScheme: schemeURL) { [unowned self] url, _ in
+                self.callback(url != nil)
+                TransactionStore.shared.clear()
+            }
+            self.authSession = authSession
+            authSession.start()
+        }
+        #else
```
cocojoe: Fallback Xcode 9

Auth0/SafariAuthenticationSession.swift
```diff
@@ -22,20 +22,50 @@
 
 import UIKit
 import SafariServices
+#if canImport(AuthenticationServices)
```
cocojoe: As this Framework does not exist in Xcode 9, it can not compile without this.

Auth0/SafariAuthenticationSession.swift
```diff
@@ -22,20 +22,50 @@
...
+                    } else {
+                        self.finish(.failure(error: $1!))
+                    }
+                    return TransactionStore.shared.clear()
+                }
+                _ = TransactionStore.shared.resume(callbackURL, options: [:])
+                }
+            self.authSession = authSession
+            authSession.start()
+        } else {
```
cocojoe: Fallback iOS <=11.0

Auth0/SafariAuthenticationCallback.swift
```diff
@@ -22,22 +22,45 @@
...
-        self.authSession = SFAuthenticationSession(url: url, callbackURLScheme: schemeURL) { [unowned self] url, _ in
+        #if canImport(AuthenticationServices)
+        if #available(iOS 12.0, *) {
+            let authSession = ASWebAuthenticationSession(url: url, callbackURLScheme: schemeURL) { [unowned self] url, _ in
+                self.callback(url != nil)
+                TransactionStore.shared.clear()
+            }
+            self.authSession = authSession
+            authSession.start()
+        } else {
```
cocojoe: Fallback iOS <=11.0

Auth0/SafariAuthenticationCallback.swift
```diff
@@ -22,22 +22,45 @@
...
+            self.authSession = authSession
+            authSession.start()
+        }
+        #else
+        let authSession = SFAuthenticationSession(url: url, callbackURLScheme: schemeURL) { [unowned self] url, _ in
             self.callback(url != nil)
             TransactionStore.shared.clear()
         }
-        self.authSession?.start()
+        self.authSession = authSession
```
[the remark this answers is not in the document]
lbalmaceda: After re-reading this.. Note that "fallback for 11" and "fallback for 9" have the exact same code. The only thing changing is `SFAuthenticationSession` and `ASWebAuthenticationSession` on 1 of the 3 cases. Please consider re-writing the code so that only that first line is wrapped around the OS check and the remaining is shared.

I imagine something like this is possible:


```swift
self.callback = callback
 #if canImport(AuthenticationServices) && #available(iOS 12.0, *)
      self.authSession = ASWebAuthenticationSession(url: url, callbackURLScheme: schemeURL) { 
      [unowned self] url, _ in
                self.callback(url != nil)
                TransactionStore.shared.clear()
            }
#else  
        self.authSession = SFAuthenticationSession(url: url, callbackURLScheme: schemeURL) { 
       [unowned self] url, _ in
                self.callback(url != nil)
                TransactionStore.shared.clear()
            }
#endif 

self.authSession.start()
```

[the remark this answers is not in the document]
cocojoe: The behaviour is slightly more complex than this, I could assign `self.authSession = ASWeb/SFAuth` However at that point it is now a common foundation type of `NSObject`. This is done so we can reuse one property, it's only real purpose here is to retain the object so the AuthenticationSession doesn't vanish until it's completed.

Problem now is you can't call `self.authSession.start()` as `start()` is not available in `NSObject`.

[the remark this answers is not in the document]
lbalmaceda: what about then, 

```swift
self.callback = callback
let authSession;
 #if canImport(AuthenticationServices) && #available(iOS 12.0, *)
      authSession = ASWebAuthenticationSession(url: url, callbackURLScheme: schemeURL) { 
      [unowned self] url, _ in
                self.callback(url != nil)
                TransactionStore.shared.clear()
            }
#else  
        authSession = SFAuthenticationSession(url: url, callbackURLScheme: schemeURL) { 
       [unowned self] url, _ in
                self.callback(url != nil)
                TransactionStore.shared.clear()
            }
#endif 
self.authSession = authSession;
authSession.start()
```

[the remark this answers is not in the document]
cocojoe: It's a good idea in theory. However, you're left with the same problem, Swift is strongly type, you must specify the type for your `let authSession` So you are back to the same problem as your previous suggestion.

Auth0/SafariAuthenticationCallback.swift
cocojoe: These are two different kinds of checks, the `#if` is a pre-processor so if it's on Xcode 10+ include this code. If it was Xcode 9 only the code in the #else would actually be visible to the compiler. The `if  #available(iOS12)` is a runtime check, no escaping it.

PR #253, opened 2019-01-10
Pods Fix - Move AuthenticationServices to weak_framework section

ivabra:
### Changes

Moved `AuthenticationServices` framework dependency from `s.ios.frameworks` to `s.ios.weak_framework` section of `Auth0.podspec` file.

### References

`AuthenticationServices` framework is available only since **iOS 12 and above**. Putting it in the `s.ios.frameworks` section makes it **required** for linking which is a cause of`dyld` error like:

```
dyld: Library not loaded: /System/Library/Frameworks/AuthenticationServices.framework/AuthenticationServices
  Referenced from: /Users/.../Library/Developer/CoreSimulator/Devices/.../data/Containers/Bundle/Application/.../SampleApp.app/Frameworks/Lock.framework/Lock
  Reason: image not found
```
 The solution is to link `AuthenticationServices.framework` weakly, so it won't be linked for apps that are being run on iOS older than 12.0.

Podspec reference:
https://guides.cocoapods.org/syntax/podspec.html#weak_frameworks

Same issue:
https://github.com/adjust/ios_sdk/issues/42

Apple docs:
https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPFrameworks/Concepts/WeakLinking.html

### Testing
- Create a sample iOS project.
- Integrate Auth0 using Podfile and `pod install` command.
- Run the project on a simulator with iOS lower than 12.0.
- Without the fix the app will be crashed with `dyld: Library not loaded: ... Reason: image not found`.
### Checklist

*  I have read the [Auth0 general contribution guidelines](https://github.com/auth0/open-source-template/blob/master/GENERAL-CONTRIBUTING.md)
* All existing and new tests complete without errors
*  All active GitHub checks have passed

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,6 +38,8 @@ jobs:
     environment:
           LC_ALL: en_US.UTF-8
           LANG: en_US.UTF-8
+          HOMEBREW_LOGS: ~/homebrew-logs
+          HOMEBREW_TEMP: ~/homebrew-temp
     steps:
       - checkout
       - run: |
@@ -65,6 +67,8 @@ jobs:
     environment:
           LC_ALL: en_US.UTF-8
           LANG: en_US.UTF-8
+          HOMEBREW_LOGS: ~/homebrew-logs
+          HOMEBREW_TEMP: ~/homebrew-temp
     steps:
       - checkout
       - run: |
@@ -93,6 +97,8 @@ jobs:
     environment:
           LC_ALL: en_US.UTF-8
           LANG: en_US.UTF-8
+          HOMEBREW_LOGS: ~/homebrew-logs
+          HOMEBREW_TEMP: ~/homebrew-temp
     steps:
       - checkout
       - run: |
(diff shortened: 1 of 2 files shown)


cocojoe:
Thanks this is most helpful, I did wonder how Pods will handle weak linking and should have looked at this in more detail vs presuming CocoaPod will just magically take care of it.  

Annoyingly there seems to be a new HomeBrew bug being thrown up and I need tests to pass. Would you mind adding:

```
HOMEBREW_LOGS=~/homebrew-logs
HOMEBREW_TEMP=~/homebrew-temp
```

To the CI Env please https://github.com/auth0/Auth0.swift/blob/master/.circleci/config.yml#L6

Issue Reference: https://github.com/Homebrew/brew/issues/5513

Thanks

ivabra:
@cocojoe No problem, glad to help. Well, at least the test where I added the variables passed.

cocojoe:
Thanks, your changes are good, it's HomeBrew being a pain now. I'm going to force merge this one in for a release and look at the HomeBrew changes separately as it is breaking for pod installations.

PR #259, opened 2019-03-12
Ensure URL encoding of + as %2B Authorize URL [SDK-691]

cocojoe:
### Changes

URL Encoding has changed to force encoding of `+` to `%2B` so `+` can be used
as expected for the `login_hint` to pre-populate the ULP form.

### References

Internally raised issue.

### Testing

### Checklist

### DIFF
--- a/Auth0/SafariWebAuth.swift
+++ b/Auth0/SafariWebAuth.swift
@@ -181,6 +181,7 @@ class SafariWebAuth: WebAuth {
 
         entries.forEach { items.append(URLQueryItem(name: $0, value: $1)) }
         components.queryItems = self.telemetry.queryItemsWithTelemetry(queryItems: items)
+        components.percentEncodedQuery = components.percentEncodedQuery?.replacingOccurrences(of: "+", with: "%2B")
         return components.url!
     }
 
--- a/Auth0Tests/WebAuthSpec.swift
+++ b/Auth0Tests/WebAuthSpec.swift
@@ -233,15 +233,11 @@ class WebAuthSpec: QuickSpec {
             
             context("encoding") {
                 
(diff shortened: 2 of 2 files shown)


Auth0/SafariWebAuth.swift
```diff
@@ -181,6 +181,7 @@ class SafariWebAuth: WebAuth {
 
         entries.forEach { items.append(URLQueryItem(name: $0, value: $1)) }
         components.queryItems = self.telemetry.queryItemsWithTelemetry(queryItems: items)
+        components.percentEncodedQuery = components.percentEncodedQuery?.replacingOccurrences(of: "+", with: "%2B")
```
cocojoe: I had a look at encoding at the key value. However, then the URL will end up double encoded. So this also appears to be the okay by apple http://www.openradar.me/24076063

Auth0Tests/WebAuthSpec.swift
```diff
@@ -230,6 +230,17 @@ class WebAuthSpec: QuickSpec {
...
                 ]
             }
+            
+            context("encoding") {
+                
+                it("should encode + as %2B"){
+                    let url = newWebAuth()
+                            .parameters(["login_hint": "first+last@host.com"])
+                            .buildAuthorizeURL(withRedirectURL: RedirectURL, defaults: defaults, state: "state")
+                    expect(url.absoluteString.contains("first%2Blast@host.com")).to(beTrue())
```
cocojoe: @ is valid in query

Auth0/SafariWebAuth.swift
lbalmaceda: Ugh. Thanks for the link

Issue #274, opened 2019-04-30
Request: custom modalPresentationStyle for SFSafariViewController


colin-chan:
### Description

For UI/UX reasons, we wish to change the default presentation style of the SFSafariViewController. By default it is pushed from the right side, we want it to be presented vertically from the bottom.

Specifically, can you allow custom modification of the 'modalPresentationStyle' property in the created SFSafariViewController in file SafariWebAuth.swift . I would like to change this property to '.overFullScreen' for our use cases.

### Prerequisites

* [x] Did you check the documentation? [Add a link]
* [x] Did you check the [Auth0 Community](https://community.auth0.com/)?
* [x] Are you reporting this to the correct repository?
* [x] Are there any related or duplicate Issues or PRs for this issue?

### Environment

Please provide the following:

- Auth0.swift version (1.14.2)
- iOS version (11.0)
- If using Carthage or CocoaPods, please include your Cartfile or Podfile
- Xcode version (10.2)

### Reproduction
I'm using the legacy authentication, not the newer SafariAuthenticationSession.
```
 Auth0.webAuth(...).useLegacyAuthentication()....start {...}
```

cocojoe:
Hi @colin-chan this is not something we have any plans to work on. However, if you wish to submit a PR to expose this I would be happy to have a look at it for you.

colin-chan:
Hello, and thanks. I've added a pull request with my changes to allow this.

PR #275, opened 2019-04-30
Added UIModalPresentationStyle when using legacy Safari Authentication

colin-chan:
### Changes
Legacy authentication using SFSafariViewController can have its modal presentation style changed from its default. For our cases, this allows the view to be presented vertically from the bottom.

Main changes to WebAuth's useLegacyAuthentication method. It accepts a 'withStyle' argument to override the default style. Backward compatible with previous signature by using a default value.

In SafariWebAuth.swift., new private property 'safariPresentationStyle' to store the default or modified enum style.

Example Usage
```
Auth0.webAuth(...)
.useLegacyAuthentication()
.start {...}
```

```
Auth0.webAuth(...)
.useLegacyAuthentication(withStyle: .overFullScreen)
.start {...}
```
### References

Issue raised: https://github.com/auth0/Auth0.swift/issues/274

### Testing

Added new unit test in Auth0Tests/WebAuthSpec.swift

### Checklist

### DIFF
--- a/Auth0/SafariWebAuth.swift
+++ b/Auth0/SafariWebAuth.swift
@@ -102,7 +102,7 @@ class SafariWebAuth: WebAuth {
         return self
     }
 
-    func useLegacyAuthentication(withStyle style: UIModalPresentationStyle = .fullScreen) -> Self {
+    func useLegacyAuthentication(withStyle style: UIModalPresentationStyle) -> Self {
         self.authenticationSession = false
         self.safariPresentationStyle = style
         return self
--- a/Auth0/SafariWebAuth.swift
+++ b/Auth0/SafariWebAuth.swift
@@ -39,6 +39,7 @@ class SafariWebAuth: WebAuth {
     var responseType: [ResponseType] = [.code]
     var nonce: String?
     private var authenticationSession = true
+    private var safariPresentationStyle = UIModalPresentationStyle.fullScreen
 
     convenience init(clientId: String, url: URL, presenter: ControllerModalPresenter = ControllerModalPresenter(), telemetry: Telemetry = Telemetry()) {
         self.init(clientId: clientId, url: url, presenter: presenter, storage: TransactionStore.shared, telemetry: telemetry)
@@ -106,6 +107,11 @@ class SafariWebAuth: WebAuth {
         return self
     }
 
+    func legacyAuthenticationPresentationStyle(_ style: UIModalPresentationStyle) -> Self {
+        self.safariPresentationStyle = style
+        return self
+    }
+
     func start(_ callback: @escaping (Result<Credentials>) -> Void) {
         guard
             let redirectURL = self.redirectURL, !redirectURL.absoluteString.hasPrefix(SafariWebAuth.NoBundleIdentifier)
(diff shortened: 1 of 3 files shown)


cocojoe:
Please can you rebase against master. Thanks

colin-chan:
I have merged in the master changes to my branch.

Auth0/WebAuth.swift
```diff
@@ -190,6 +190,14 @@ public protocol WebAuth: Trackable, Loggable {
     @available(iOS 11, *)
     func useLegacyAuthentication() -> Self
 
+    /**
+     Change modal presentation style if using `SFSafariViewController`.
+     
+     - returns: the same WebAuth instance to allow method chaining
+     */
+    @available(iOS 11, *)
+    func legacyAuthenticationPresentationStyle(_ style: UIModalPresentationStyle) -> Self
```
cocojoe: I'm not convinced that we need a new protocol method, also a UI method doesn't feel right in this protocol.  Perhaps should be combined as a param in the `useLegacyAuthentication(withStyle: ...)` ?

colin-chan: I have made the change request, have extended useLegacyAuthentication with a withStyle parameter to accept a custom style or use the previous default value.

Auth0/SafariWebAuth.swift
```diff
@@ -101,8 +102,9 @@ class SafariWebAuth: WebAuth {
         return self
     }
 
-    func useLegacyAuthentication() -> Self {
+    func useLegacyAuthentication(withStyle style: UIModalPresentationStyle = .fullScreen) -> Self {
```
cocojoe: as this param is required and must be specified when used, the default is not required.

colin-chan: If I remove the default value then previous versions of this method won't compile. I'm happy to remove the default if you want users to update their code to the default value manually.

colin-chan: Default value has been removed.

cocojoe: The previous version will work as your Public extension is the code that allows facilitates that.

cocojoe: Apologies, you're correct, I had to run this in Xcode as my mental compiler wasn't running properly. I had mixed up the use of the `WebAuth` Protocol vs the `SafariWebAuth` object.

The use case I was thinking about was standard login e.g. Which will be fine as `.webAuth` returns the Protocol and the Public Extension maintains the Public API and Backwards compatibility.

```swift
  Auth0.webAuth()
            .logging(enabled: true)
            .useLegacyAuthentication()
            .start(onAuth)
        }
```

In tests the `SafariWebAuth` is used directly which does not have any default. So please restore :)

Auth0/WebAuth.swift
```diff
@@ -243,3 +244,18 @@ public protocol WebAuth: Trackable, Loggable {
      */
     func clearSession(federated: Bool, callback: @escaping (Bool) -> Void)
 }
+
+public extension WebAuth {
+
```
cocojoe: The CI is not picking this up. However, on this line there is extra whitespace which should be removed.  I would suggest installing SwiftLint. SwiftLint is auto called in this Libraries build script and will pop up any warnings in the IDE. Thanks

colin-chan: I have reverted the last commit and deleted the excess whitespace.

cocojoe:
Please rebase your branch for latest changes. thanks

colin-chan:
I have merged in the latest changes.

colin-chan:
Thank you for the approval. Can you you merge my changes for the next release.

PR #277, opened 2019-05-10
Add support for custom Keychain instance in the CredentialsManager

joakes90:
### Changes

Credential Manager can now be initialized with a custom A0SimpleKeychain object. The current A0SimpleKeychain object is provided as a default value to maintain compatibility with current integrations.

- Classes and methods added, deleted, deprecated, or changed

 added optional '''storage''' argument to '''CredentialManager''' initializer.

- A summary of usage if this is a new feature or change to a public API (this should also be added to relevant documentation once released)

This is relevant to users needing to use a keychain group to access a single set of credentials across multiple sandboxed executables or implement a custom A0SimpleKeychain object to be used by Credential Manager for other reasons.

### References

Heroku ticket 702077
https://help.heroku.com/tickets/702077?authuser=justin.oakes@loves.com#event_9d49b2df-4e03-441e-8e59-c14533354fcd

Call with Sandra, Paul and Alex on 4/22
https://drive.google.com/file/d/1g8AlnNCuBFatexpp4qxxJZjjPDfxv-JW/view?usp=sharing

### Testing

This can be tested by implementing Auth0 in 2 applications signed from the same account in different sandboxes and using keychain groups to read data written in one app from the other

https://developer.apple.com/documentation/security/keychain_services/keychain_items/sharing_access_to_keychain_items_among_a_collection_of_apps

* [√] This change has been tested on the latest version of the platform/language or why not

### Checklist

* [√] I have read the [Auth0 general contribution guidelines](https://github.com/auth0/open-source-template/blob/master/GENERAL-CONTRIBUTING.md)
* [√] All existing and new tests complete without errors
* [√] All active GitHub checks have passed

### DIFF
--- a/Auth0/CredentialsManager.swift
+++ b/Auth0/CredentialsManager.swift
@@ -29,7 +29,7 @@ import LocalAuthentication
 /// Credentials management utility
 public struct CredentialsManager {
 
-    private let storage = A0SimpleKeychain()
+    private let storage: A0SimpleKeychain
     private let storeKey: String
     private let authentication: Authentication
     #if os(iOS)
@@ -41,9 +41,10 @@ public struct CredentialsManager {
     /// - Parameters:
     ///   - authentication: Auth0 authentication instance
     ///   - storeKey: Key used to store user credentials in the keychain, defaults to "credentials"
-    public init(authentication: Authentication, storeKey: String = "credentials") {
+    public init(authentication: Authentication, storeKey: String = "credentials", storage: A0SimpleKeychain = A0SimpleKeychain()) {
         self.storeKey = storeKey
         self.authentication = authentication
+        self.storage = storage
     }
 
     /// Enable Touch ID Authentication for additional security during credentials retrieval.
--- a/Auth0/CredentialsManager.swift
+++ b/Auth0/CredentialsManager.swift
@@ -29,7 +29,7 @@ import LocalAuthentication
 /// Credentials management utility
 public struct CredentialsManager {
 
-    internal let storage: A0SimpleKeychain
+    private let storage: A0SimpleKeychain
     private let storeKey: String
     private let authentication: Authentication
     #if os(iOS)
--- a/Auth0/CredentialsManager.swift
+++ b/Auth0/CredentialsManager.swift
@@ -41,7 +41,7 @@ public struct CredentialsManager {
     /// - Parameters:
     ///   - authentication: Auth0 authentication instance
     ///   - storeKey: Key used to store user credentials in the keychain, defaults to "credentials"
-    ///   - storage: The A0SimpleKeychain object that user credentials are saved to. Defaults to the standard application keychain
+    ///   - storage: The A0SimpleKeychain instance used to manage credentials storage. Defaults to a standard A0SimpleKeychain instance
     public init(authentication: Authentication, storeKey: String = "credentials", storage: A0SimpleKeychain = A0SimpleKeychain()) {
         self.storeKey = storeKey
         self.authentication = authentication
--- a/Auth0/CredentialsManager.swift
+++ b/Auth0/CredentialsManager.swift
@@ -41,6 +41,7 @@ public struct CredentialsManager {
     /// - Parameters:
     ///   - authentication: Auth0 authentication instance
     ///   - storeKey: Key used to store user credentials in the keychain, defaults to "credentials"
+    ///   - storage: The A0SimpleKeychain object that user credentials are saved to. Defaults to the standard application keychain
     public init(authentication: Authentication, storeKey: String = "credentials", storage: A0SimpleKeychain = A0SimpleKeychain()) {
         self.storeKey = storeKey
         self.authentication = authentication
--- a/Auth0/CredentialsManager.swift
+++ b/Auth0/CredentialsManager.swift
@@ -29,7 +29,7 @@ import LocalAuthentication
 /// Credentials management utility
 public struct CredentialsManager {
 
-    private let storage: A0SimpleKeychain
+    internal let storage: A0SimpleKeychain
     private let storeKey: String
     private let authentication: Authentication
     #if os(iOS)
--- a/Auth0Tests/CredentialsManagerSpec.swift
+++ b/Auth0Tests/CredentialsManagerSpec.swift
@@ -309,10 +309,10 @@ class CredentialsManagerSpec: QuickSpec {
                                                             storage: storage)
                 }
                 
-                it("should accept custom keychains") {
-                    expect(credentialsManager.store(credentials: credentials)).to(beTrue())
+                it("custom keychain should successfully set and clear credentials") {
+                    _ = credentialsManager.store(credentials: credentials)
                     expect(storage.data(forKey: "credentials")).toNot(beNil())
-                    expect(credentialsManager.clear()).to(beTrue())
+                    _ = credentialsManager.clear()
                     expect(storage.data(forKey: "credentials")).to(beNil())
                 }
             }
--- a/Auth0Tests/CredentialsManagerSpec.swift
+++ b/Auth0Tests/CredentialsManagerSpec.swift
@@ -45,12 +45,11 @@ class CredentialsManagerSpec: QuickSpec {
     override func spec() {
 
         let authentication = Auth0.authentication(clientId: ClientId, domain: Domain)
-        let storage = A0SimpleKeychain(service: "test_service")
         var credentialsManager: CredentialsManager!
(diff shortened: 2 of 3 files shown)


cocojoe:
This seems resoanable for advanced use case. @joakes90 please rebase against master and optionally clean up your commits, as it's only 3 lines of changes :)

Auth0/CredentialsManager.swift
```diff
@@ -41,9 +41,10 @@ public struct CredentialsManager {
     /// - Parameters:
     ///   - authentication: Auth0 authentication instance
     ///   - storeKey: Key used to store user credentials in the keychain, defaults to "credentials"
```
cocojoe: Please update Doc Header with new param.

Auth0/CredentialsManager.swift
```diff
@@ -29,7 +29,7 @@ import LocalAuthentication
 /// Credentials management utility
 public struct CredentialsManager {
 
-    private let storage = A0SimpleKeychain()
+    internal let storage: A0SimpleKeychain
```
cocojoe: Which change to internal? The configured keychain should be passed in via the init?

cocojoe:
Please update

Auth0/CredentialsManager.swift
joakes90: access to this property is changed back to private in the latest commit.

Auth0/CredentialsManager.swift
joakes90: Documentation for the new argument is now provided in the latest commit.

Auth0/CredentialsManager.swift
joakes90: Observing the storage property in needed to test that it is indeed using a bespoke keychain

Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -45,11 +45,12 @@ class CredentialsManagerSpec: QuickSpec {
     override func spec() {
 
         let authentication = Auth0.authentication(clientId: ClientId, domain: Domain)
+        let storage = A0SimpleKeychain(service: "test_service")
         var credentialsManager: CredentialsManager!
         var credentials: Credentials!
 
         beforeEach {
-            credentialsManager = CredentialsManager(authentication: authentication)
+            credentialsManager = CredentialsManager(authentication: authentication, storage: storage)
```
cocojoe: I don't think you should change the default behaviour to use a custom storage for tests, this is your use case but it is not the 90%+ customer use case.
It makes sense now that you can change the keychain to add a test here for the `credentials` service as this is default.

Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -71,7 +72,11 @@ class CredentialsManagerSpec: QuickSpec {
             it("should fail to clear credentials") {
                 expect(credentialsManager.clear()).to(beFalse())
             }
-            
+
+            it("should accept custom keychains") {
```
cocojoe: Structure wise I think this makes more sense to wrap this up as part of a `context block` related to passing a custom keychain. Also in this particular test the `toNot` is redundant.

Auth0/CredentialsManager.swift
cocojoe: Ideally we could use a mock, although no protocol here. I would move this to it's original state of private and test against your custom instance.

Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -303,6 +302,17 @@ class CredentialsManagerSpec: QuickSpec {
...
             
+            context("custom keychain", closure: {
+                beforeEach {
+                    let storage = A0SimpleKeychain(service: "test_service")
+                    credentialsManager = CredentialsManager(authentication: authentication,
+                                                            storage: storage)
+                }
+                
+                it("should accept custom keychains") {
+                    expect(credentialsManager.storage.service).to(match("test_service"))
```
cocojoe: This is looking better, to support restoring the access level to private of the storage property.  You should test against the `storage` instance. So `expect(storage.service` should work fine.

Auth0/CredentialsManager.swift
```diff
@@ -41,9 +41,11 @@ public struct CredentialsManager {
     /// - Parameters:
     ///   - authentication: Auth0 authentication instance
     ///   - storeKey: Key used to store user credentials in the keychain, defaults to "credentials"
-    public init(authentication: Authentication, storeKey: String = "credentials") {
+    ///   - storage: The A0SimpleKeychain object that user credentials are saved to. Defaults to the standard application keychain
```
cocojoe: ///   - storage: The A0SimpleKeychain instance used to manage credentials storage. Defaults to a standard A0SimpleKeychain instance

Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -47,6 +47,7 @@ class CredentialsManagerSpec: QuickSpec {
         let authentication = Auth0.authentication(clientId: ClientId, domain: Domain)
         var credentialsManager: CredentialsManager!
         var credentials: Credentials!
+        var storage: A0SimpleKeychain?
```
cocojoe: This isn't needed if you are only using it for the testing below.

Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -303,6 +303,17 @@ class CredentialsManagerSpec: QuickSpec {
                 }
             }
             
+            context("custom keychain", closure: {
```
cocojoe: For consistency there is no need to add the closure param name.
You can use `context("custom keychain") { .... }`

The storage property can then be set within the scope of this block so it will be accessible for `beforeEach` and the `it` tests, which also removes the need for the nil coalescing operator when assigning storage.

Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -303,6 +303,17 @@ class CredentialsManagerSpec: QuickSpec {
...
+                beforeEach {
+                    storage = A0SimpleKeychain(service: "test_service")
+                    credentialsManager = CredentialsManager(authentication: authentication,
+                                                            storage: storage ?? A0SimpleKeychain())
+                }
+                
+                it("should accept custom keychains") {
+                    expect(storage?.service).to(match("test_service"))
```
cocojoe: This only tests that you set the service name in the service you previously created.  I don't think I explained this at all in the last comment.  
What we need to do here is set the storage object as you do. Then perform some kind of action. For example you can manually add something to the keychain instance and then call on the credentials manager to clear it, then test it's empty.

Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -303,17 +302,20 @@ class CredentialsManagerSpec: QuickSpec {
...
                 beforeEach {
-                    storage = A0SimpleKeychain(service: "test_service")
                     credentialsManager = CredentialsManager(authentication: authentication,
-                                                            storage: storage ?? A0SimpleKeychain())
+                                                            storage: storage)
                 }
                 
                 it("should accept custom keychains") {
-                    expect(storage?.service).to(match("test_service"))
+                    expect(credentialsManager.store(credentials: credentials)).to(beTrue())
```
cocojoe: don't need to test the outcome here as we are testing for it through the data tests

Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -303,17 +302,20 @@ class CredentialsManagerSpec: QuickSpec {
...
+                    expect(storage.data(forKey: "credentials")).toNot(beNil())
+                    expect(credentialsManager.clear()).to(beTrue())
```
cocojoe: don't need to test the outcome here either as we are testing for the outcome in the data tests

Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -303,17 +302,20 @@ class CredentialsManagerSpec: QuickSpec {
...
-                    storage = A0SimpleKeychain(service: "test_service")
                     credentialsManager = CredentialsManager(authentication: authentication,
-                                                            storage: storage ?? A0SimpleKeychain())
+                                                            storage: storage)
                 }
                 
                 it("should accept custom keychains") {
-                    expect(storage?.service).to(match("test_service"))
+                    expect(credentialsManager.store(credentials: credentials)).to(beTrue())
+                    expect(storage.data(forKey: "credentials")).toNot(beNil())
```
cocojoe: good sanity check test 👍

PR #280, opened 2019-05-28
Update app configuration error message for PKCE

lbalmaceda:
### Changes

The current requirements state that the application should be of type "native". Normally that's enough, but some users had the token endpoint authentication method set to other than "none", making the whole code+PKCE flow fail. 
This PR attempts to prevent this by clarifying that both settings should have a given value for PKCE to work.

### Testing

### Checklist

### DIFF
--- a/Auth0/OAuth2Grant.swift
+++ b/Auth0/OAuth2Grant.swift
@@ -106,7 +106,7 @@ struct PKCE: OAuth2Grant {
             .start { result in
                 // Special case for PKCE when the correct method for token endpoint authentication is not set (it should be None)
                 if case .failure(let cause as AuthenticationError) = result, cause.description == "Unauthorized" {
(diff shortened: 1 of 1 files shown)


Auth0/OAuth2Grant.swift
```diff
@@ -106,7 +106,7 @@ struct PKCE: OAuth2Grant {
             .start { result in
                 // Special case for PKCE when the correct method for token endpoint authentication is not set (it should be None)
                 if case .failure(let cause as AuthenticationError) = result, cause.description == "Unauthorized" {
-                    let error = WebAuthError.pkceNotAllowed("Please go to 'https://manage.auth0.com/#/applications/\(clientId)/settings' and make sure 'Client Type' is 'Native' to enable PKCE.")
+                    let error = WebAuthError.pkceNotAllowed("Please go to 'https://manage.auth0.com/#/applications/\(clientId)/settings' and make sure 'Token Endpoint Authentication Method' is 'None' and 'Application Type' is 'Native' to enable PKCE.")
```
joshcanhelp: > to enable PKCE

Consider:

> to allow PKCE

Issue #213, opened 2018-08-20
Auth0 login EXC_BAD_ACCESS after callback


nickvelloff:
I'm getting a `EXC_BAD_ACCESS` right after logging in. This happens after clicking the "magic link" that is hosted. It opens the app, calls the delegate method `func application(_ app: UIApplication, open url: URL, options: [UIApplicationOpenURLOptionsKey : Any]) -> Bool` correctly, then crashes. 

I had this issue in my main project, so I am testing with a completely new clean project and have the same issue. 

Seems to have something to do with `*** -[SFAuthenticationSession release]: message sent to deallocated instance 0x1c4244290` when I enable Zombie Objects.

Dashboard logs show a successful flow:
```
Code/Link Sent
Success Login
Success Exchange
```

SO I must just have something configured incorrectly. 

Xcode Version 9.4.1 (9F2000)
iOS 11.4.1 (15G77)

Crash reports. 
```
Incident Identifier: B4B68107-E3E4-423B-82CB-31384B937C07
CrashReporter Key:   59d877a4b8439d465c8db72709bfa3ca033ae4e0
Hardware Model:      iPhone10,6
Process:             Auth0Test [13660]
Path:                /private/var/containers/Bundle/Application/5D68C0B5-2096-4978-B70E-480CBBEABD9D/Auth0Test.app/Auth0Test
Identifier:          com.Auth0Test
Version:             1 (1.0)
Code Type:           ARM-64 (Native)
Role:                Foreground
Parent Process:      launchd [1]
Coalition:           com.Auth0Test [6329]


Date/Time:           2018-08-20 01:05:12.5728 -0400
Launch Time:         2018-08-20 01:04:47.6512 -0400
OS Version:          iPhone OS 11.4.1 (15G77)
Baseband Version:    1.04.80
Report Version:      104

Exception Type:  EXC_BAD_ACCESS (SIGSEGV)
Exception Subtype: KERN_INVALID_ADDRESS at 0x0000000200000000
VM Region Info: 0x200000000 is not in any region.  Bytes after previous region: 536870913  
      REGION TYPE                      START - END             [ VSIZE] PRT/MAX SHRMOD  REGION DETAIL
      MALLOC_NANO (reserved) 00000001d8000000-00000001e0000000 [128.0M] rw-/rwx SM=NUL  ...(unallocated)
--->  
      UNUSED SPACE AT END

Termination Signal: Segmentation fault: 11
[...]

PR #281, opened 2019-06-07
Fix Dismiss AS/SF authentication sessions upon deep-link callback

cysp:
### Changes

Cancel any active `ASWebAuthenticationSession` of `SFAuthenticationSession` when completing via a deep-link callback.
Currently this scenario results in a crash due to messaging a deallocated authentication session object, as reported in https://github.com/auth0/Auth0.swift/issues/213

Technical implementation:
- create a wrapper that can hold either a `ASWebAuthenticationSession` or a  `SFAuthenticationSession`, exposing their common API
- in `SafariAuthenticationSession`, override `AuthSession`'s implementation of `resume(_:options:)`, forwarding to `AuthSession` and, if successfully resumed, dismissing the relevant authentication session
- whilst here, also override `cancel()` to dismiss the authentication session

### References

- https://github.com/auth0/Auth0.swift/issues/213

### Testing

I couldn't see any existing test suite for this functionality so I _tried_ to write some to cover the completion / cancellation logic but I have not had any success. 😕 

To manually test this, configure an Auth0 tenant with passwordless authentication via email and attempt to sign in using the link in the email.

I haven't executed the `!swift(>3.2)` branches here, but I didn't think that I should just abandon them, nor remove them in this PR. 🙃 

### Checklist

### DIFF
--- a/Auth0/SafariAuthenticationSession.swift
+++ b/Auth0/SafariAuthenticationSession.swift
@@ -29,14 +29,37 @@ import AuthenticationServices
 #if swift(>=3.2)
 @available(iOS 11.0, *)
 class SafariAuthenticationSession: AuthSession {
+    private enum AuthenticationSession {
+        @available(iOS 12.0, *)
+        case authenticationServices(ASWebAuthenticationSession)
+        case safariServices(SFAuthenticationSession)
 
-    private var authSession: NSObject?
+        @available(iOS 12.0, *)
+        init(_ session: ASWebAuthenticationSession) {
+            self = .authenticationServices(session)
+        }
+
+        init(_ session: SFAuthenticationSession) {
+            self = .safariServices(session)
+        }
+
+        func start() -> Bool {
+            switch self {
+            case .authenticationServices(let session):
+                return session.start()
+            case .safariServices(let session):
+                return session.start()
+            }
+        }
+    }
+
+    private var authSession: AuthenticationSession?
 
     init(authorizeURL: URL, redirectURL: URL, state: String? = nil, handler: OAuth2Grant, finish: @escaping FinishSession, logger: Logger?) {
         super.init(redirectURL: redirectURL, state: state, handler: handler, finish: finish, logger: logger)
         #if canImport(AuthenticationServices)
         if #available(iOS 12.0, *) {
-            let authSession = ASWebAuthenticationSession(url: authorizeURL, callbackURLScheme: self.redirectURL.absoluteString) { [unowned self] in
+            authSession = .init(ASWebAuthenticationSession(url: authorizeURL, callbackURLScheme: self.redirectURL.absoluteString) { [unowned self] in
                 guard $1 == nil, let callbackURL = $0 else {
                     let authError = $1 ?? WebAuthError.unknownError
                     if case ASWebAuthenticationSessionError.canceledLogin = authError {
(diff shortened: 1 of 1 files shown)


cocojoe:
Can you add detailed, test steps please. Will have someone validate this. Thanks

cocojoe:
@cysp please can you add a before / after video so I can see your testing in action. Thanks

cysp:
Hi @cocojoe, sorry for the radio silence. 😬 
I can certainly do up some recordings showing the behavioural difference! 🙂 
This is an unreleased app, so would it be okay if I emailed them to you rather than posting them here in public? Is your git committer address okay for that?

cocojoe:
Sure that's fine. 👍

cysp:
Hi @cocojoe, I've finally gotten around to taking those screen recordings and have sent them through to the email address used in your commits in this repo.

cocojoe:
Thanks, videos look fine. Can you rebase your branch please.

cysp:
Cool, rebased. 🙂

Auth0/SafariAuthenticationSession.swift
```diff
@@ -29,14 +29,46 @@ import AuthenticationServices
 #if swift(>=3.2)
 @available(iOS 11.0, *)
 class SafariAuthenticationSession: AuthSession {
+    private enum AuthenticationSession {
```
hzalaz: Can we use a regular class instead of an enum?

Auth0/SafariAuthenticationSession.swift
```diff
@@ -75,10 +103,23 @@ class SafariAuthenticationSession: AuthSession {
...
-        authSession.start()
+        })
         #endif
+
+        _ = authSession?.start()
+    }
+
+    override func resume(_ url: URL, options: [A0URLOptionsKey : Any]) -> Bool {
+        if super.resume(url, options: options) {
+            authSession?.cancel()
```
hzalaz: Should we `nil` the `authSession`?

Auth0/SafariAuthenticationSession.swift
```diff
@@ -75,10 +103,23 @@ class SafariAuthenticationSession: AuthSession {
...
+            return true
+        }
+        return false
+    }
+
+    override func cancel() {
+        super.cancel()
+        authSession?.cancel()
```
hzalaz: Should we `nil` the `authSession`?

cocojoe:
@cysp ☝️

Auth0/SafariAuthenticationSession.swift
cysp: What kind of design do you have in mind using a class or struct?
Also is there a rationale for avoiding enumerations? They're a first-class type in Swift and this use-case seems like a perfect fit for an enumeration.

Auth0/SafariAuthenticationSession.swift
cysp: Probably wouldn't hurt? ¯\\_(ツ)_/¯

Auth0/SafariAuthenticationSession.swift
hzalaz: I dont see it as a natural use of an enum or us having more options than these two, maybe something like this:

```swift
struct NativeAuthSession {

  @available(iOS 12.0, *)
  private var authServiceSession: ASWebAuthenticationSession?
  private var safariSession: SFAuthenticationSession?
  
  @available(iOS 12.0, *)
  init(_ session: ASWebAuthenticationSession) {
    this.authServiceSession = session;
  }
 
  init(_ session: SFAuthenticationSession) {
    this.safariSession = session;
  }

  func start() -> Bool {
    if #available(iOS 12.0, *) {
      this.authServiceSession.start();
    } else {
      this.safariServices.start();
    }
  }

  func cancel() {
    if #available(iOS 12.0, *) {
      this.authServiceSession?.cancel();
    } else {
      this.safariServices?.cancel();
    }
  }
}
```

(might have typos since I coded it on the GH UI)

But probably is going into too mich details

cysp: The way I see it, we have the situation where we know that an object is one of two types, and that we will only have one, not both of those types of object. To me this seems like a perfect fit for an enum, since we can encode the fact that there's exactly one of those objects, and the type of that object into the type system.
The design where a struct has multiple optional properties feels like a less good fit because the type system admits the possibility that both of those properties may be nil at the same time, and requires unwrapping the optional in order to operate on the contained object.

Is there a style guide or a reason that you would prefer not to use an enum for this use case?

cocojoe:
@cysp can you re-base please and we can get this merged, I'm guessing there are a couple of conflicts now from most recent PR #286  as there is no option to update for me.

cysp:
Sure, I've rebased onto master and it looks good to me.

PR #291, opened 2019-07-26
Fix issue #290 Make Telemetry a class instead of a struct

### LINKED ISSUE #290: Memory leak once `Auth0.authentication` method is called
### Description

There is a memory leak every time we create and keep a new `Authentication` object by calling `Auth0.authentication` static method. 

### Environment

- Auth0.swift version: 1.16.0
- iOS version: 12.4
- CocoaPods version: 1.7.5
- Xcode version: 10.3 (10G8)

### Reproduction

* Create a new Single View App project on XCode
* Add Podfile to the project and install `Auth0` pod (version 1.16.0)
* In the ViewController write this simple code:
```
import UIKit
import Auth0

class ViewController: UIViewController {   
    var authentication: Authentication?

    override func viewDidLoad() {
        super.viewDidLoad()
        // Do any additional setup after loading the view.
        
        authentication = Auth0.authentication(clientId: "****************", domain: "**********")
    }
}
```

* Build project and Profile it (Product -> Profile). 
* You will see a memory leak triggered by calling `Auth0.authentication` method (see screenshot attached here below)

Screenshot 2019-07-26 at 16 38 43

MarcoFilosi:
### Changes

Fix issue #290 
Made Telemetry a class instead of a struct to avoid potential memory leak while calling `Auth0.authentication` method.

### References

Telemetry was creating a strong reference cycle into the struct:
https://stackoverflow.com/questions/34406196/swift-struct-memory-leak


Widcket:
Just an update, the investigation of the memory leak and its fix is in our backlog. We haven't gotten around it yet, as there are other higher priority items.

Widcket:
Closing, as the [original issue](https://github.com/auth0/Auth0.swift/issues/290) could not be reproduced.

PR #296, opened 2019-08-13
Added authentication method for default directory login

cocojoe:
### Changes

Added support for `password` grant and authentication with default directory

Method Added

```swift
func loginDefaultDirectory(withUsername username: String, password: String, audience: String? = nil, scope: String? = nil, parameters: [String: Any]? = nil)
```

### References

ESD-1363

See: https://auth0.com/docs/api/authentication#resource-owner-password

### Testing

### Checklist

### DIFF
--- a/Auth0/Auth0Authentication.swift
+++ b/Auth0/Auth0Authentication.swift
@@ -72,6 +72,22 @@ struct Auth0Authentication: Authentication {
         return Request(session: session, url: resourceOwner, method: "POST", handle: authenticationObject, payload: payload, logger: self.logger, telemetry: self.telemetry)
     }
 
+    func loginDefaultDirectory(withUsername username: String, password: String, audience: String? = nil, scope: String? = nil, parameters: [String: Any]? = nil) -> Request<Credentials, AuthenticationError> {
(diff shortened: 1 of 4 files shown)


Auth0/Auth0Authentication.swift
```diff
@@ -72,6 +72,22 @@ struct Auth0Authentication: Authentication {
...
 
+    func loginDefaultDirectory(withUsername username: String, password: String, audience: String? = nil, scope: String? = nil, parameters: [String: Any]? = nil) -> Request<Credentials, AuthenticationError> {
+        let resourceOwner = URL(string: "/oauth/token", relativeTo: self.url)!
+        var payload: [String: Any] = [
+            "username": username,
+            "password": password,
+            "grant_type": "password",
+            "client_id": self.clientId
+        ]
+        payload["audience"] = audience
```
damieng: Why can't audience and scope be set in the initializer with the other payloads?

cocojoe: It is possible although as these params are optionals `<type>?`. However, the dictionary is not `: Any` you will get a compiler warning to either cast it or provide defaults. If you provide defaults such as '' then you will be sending params with empty strings in the request.

Issue #295, opened 2019-08-13
iOS 13 Beta changes, Xcode 11 Beta and Auth0 Swift SDK.


CapeFurSeal:
### Description

When calling the webAuth function in iOS 13 nothing happens as the new delegate has not been implemented. Will this be fixed before the iOS 13 release?

```txt
Error:
Failed with Error Domain=com.apple.AuthenticationServices.WebAuthenticationSession Code=2 "Cannot start ASWebAuthenticationSession without providing presentation context. Set presentationContextProvider before calling -start." UserInfo={NSDebugDescription=Cannot start ASWebAuthenticationSession without providing presentation context. Set presentationContextProvider before calling -start.}
```

<img width="305" alt="Screenshot 2019-08-17 at 09 21 38" src="https://user-images.githubusercontent.com/928115/63210078-594e7780-c0e1-11e9-9a61-6eddf81104a0.png">

```swift
Auth0
        .authentication(clientId:"", domain:"")
        .webAuth(withConnection:"")
        .start { result in
          switch result {
          case .success(let credentials):
            print("credentials: \(credentials)")
          case .failure(let error):
            print("Failed with \(error)")
          }
      }
```

### Prerequisites

* [x] Did you check the documentation? [Add a link]
* [x] Did you check the [Auth0 Community](https://community.auth0.com/)?
* [x] Are you reporting this to the correct repository?
* [x] Are there any related or duplicate Issues or PRs for this issue?

cocojoe:
Support will be made for iOS 13. You can have a play with the beta branch in #297 and feedback if you like.

CapeFurSeal:
Thanks @cocojoe . Much appreciated.

CapeFurSeal:
This worked perfectly on iOS 13.1 Public beta 2 with branch #297 .

wolfspy:
We are still experiencing this issue on v1.17.1 using the GM seed of iOS 13.

cocojoe:
All of this is addressed in 17.1 release.

Tirth21896:
I have updated my pods with all new pod release 1.18.0 and it's still giving me "`Error Domain=com.apple.AuthenticationServices.WebAuthenticationSession Code=2 "Cannot start ASWebAuthenticationSession without providing presentation context. Set presentationContextProvider before calling -start." UserInfo={NSDebugDescription=Cannot start ASWebAuthenticationSession without providing presentation context. Set presentationContextProvider before calling -start.}`" error.

[...]

PR #297, opened 2019-08-14
Added support for iOS13 changes in ASWebAuthenticationSession

cocojoe:
### Changes

iOS 13 Requires [presentationContextProvider](https://developer.apple.com/documentation/authenticationservices/aswebauthenticationsession/3237232-presentationcontextprovider) to be set in [ASWebAuthenticationSession](https://developer.apple.com/documentation/authenticationservices/aswebauthenticationsession) so the Browser can be displayed. Without this, the WebAuthentication flow will fail.

This PR adds default window for this so there is no breaking changing required for this to be set by the developer.

_Notes:_
Developers will need to update to this minor otherwise the previous version which will compile will have a run time issue due to the breaking change in iOS 13 behaviour. See References

### References

#295 

### Testing

### Checklist


Auth0/Telemetry.swift
```diff
@@ -100,7 +100,9 @@ public struct Telemetry {
     }
 
     static func swiftVersion() -> String {
-        #if swift(>=4.0)
```
cocojoe: Quick change to add support for 5.x logging and no you can't simply ask for the literal string version

Auth0/SafariWebAuth.swift
```diff
@@ -236,7 +239,7 @@ class SafariWebAuth: WebAuth {
 }
 
 private func generateDefaultState() -> String? {
-    var data = Data(count: 32)
```
cocojoe: compiler warning fix

Auth0/SafariAuthenticationSession.swift
```diff
@@ -127,4 +135,13 @@ class SafariAuthenticationSession: AuthSession {
         authSession = nil
     }
 }
+
+#if swift(>=5.1)
```
cocojoe: Basically asks is this Xcode 11 as the compiler is 5.1

.circleci/config.yml
```diff
@@ -1,8 +1,43 @@
...
+    environment:
+          LC_ALL: en_US.UTF-8
+          LANG: en_US.UTF-8
+          HOMEBREW_LOGS: ~/homebrew-logs
+          HOMEBREW_TEMP: ~/homebrew-temp
+          HOMEBREW_NO_AUTO_UPDATE: 1
+    steps:
+      - run:
+        name: Set Ruby Version
+        command:  echo "ruby-2.5" > ~/.ruby-version
```
cocojoe: Default ruby not working in this new image, this is the workaround.

Auth0.xcodeproj/project.pbxproj
```diff
@@ -1276,7 +1276,7 @@
 			isa = PBXProject;
```
cocojoe: In general you can ignore these, it's just updating to latest version

Auth0/SafariAuthenticationSession.swift
```diff
@@ -64,11 +67,11 @@ class SafariAuthenticationSession: AuthSession {
 
     private var authSession: AuthenticationSession?
 
-    init(authorizeURL: URL, redirectURL: URL, state: String? = nil, handler: OAuth2Grant, finish: @escaping FinishSession, logger: Logger?) {
+    init(authorizeURL: URL, redirectURL: URL, state: String? = nil, handler: OAuth2Grant, finish: @escaping FinishSession, logger: Logger?, presentationContextController: Any? = nil) {
```
damieng: Where is the presentationContextController parameter used?  I see a webAuthenticationSession.presentationContextProvider set statement on line 88 but it sets to self.

cocojoe: Good catch, this was left over from refactoring this code. It's no longer required as this class now implements the protocol via an extension at https://github.com/auth0/Auth0.swift/blob/added-ios13-webauth-compatibility/Auth0/SafariAuthenticationSession.swift#L141

PR #301, opened 2019-08-27
Fixed ability to clear WebAuth session in iOS 13

matt:
Thank you, @cocojoe, for helping start the migration to Xcode 11/iOS 13 compatibility.

### Description

I updated a project I'm working on to version 1.17. However, I noticed I am still unable to successfully clear a WebAuth session, `Auth0.webAuth().clearSession(federated: false) { isSuccess in ... }`.

### Changes

iOS 13 Requires presentationContextProvider to be set in ASWebAuthenticationSession so the Browser can be displayed. Without this, the WebAuthentication flow will fail.

This PR adds default window for this so there is no breaking changing required for this to be set by the developer.

_Notes:_
Developers will need to update to this minor otherwise the previous version which will compile will have a run time issue due to the breaking change in iOS 13 behaviour. See References

### References

#295 
#297 

### Testing

1. Start a WebAuth session
2. Clear a WebAuth session

* [x] This change has been tested on the latest version of the platform

### Checklist

### DIFF
--- a/Auth0/SafariAuthenticationCallback.swift
+++ b/Auth0/SafariAuthenticationCallback.swift
@@ -28,7 +28,7 @@ import AuthenticationServices
 
 #if swift(>=3.2)
 @available(iOS 11.0, *)
-class SafariAuthenticationSessionCallback: AuthTransaction {
+class SafariAuthenticationSessionCallback: NSObject, AuthTransaction {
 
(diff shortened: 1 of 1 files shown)


cocojoe:
Good catch on that, let me take a look later today. Thanks

PR #305, opened 2019-09-01
Add Ability To Set `DismissButtonStyle` When Using `SFSafariViewController`

andrewfoghel:
###### Issue: https://community.auth0.com/t/done-button-in-webview/23718/4

# Description
This PR handles setting the `SFSafariViewController.dismissButtonStyle` to `.cancel` when using  the `useLegacyAuthentication(withStyle:)` method. Above are 2 places where people have asked for this feature, and I'm sure they aren't the only ones who need this support.

# Testing
We've added a spec to ensure that this gets set properly when using `useLegacyAuthentication(withStyle:)`, however; the test will only run on iOS 11.0 and up. If there are any ideas for testing this please let me know. 

# Screenshots/GIFs
Screen Shot 2019-09-01 at 12 51 14 AM

### DIFF
--- a/App/Info.plist
+++ b/App/Info.plist
@@ -15,7 +15,7 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string>1.17.0</string>
+	<string>1.17.1</string>
 	<key>CFBundleSignature</key>
 	<string>????</string>
 	<key>CFBundleURLTypes</key>
--- a/Auth0.xcodeproj/project.pbxproj
+++ b/Auth0.xcodeproj/project.pbxproj
@@ -7,7 +7,6 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
-		0B827B04231B7EB10006CC57 /* WebAuthSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5F53F5D61CFFAA4A00476A46 /* WebAuthSpec.swift */; };
 		5B0893E620F8A52100FBF962 /* CredentialsManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5BEDE1891EC21B040007300D /* CredentialsManager.swift */; };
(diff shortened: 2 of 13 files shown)


andrewfoghel:
Closing as there are failures, going to investigate.

cocojoe:
@andrewfoghel Thanks for the PR and screenshot. What is the use case? I noticed the other issues but they also give no reason as to why it's important.

Looking at the above I feel that the current default `Done` is not really the optimal value and the value should instead align with Apple's recommended WebAuth flow in iOS 11+. This uses `Cancel` which makes sense from an end user perspective as you are cancelling Web Auth.

So I think it should default to `Cancel` to match Apple's behaviour. Does this align with the goals you had in mind?

andrewfoghel:
@cocojoe Thanks for the response here! This use case is more so for people using the `useLegacyAuthentication()` helper. It gives people access to customize that button when using an `SFSafariViewController`. I believe that when using `ASWebAuthenticationSession` it is always `Cancel` and that you can't change that value. However, for `SFSafariViewController` it gets defaulted to `Done` so allowing them to change it to `Close` or `Cancel` gives them the opportunity to maintain more consistency throughout their application. Thoughts?

andrewfoghel:
We will not be using the `useLegacyAuthentication()` helper anymore, so we personally don't need this as a part of the SDK. However, I don't want to take from the community that uses that method, thus; I'll leave this open. Feel free to close this if this functionality is deemed as unnecessary or overkill. Thanks for the careful consideration with this!

On a separate note, if you all decide that you want this feature in I will maintain this PR.

cocojoe:
@andrewfoghel I feel we can address the root cause here changing the default of the Library to specify `Cancel` which aligns with Apple's current default messaging in `ASWebAuthentictionSession`.

This should address the concern without having to expose additional methods to allow this to be changed by the developer.

andrewfoghel:
@cocojoe I was thinking it would be nice to give the developer the option to choose whichever option they want; however, I'm not opposed to simply defaulting the button to say `Cancel` to match `ASWebAuthenticationSession`. Those changes should be reflected in the last few commits, let me know what you think!

cocojoe:
@andrewfoghel Thanks for changing, I was on vacation, hence the delay in response :) Looks good 👍

andrewfoghel:
@cocojoe thanks for the approval here! I'm going to delete the branch for prosperity, til next time! (:

Issue #314, opened 2019-10-10
Include reason why removing Auth0 session did not succeed to improve client-side error handling.


LinkAndreas:
## Improve client-side error handling when removing Auth0 session

Affected method in `WebAuth.swift`:

`func clearSession(federated: Bool, callback: @escaping (Bool) -> Void)`

### Problem:

Currently, when removing the Auth0- and/or Identity Provider session, a boolean is used to inform the client whether the operation succeeded or failed. Hence, client-wise error handling capabilities are limited, since the client only knows if an error occurred, but not what type of error occurred. Being able to distinguish the types of errors that may occur during an **ASWebAuthenticationSession** or **SFAuthenticationSession** allows deciding whether an alert should be shown to the user or not. For instance, during an ASWebAuthentication- or SFAuthenticationSession users can share existing login information. Hence, the system requests the user's consent to access his login information. If the user decides to not grant permission, an the authentication session returns with either an **ASWebAuthenticationSessionError.canceledLogin** or **SFAuthenticationError.canceledLogin** error. In this case, it may be desired to handle the error silently client-wise.

### Proposed Solution:
Instead of only returning a boolean, a result (`Result<Void>`) is returned including the error if one occurred:

`func clearSession(federated: Bool, callback: @escaping (Result<Void>) -> Void)`

This allows clients to identify the reason why the operation did not succeed and to include dedicated error handling.

![dialog](https://user-images.githubusercontent.com/2820215/66543936-07512d80-eb37-11e9-9c63-248670572ad3.png)

cocojoe:
I believe when this was added it was only Safari based Auth available at the time. As you pointed out there is the additional outcome of cancellation during the permission dialog. 

The BOOL result is based upon the URL being loaded in the browser, so failure around this scenario should only be an issue if Auth0 itself is down or internet connectivity issues.

Any change to the Public API is breaking and would require a new major release of the library to support it, I don't think it would be justified in this case.

PR #315, opened 2019-10-10
Improve authentication session error handling

LinkAndreas:
### Changes

Improves client-wise error handling when removing an Auth0 Session by including the reason if `clearSession` did not succeed. Hence, the proposed solution changes:

`func clearSession(federated: Bool, callback: @escaping (Bool) -> Void)`        to
`func clearSession(federated: Bool, callback: @escaping (Result<Void>) -> Void)`

During an _ASWebAuthenticationSession_ or _SFAuthenticationSession_ the user might cancel the request to share existing login information. In this case, the client may desire to handle the error silently.

dialog

### References

The proposed solution has been implemented in the following pull request: #314 

Additional information about possible errors like **ASWebAuthenticationSessionError.canceledLogin** or **SFAuthenticationError.canceledLogin**:

- https://developer.apple.com/documentation/authenticationservices/aswebauthenticationsession
- https://developer.apple.com/documentation/safariservices/sfauthenticationsession

### Testing

Unit tests in `WebAuthSpec.swift` have been adjusted accordingly.

### Checklist

### DIFF
--- a/Auth0/SafariAuthenticationCallback.swift
+++ b/Auth0/SafariAuthenticationCallback.swift
@@ -31,18 +31,17 @@ import AuthenticationServices
 class SafariAuthenticationSessionCallback: NSObject, AuthTransaction {
 
     var state: String?
-    var callback: (Bool) -> Void = { _ in }
+    var callback: (Result<Void>) -> Void = { _ in }
 
     private var authSession: NSObject?
 
-    init(url: URL, schemeURL: String, callback: @escaping (Bool) -> Void) {
+    init(url: URL, schemeURL: String, callback: @escaping (Result<Void>) -> Void) {
         super.init()
         self.callback = callback
         #if canImport(AuthenticationServices)
         if #available(iOS 12.0, *) {
-            let authSession = ASWebAuthenticationSession(url: url, callbackURLScheme: schemeURL) { [unowned self] url, _ in
-                self.callback(url != nil)
-                TransactionStore.shared.clear()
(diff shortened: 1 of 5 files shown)


LinkAndreas:
@joshcanhelp Thanks for your quick response!

**Regarding your first remark:**
I updated unit tests included in `WebAuthSpec.swift` to handle the updated function signature. 

**Regarding your second remark:**
Clients using the SDK would have to update their implementation to handle `Result<Void>` rather than `Bool`. To avoid breaking changes, one could think about additionally supporting the old signature in `WebAuth.swift`:

```
func clearSession(federated: Bool, callback: @escaping (Bool) -> Void)
```
 
with the following implementation in `SafariWebAuth.swift`:

```
func clearSession(federated: Bool, callback: @escaping (Bool) -> Void) {
    clearSession(federated: federated) { (result: Result<Void>) in
        switch result {
        case .success:
            callback(true)

        case .failure:
            callback(false)
        }
    }
}
```

What are your thoughts? Feel free to come up with an alternative approach. I would be glad to update my implemantation according to your suggestions.

Thanks for your interest!

joshcanhelp:
@LinkAndreas - I asked about the tests as the coverage check is complaining about it being too low. 

In general, we'd like to avoid breaking changes so we can get positive changes released sooner rather than waiting for a major. If you're able to adjust this to do that, it's more likely we can get it merged in 👍

cocojoe:
See #314

Issue #309, opened 2019-09-18
id token not renewed during getCredentials call unless access token is expired.


malainp:
### Description

Description of the bug or feature request and why it's a problem. Consider including:

Got the same behavior as Android library described in [Android Issue 253](https://github.com/auth0/Auth0.Android/issues/253) in iOS App.

The IdToken doesn't refresh when expired. The token expiration time was set down to 1 hour and the library never called the `renew` method.

### Prerequisites

* [x] Did you check the documentation? [Add a link]
* [x] Did you check the [Auth0 Community](https://community.auth0.com/)?

### Environment

Please provide the following:

- Auth0.swift: 1.17.1
- iOS version: 12.4.1
- pod 'Auth0'
- Xcode version 10.3

### Reproduction

Like the Android issue, drop JWT expiration down to 5 minutes. Authenticate. Wait 5 minutes, then make a call to `credentials.` The idToken won't be renewed--no call to `authentication.renew` is ever made.

lbalmaceda:
Hi 👋@malainp we're already aware of this and tracking it on our backlog. I'll leave this open for you to track it as well.

JMCPH:
I'm experiencing the same issue, the user is logged out after the expiration time (24 hours) as the access token is expired, but the credentials manager does not renew the token.

Is there a way I can do this manually until a fix is comitted?

malainp:
@JMCPH if you have a request interceptor in your app you can evaluate the token life with [JWT library](https://github.com/auth0/JWTDecode.swift). If it is expired try to renew it manually like @camhart [did in android](https://github.com/auth0/Auth0.Android/commit/10e2c9cc0317c12a6667133c40f716f517b871e3#diff-3c60b1187240038264053066beedba42R145).

JMCPH:
@malainp Wouldn't it be easier just to save the email+password encrypted, and then just call the Auth0.login function instead?

malainp:
@JMCPH that's one way to do it but I think is better to follow the steps of Auth0 while solving the problem.

camhart:
@JMCPH You never want to save passwords in a recoverable state.  If you salted the password before you hashed it you won't be able to recover it.  Benefits of auth0 is you'll be given a limited use access token that can be refreshed or revoked without the need of adjusting passwords.  I'm sure there's more to it than that as well.

cocojoe:
Great to get some feedback on #319 which should address this.

PR #319, opened 2019-10-24
Renew tokens in Credentials Manager if either token has expired [SDK-999]

cocojoe:
### Changes

There are no changes to public API, changes are internal.
Existing behaviour was to validate only the `expires_in` as the use case was for renewing the AT. However, this has been expanded now to include the ID Token `exp` as part of the validation.

So if either the AT or IDT are expired then the credentials will be invalid and have to be renewed.

### References

#309 

#### Testing

### Checklist

### DIFF
--- a/Auth0/CredentialsManager.swift
+++ b/Auth0/CredentialsManager.swift
@@ -123,10 +123,9 @@ public struct CredentialsManager {
         guard
             let data = self.storage.data(forKey: self.storeKey),
             let credentials = NSKeyedUnarchiver.unarchiveObject(with: data) as? Credentials,
-            credentials.accessToken != nil,
-            let expiresIn = credentials.expiresIn
+            credentials.accessToken != nil
             else { return false }
-        return expiresIn > Date() || credentials.refreshToken != nil
+        return self.hasExpired(credentials) || credentials.refreshToken != nil
     }
 
     /// Retrieve credentials from keychain and yield new credentials using refreshToken if accessToken has expired
@@ -173,8 +172,8 @@ public struct CredentialsManager {
             let data = self.storage.data(forKey: self.storeKey),
             let credentials = NSKeyedUnarchiver.unarchiveObject(with: data) as? Credentials
             else { return callback(.noCredentials, nil) }
-        guard let expiresIn = credentials.expiresIn else { return callback(.noCredentials, nil) }
-        guard expiresIn < Date() else { return callback(nil, credentials) }
+        guard credentials.expiresIn != nil else { return callback(.noCredentials, nil) }
+        guard !self.hasExpired(credentials) else { return callback(nil, credentials) }
         guard let refreshToken = credentials.refreshToken else { return callback(.noRefreshToken, nil) }
 
         self.authentication.renew(withRefreshToken: refreshToken, scope: scope).start {
@@ -192,5 +191,21 @@ public struct CredentialsManager {
                 callback(.failedRefresh(error), nil)
             }
         }
+    }
+
+    private func hasExpired(_ credentials: Credentials) -> Bool {
+
+        if let expiresAT = credentials.expiresIn {
+            if expiresAT < Date() { return false }
+        }
+
+        if let idToken = credentials.idToken,
+            let jwt = decode(jwt: idToken),
+            let jwtExp = jwt["exp"] as? Double {
+                let expiresToken = Date(timeIntervalSince1970: jwtExp)
+                if expiresToken < Date() { return false }
+        }
+
+        return true
     }
 }
--- a/Auth0/CredentialsManager.swift
+++ b/Auth0/CredentialsManager.swift
@@ -125,7 +125,7 @@ public struct CredentialsManager {
             let credentials = NSKeyedUnarchiver.unarchiveObject(with: data) as? Credentials,
             credentials.accessToken != nil
             else { return false }
-        return self.hasExpired(credentials) || credentials.refreshToken != nil
+        return !self.hasExpired(credentials) || credentials.refreshToken != nil
     }
 
     /// Retrieve credentials from keychain and yield new credentials using refreshToken if accessToken has expired
@@ -173,7 +173,7 @@ public struct CredentialsManager {
             let credentials = NSKeyedUnarchiver.unarchiveObject(with: data) as? Credentials
             else { return callback(.noCredentials, nil) }
         guard credentials.expiresIn != nil else { return callback(.noCredentials, nil) }
-        guard !self.hasExpired(credentials) else { return callback(nil, credentials) }
+        guard self.hasExpired(credentials) else { return callback(nil, credentials) }
         guard let refreshToken = credentials.refreshToken else { return callback(.noRefreshToken, nil) }
 
         self.authentication.renew(withRefreshToken: refreshToken, scope: scope).start {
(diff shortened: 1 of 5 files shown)


Auth0/SafariWebAuth.swift
```diff
@@ -150,11 +150,11 @@ class SafariWebAuth: WebAuth {
     func newSafari(_ authorizeURL: URL, callback: @escaping (Result<Credentials>) -> Void) -> (SFSafariViewController, (Result<Credentials>) -> Void) {
         let controller = SFSafariViewController(url: authorizeURL)
         controller.modalPresentationStyle = safariPresentationStyle
-
```
cocojoe: The SwiftLinter picked these up.

Auth0/CredentialsManager.swift
```diff
@@ -193,4 +192,42 @@ public struct CredentialsManager {
...
+            let jwtExp = jwt["exp"] as? Double {
+                let expiresToken = Date(timeIntervalSince1970: jwtExp)
+                if expiresToken < Date() { return false }
+        }
+
+        return true
+    }
+}
+
+func decode(jwt: String) -> [String: Any]? {
```
joshcanhelp: Is there no JWT decoding in this library currently? Would it make sense to pull in whatever lib might be used for the OIDC improvements?

Auth0/CredentialsManager.swift
```diff
@@ -193,4 +192,42 @@ public struct CredentialsManager {
             }
         }
     }
+
+    private func hasExpired(_ credentials: Credentials) -> Bool {
+
+        if let expiresAT = credentials.expiresIn {
```
joshcanhelp:
```suggestion
        if let accessTokenExpires = credentials.expiresIn {
```

Auth0/CredentialsManager.swift
```diff
@@ -193,4 +192,42 @@ public struct CredentialsManager {
...
+            if expiresAT < Date() { return false }
```
joshcanhelp:
```suggestion
            if accessTokenExpires < Date() { return false }
```

Auth0/CredentialsManager.swift
```diff
@@ -193,4 +192,42 @@ public struct CredentialsManager {
...
+        }
+
+        if let idToken = credentials.idToken,
+            let jwt = decode(jwt: idToken),
```
joshcanhelp:
```suggestion
            let idTokenDecoded = decode(jwt: idToken),
```

Auth0/CredentialsManager.swift
```diff
@@ -193,4 +192,42 @@ public struct CredentialsManager {
...
+            let jwtExp = jwt["exp"] as? Double {
```
joshcanhelp:
```suggestion
            let idTokenExp = idTokenDecoded["exp"] as? Double {
```

Auth0/CredentialsManager.swift
```diff
@@ -193,4 +192,42 @@ public struct CredentialsManager {
...
+                let expiresToken = Date(timeIntervalSince1970: jwtExp)
```
joshcanhelp:
```suggestion
                let idTokenExpires = Date(timeIntervalSince1970: idTokenExp)
```

Auth0/CredentialsManager.swift
```diff
@@ -193,4 +192,42 @@ public struct CredentialsManager {
...
+                if expiresToken < Date() { return false }
```
joshcanhelp:
```suggestion
                if idTokenExpires < Date() { return false }
```

Auth0/CredentialsManager.swift
joshcanhelp: Just noticed this is from another class ... is this where it will be located later?

cocojoe:
@joshcanhelp turns out the logic was actually a little broken, I cleaned up naming a little.

Auth0/CredentialsManager.swift
```diff
@@ -193,4 +192,44 @@ public struct CredentialsManager {
             }
         }
     }
+
+    func hasExpired(_ credentials: Credentials) -> Bool {
```
cocojoe: Check logic here please, basically if either token has expired then it should return true

Auth0/CredentialsManager.swift
```diff
@@ -173,8 +172,8 @@ public struct CredentialsManager {
             let data = self.storage.data(forKey: self.storeKey),
             let credentials = NSKeyedUnarchiver.unarchiveObject(with: data) as? Credentials
             else { return callback(.noCredentials, nil) }
-        guard let expiresIn = credentials.expiresIn else { return callback(.noCredentials, nil) }
-        guard expiresIn < Date() else { return callback(nil, credentials) }
+        guard credentials.expiresIn != nil else { return callback(.noCredentials, nil) }
+        guard self.hasExpired(credentials) else { return callback(nil, credentials) }
```
cocojoe: if it has expired continue executing else return credentials we have stored

Auth0/CredentialsManager.swift
```diff
@@ -123,10 +123,9 @@ public struct CredentialsManager {
         guard
             let data = self.storage.data(forKey: self.storeKey),
             let credentials = NSKeyedUnarchiver.unarchiveObject(with: data) as? Credentials,
-            credentials.accessToken != nil,
-            let expiresIn = credentials.expiresIn
+            credentials.accessToken != nil
             else { return false }
-        return expiresIn > Date() || credentials.refreshToken != nil
+        return !self.hasExpired(credentials) || credentials.refreshToken != nil
```
cocojoe: if neither AT or IDT have expired OR we have a refresh token as we can use that to gain new tokens so technically we have valid credentials. Although the RT could be invalid, not possible to know until calling renew.

cocojoe:
@lbalmaceda please check logic aligns with your PR thanks

Auth0/CredentialsManager.swift
joshcanhelp: Why would we be using a refresh token for a new ID token? Particularly in the case of a native app.

Auth0/CredentialsManager.swift
joshcanhelp: Current logic looks correct (and tests should verify) ... but we could be much more clear here if we returned on the expired condition instead of setting variables:

```swift
if credentials.expiresIn <= Date() { 
    return true
}

var tokenDecoded = decode(jwt: credentials.idToken);
var exp = tokenDecoded["exp"] as? Double;

if Date(timeIntervalSince1970: exp) < Date() { 
    return true
}

return false
```

That's unlikely to be correct (let alone idiomatic) Swift 😆 but hopefully the message is clear.

Auth0/CredentialsManager.swift
lbalmaceda: @joshcanhelp To obtain a fresh copy of the user profile, I imagine. Either that or use the access token to call /userinfo and obtain it that way.

Auth0/CredentialsManager.swift
lbalmaceda: @cocojoe The `hasExpired` logic is correct, as josh pointed out. But I think the performance impact could be reduced a lot if instead of checking the exp date **of each token** and _decode_ the id token **every time** you call this method, you just save the exp date most close to expire at "save credentials" step, and then just have a single date comparison check on this function.

cocojoe: Yeah I had similar logic to above but I think a mix up in tests sent me down the wrong path, have corrected now.

I don't mind the idea of a single expiry time based upon lowest. Will leave it for a future internal refactor though.

PR #333, opened 2020-01-13
Improve OIDC compliance: add JWKS support

Widcket:
### Changes

This update improves the SDK support for **OpenID Connect (OIDC)**. In particular, it adds support for **JSON Web Key Set (JWKS)**. Additional PRs will further expand the SDK support for OIDC.

#### What’s being added in this PR
- JWKS endpoint. This PR does not include tests for this (will be in an upcoming PR), as they’re in the Authentication spec and that file includes changes that belong in other PRs, and I’m trying to keep the PR split simple.
- JWK and JWKS models.
- Logic for creating a SecKey from a modulus and exponent, and tests for that.
- Supported algorithms and some of its tests. The remaining tests on this will be in the next PR (signature validation).
- JWT and JWK generators for testing. This includes logic from Guardian for exporting the data from a SecKey and parsing it in order to extract the modulus and exponent. I added comments in the code explaining what it does as it’s hard to follow, since it involves pointers and low level stuff.
- [JWTDecode](https://github.com/auth0/JWTDecode.swift) as a dependency.

All this logic is isolated ATM, as the integration will come in the last PR.

##### Additions
- A public method in the Authentication API to retrieve the tenant’s JWKS data.
- The JWKS and JWK models.
- An extension to decode Base64URL data. This was made public as it’s the counterpart of the already-public a0_encodeBase64URLSafe().
- A0SHA and A0RSA, ObjC classes from Guardian that implement a wrapper around some CommonCrypto functionality.
##### Changes
None.

### Testing

- [X] This change adds unit test coverage
- [X] This change has been tested on the latest version of the platform/language or why not

### Checklist

- [X] I have read the [Auth0 general contribution guidelines](https://github.com/auth0/open-source-template/blob/master/GENERAL-CONTRIBUTING.md)
- [X] I have read the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md)
- [X] All existing and new tests complete without errors


Auth0/JWKS.swift
```diff
@@ -0,0 +1,91 @@
...
+        guard let modulus = rsaModulus?.a0_decodeBase64URLSafe(),
+            let exponent = rsaExponent?.a0_decodeBase64URLSafe() else { return nil }
+        let encodedKey = encodeRSAPublicKey(modulus: [UInt8](modulus), exponent: [UInt8](exponent))
+        if #available(iOS 10, OSX 10.12, tvOS 10, watchOS 3, *) {
+            return generateRSAPublicKey(from: encodedKey)
+        }
+        let tag = "com.auth0.tmp.RSAPublicKey"
+        let keychain = A0SimpleKeychain()
+        guard keychain.setRSAPublicKey(data: encodedKey, forKey: tag) else { return nil }
+        return keychain.keyRefOfRSAKey(withTag: tag).takeRetainedValue()
```
Widcket: This fails on macOS 10.11 (El Capitan), which happens to be the minimum supported version. So in that specific macOS version, the signature validation is skipped. Check https://github.com/auth0/Auth0.swift/pull/333/files#diff-2345af69c12dd8c93d94fdb71f422ba3R38.

In https://github.com/auth0/Auth0.swift/pull/333/files#diff-622b00adee0ac311c480485596339902R38 and https://github.com/auth0/Auth0.swift/pull/333/files#diff-b2bf319ef5a2f202615bba8e8cd28887R50, version checks were added to acomodate this.

Auth0/A0RSA.h
```diff
@@ -0,0 +1,35 @@
+// A0RSA.h
```
Widcket: Wrapper around CommonCrypto, so it can be used from any Swift version.

Auth0/A0RSA.m
```diff
@@ -0,0 +1,71 @@
+// A0RSA.m
```
Widcket: Wrapper around CommonCrypto, so it can be used from any Swift version.

Auth0/A0SHA.m
```diff
@@ -0,0 +1,55 @@
+// A0SHA.m
```
Widcket: Wrapper around CommonCrypto, so it can be used from any Swift version.

Auth0/A0SimpleKeychain+RSAPublicKey.swift
```diff
@@ -0,0 +1,40 @@
+// A0SimpleKeychain+RSAPublicKey.swift
```
Widcket: Extension to save a RSA key in the Keychain.

Auth0/Array+Encode.swift
```diff
@@ -0,0 +1,40 @@
+// Array+Encode.swift
```
Widcket: Extension to der-encode a byte array.

Auth0/Auth0Authentication.swift
```diff
@@ -59,7 +59,13 @@ struct Auth0Authentication: Authentication {
             ]
```
Widcket: Authentication API implementation.

Auth0/Authentication.swift
```diff
@@ -434,7 +434,7 @@ public protocol Authentication: Trackable, Loggable {
     func startPasswordless(phoneNumber: String, type: PasswordlessType, connection: String) -> Request<Void, AuthenticationError>
```
Widcket: Authentication API protocol.

Auth0/Handlers.swift
```diff
@@ -35,6 +35,22 @@ func plainJson(from response: Response<AuthenticationError>, callback: Request<[
     }
```
Widcket: JWKS request handler.

Auth0/JWKS.swift
```diff
@@ -0,0 +1,91 @@
+// JWKS.swift
```
Widcket: JWKS and JWK models.

Auth0/JWKS.swift
```diff
@@ -0,0 +1,91 @@
...
+        case certChain = "x5c"
+        case rsaModulus = "n"
+        case rsaExponent = "e"
+    }
+}
+
+extension JWK {
+    var rsaPublicKey: SecKey? {
+        guard let modulus = rsaModulus?.a0_decodeBase64URLSafe(),
+            let exponent = rsaExponent?.a0_decodeBase64URLSafe() else { return nil }
```
lbalmaceda: would this "else return nil" apply to the modulus or only to the exponent as well?

Auth0/JWKS.swift
```diff
@@ -0,0 +1,91 @@
...
+        let encodedKey = encodeRSAPublicKey(modulus: [UInt8](modulus), exponent: [UInt8](exponent))
+        if #available(iOS 10, OSX 10.12, tvOS 10, watchOS 3, *) {
+            return generateRSAPublicKey(from: encodedKey)
+        }
+        let tag = "com.auth0.tmp.RSAPublicKey"
```
lbalmaceda: should this be a constant?

Auth0/JWTAlgorithm.swift
```diff
@@ -0,0 +1,48 @@
+// JWTAlgorithm.swift
```
Widcket: Supported algorithms.

Auth0/String+URLSafe.swift
```diff
@@ -0,0 +1,35 @@
+// String+URLSafe.swift
```
Widcket: Extension to decode data encoded in Base64Url.

Widcket: It's public because it's the counterpart of the already-public `a0_encodeBase64URLSafe()`.

Auth0/JWTAlgorithm.swift
```diff
@@ -0,0 +1,48 @@
...
+enum JWTAlgorithm: String {
+    case rs256 = "RS256"
+    case hs256 = "HS256"
+    
+    func verify(_ jwt: JWT, using jwk: JWK) -> Bool {
+        let separator = "."
+        let parts = jwt.string.components(separatedBy: separator).dropLast().joined(separator: separator)
+        guard let data = parts.data(using: .utf8),
+            let signature = jwt.signature?.a0_decodeBase64URLSafe(),
+            !signature.isEmpty else { return false }
```
lbalmaceda: does this mean that if the signature is empty the verification fails gracefully?

Auth0/JWTAlgorithm.swift
```diff
@@ -0,0 +1,48 @@
...
+        switch self {
+        case .rs256:
+            #if os(OSX)
+            guard #available(OSX 10.12, *) else { return true }
+            #endif
+            guard let publicKey = jwk.rsaPublicKey,
+                let sha256 = A0SHA(algorithm: "sha256"),
```
lbalmaceda: do we need to pass the hashing algorithm, knowing we will only support one?

Auth0Tests/Extensions.swift
```diff
@@ -0,0 +1,45 @@
+// Extensions.swift
```
Widcket: Extensions part of the test targets that support the JWT and JWKS generators.

Auth0Tests/Generators.swift
```diff
@@ -0,0 +1,198 @@
+// Generators.swift
```
Widcket: JWT and JWKS generators.

Auth0/String+URLSafe.swift
```diff
@@ -0,0 +1,35 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import Foundation
+
+public extension String {
+    func a0_decodeBase64URLSafe() -> Data? {
+        let paddingLength = 4 - count % 4
```
lbalmaceda: what is this magic number? maybe move to a constant, give it a proper name/description?

Auth0Tests/Extensions.swift
```diff
@@ -0,0 +1,45 @@
...
+    func export() -> Data {
+        return SecKeyCopyExternalRepresentation(self, nil)! as Data
+    }
+}
+
+extension JWTAlgorithm {
+    func sign(value: Data, key: SecKey = TestKeys.rsaPrivate) -> Data {
+        switch self {
+        case .rs256:
+            let sha256 = A0SHA(algorithm: "sha256")!
```
lbalmaceda: consider moving `"sha256"` inside the `A0SHA` class

Auth0Tests/Extensions.swift
```diff
@@ -0,0 +1,45 @@
...
+            let rsa = A0RSA(key: key)!
+            
+            return rsa.sign(sha256.hash(value))
+        case .hs256: return value
```
lbalmaceda: If I got this right, signing with HS256 is not signing but returning the data untouched?

Auth0Tests/Generators.swift
```diff
@@ -0,0 +1,198 @@
...
+    
+    return encodeJWTPart(from: bodyDict)
+}
+
+func generateJWT(alg: String = JWTAlgorithm.rs256.rawValue,
+                 kid: String = "key123",
+                 iss: String? = "https://tokens-test.auth0.com/",
+                 sub: String? = "auth0|123456789",
+                 aud: [String]? = ["tokens-test-123"],
+                 exp: Date? = Date().addingTimeInterval(100000),
```
lbalmaceda: magic numbers

Auth0Tests/Generators.swift
```diff
@@ -0,0 +1,198 @@
...
+    else if let algorithm = JWTAlgorithm(rawValue: alg) {
+        signaturePart = algorithm.sign(value: signableParts.data(using: .utf8)!).a0_encodeBase64URLSafe()!
+    }
+    
+    return try! decode(jwt: "\(signableParts).\(signaturePart)")
+}
+
+// MARK: - JWK
+
+private func calculateLength(from bytes: UnsafePointer<UInt8>) -> (UnsafePointer<UInt8>, Int) {
```
lbalmaceda: you might want to at least comment all these helper functions. With the byte manipulation in place, they are difficult to follow.

Auth0Tests/Generators.swift
```diff
@@ -0,0 +1,198 @@
...
+    guard bytes.pointee == 0x02 else { return nil }
+    
+    let (valueBytes, valueLength) = calculateLength(from: bytes + 1)
+    let data = Data(bytes: valueBytes, count: valueLength)
+    let pointer = valueBytes + valueLength
+    
+    return (pointer, data)
+}
+
+func generateRSAJWK(from publicKey: SecKey = TestKeys.rsaPublic, keyId: String = "key123") -> JWK {
```
lbalmaceda: the second time I see this `key123` value. Seems a good candidate for a constant value

Auth0Tests/JWKSpec.swift
```diff
@@ -0,0 +1,82 @@
...
+                    it("should generate a RSA public key") {
+                        let publicKey = jwk.rsaPublicKey!
+                        let keyAttributes = SecKeyCopyAttributes(publicKey) as! [String: Any]
+                        
+                        expect(keyAttributes[String(kSecAttrKeyType)] as? String).to(equal(String(kSecAttrKeyTypeRSA)))
+                    }
+                }
+            }
+            
+            context("unsuccessful generation") {
```
lbalmaceda: are exponent and modulus the only reasons why it could fail? What about having a different usage or algorithm? I don't expect all the cases to be covered, but those 2 I'd say are a must

Auth0Tests/JWTAlgorithmSpec.swift
```diff
@@ -0,0 +1,77 @@
...
+            context("RS256") {
+                let alg = "RS256"
+                
+                it("should return true with a correct RS256 signature") {
+                    let jwt = generateJWT(alg: alg)
+                    
+                    expect(JWTAlgorithm.rs256.verify(jwt, using: jwk)).to(beTrue())
+                }
+                
+                it("should return false with an empty signature") {
```
lbalmaceda: this answers one of my questions above 👍

Auth0Tests/JWTAlgorithmSpec.swift
```diff
@@ -0,0 +1,77 @@
...
+                    
+                    expect(JWTAlgorithm.hs256.verify(jwt, using: jwk)).to(beTrue())
+                }
+                
+                it("should return false with an empty signature") {
+                    let jwt = generateJWT(alg: alg, signature: "")
+                    
+                    expect(JWTAlgorithm.hs256.verify(jwt, using: jwk)).to(beFalse())
+                }
+            }
```
lbalmaceda: what file is checking that any other algorithm name returns false as well?

Auth0/JWKS.swift
Widcket: Would apply to both of them. It's equivalent to the following if:
```swift
    if rsaModulus?.a0_decodeBase64URLSafe() != nil && rsaExponent?.a0_decodeBase64URLSafe() != nil {}
```

Widcket: Would apply to both of them. It's equivalent to the following if:
```swift
    if rsaModulus?.a0_decodeBase64URLSafe() != nil && rsaExponent?.a0_decodeBase64URLSafe() != nil {
        let modulus = rsaModulus!.a0_decodeBase64URLSafe()!
        let exponent = rsaExponent!.a0_decodeBase64URLSafe()!
}

```

Auth0Tests/JWTAlgorithmSpec.swift
Widcket: It's going to be in the next PR.

Auth0/JWTAlgorithm.swift
Widcket: Yes

Auth0/JWTAlgorithm.swift
Widcket: That's how the implementation from Guardian is written, and I didn't want to touch those crypto ObjC files.

Auth0/JWKS.swift
Widcket: I don't think so, since this is only used in this particular computed property of this particular extension.

Auth0Tests/Generators.swift
Widcket: I'll add some comments.

Auth0Tests/Extensions.swift
Widcket: Yes, exactly.

Auth0Tests/JWKSpec.swift
Widcket: The usage is not used at all in the `.rsaPublicKey` implementation. All that's needed is valid values in the properties `rsaModulus` and `rsaExponent`.

Auth0Tests/Generators.swift
Widcket: All the default param values in that method are sane defaults for the tests. What that particular line does is create a date 100000 milliseconds in the future.

Auth0Tests/Generators.swift
Widcket: This method calculates the length of a given DER triplet. It starts at the triplet's base address, the proceeds from there.

Widcket: It returns a pointer to the end of the triplet plus its length.

Auth0Tests/Generators.swift
```diff
@@ -0,0 +1,198 @@
...
+    guard bytes.pointee > 0x7f else { return (bytes + 1, Int(bytes.pointee)) }
+
+    let count = Int(bytes.pointee & 0x7f)
+    let length = (1...count).reduce(0) { ($0 << 8) + Int(bytes[$1]) }
+    let pointer = bytes + (1 + count)
+    
+    return (pointer, length)
+}
+
+private func extractData(from bytes: UnsafePointer<UInt8>) -> (UnsafePointer<UInt8>, Data)? {
```
Widcket: This method extracts the data portion from a given DER triplet. It starts at the triplet's base address, then proceeds from there. It returns a pointer to the end of the triplet plus its data.

Auth0Tests/Generators.swift
Widcket: Done.

Auth0Tests/Generators.swift
```diff
@@ -0,0 +1,198 @@
...
+    guard bytes.pointee == 0x02 else { return nil }
```
Widcket: Checks that this is an integer triplet.

Auth0Tests/JWKSpec.swift
Widcket: I can add a check in the guard clause of `.rsaPublicKey ` to return nil if the algorithm != "RSA" and usage != "sig".

Widcket: After all, it's an internal (package-private) property.

Auth0/String+URLSafe.swift
Widcket: Base64 requires the string length to be a multiple of four. I can add a let constant in the line above with it.

Auth0Tests/Generators.swift
Widcket: I'll add a constant.

Auth0Tests/Extensions.swift
Widcket: I'll add another initializer to `A0SHA` with `sha256 ` as the default alg.

Auth0/String+URLSafe.swift
Widcket: Added in 123af59498094926539a355378c5ce78efa7a1bb.

Auth0Tests/Extensions.swift
Widcket: Done in 123af59498094926539a355378c5ce78efa7a1bb.

Auth0Tests/Generators.swift
Widcket: Comment added in 123af59498094926539a355378c5ce78efa7a1bb.

Auth0Tests/Generators.swift
Widcket: Comment added in 123af59498094926539a355378c5ce78efa7a1bb.

Auth0Tests/Generators.swift
Widcket: Comment added in 123af59498094926539a355378c5ce78efa7a1bb.

Auth0Tests/Generators.swift
Widcket: Added in 123af59498094926539a355378c5ce78efa7a1bb.

Auth0Tests/JWKSpec.swift
Widcket: Done in 123af59498094926539a355378c5ce78efa7a1bb. Thanks for spotting it!

cocojoe:
@Widcket The JWTDecode framework files shouldn't be part of the repo, just as SimpleKeyChain framework files are not.

Widcket:
@cocojoe removed.

Auth0Tests/Generators.swift
lbalmaceda: I get that, but let's give it a name or at least a comment on the side telling what is the intention. e.g. `// tomorrow` or `addingTimeInterval(ONE_DAY_IN_MS)`. Helps improving the readability

Auth0Tests/JWKSpec.swift
lbalmaceda: @Widcket for clarification. The `use` key is optional according to the spec, but when present it should be used to check whether the key you are reading can be used for the operation you want to perform. 

For this particular implementation, it means that if `use!='sig'` then the key is invalid and should not be used.

Widcket: Then I should update the implementation, because I added a check in the guard clause testing if usage == "sig". I'll need to split it into a separate check, because `use` is optional, and only if present should be part of the validation. Thanks for the clarification.

Auth0Tests/JWTAlgorithmSpec.swift
lbalmaceda: Should that be noted on a `//TODO` comment for posterity?

Auth0/JWKS.swift
```diff
@@ -0,0 +1,94 @@
...
+        case certThumbprint = "x5t"
+        case certChain = "x5c"
+        case rsaModulus = "n"
+        case rsaExponent = "e"
+    }
+}
+
+extension JWK {
+    var rsaPublicKey: SecKey? {
+        if let usage = usage, usage != "sig" { return nil }
```
Widcket: @lbalmaceda updated to reflect the fact that `use` is optional.

Auth0Tests/JWTAlgorithmSpec.swift
Widcket: No need, since it's already done.

Auth0/Authentication.swift
```diff
@@ -434,7 +434,7 @@ public protocol Authentication: Trackable, Loggable {
...
 
     /**
-     Returns token information by performing a request to /tokeninfo endpoint
```
cocojoe: again would avoid PR noise with formatting in the future

Widcket:
> Can you update the description as discussed internally as lot going on. Also new lib added.

Done.

PR #336, opened 2020-01-15
Added missing watchOS shared scheme

### LINKED ISSUE #334: Carthage: watchOS missing shared scheme.
### Description

According to the main README.md Auth0 supports watchOS 2.0+, however, Carthage is missing the shared build scheme, which produces the following output when trying to target watchOS:

> Dependency "Auth0.swift" has no shared framework schemes for any of the platforms: watchOS

As seen at the following link the shared schema isn't present: https://github.com/auth0/Auth0.swift/tree/master/Auth0.xcodeproj/xcshareddata/xcschemes 

### Reproduction

> carthage bootstrap Auth0.swift --use-ssh --platform watchOS

### Environment

- **Version of this library used:** 1.20.0
- **Version of the platform or framework used, if applicable:** watchOS 6
- **Other modules/plugins/libraries that might be involved:**  Carthage 0.34.0

Widcket:
### Changes

- Added missing shared scheme for watchOS. [Schemes](https://developer.apple.com/library/archive/documentation/IDEs/Conceptual/xcode_guide-continuous_integration/ConfigureBots.html#//apple_ref/doc/uid/TP40013292-CH9-SW3) that are not shared cannot be built outside of Xcode. Since Carthage clones and builds the dependencies on its own, the SDK's schemes must be shared or [Carthage won't be able to build the SDK](https://github.com/Carthage/Carthage#share-your-xcode-schemes).

### References

- Fixes #334 

### Testing

### Checklist

### DIFF
--- a/Auth0.xcodeproj/xcshareddata/xcschemes/Auth0.watchOS.xcscheme
+++ b/Auth0.xcodeproj/xcshareddata/xcschemes/Auth0.watchOS.xcscheme
@@ -1,0 +1,67 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Scheme
+   LastUpgradeVersion = "1130"
+   version = "1.3">
+   <BuildAction
+      parallelizeBuildables = "YES"
+      buildImplicitDependencies = "YES">
+      <BuildActionEntries>
+         <BuildActionEntry
+            buildForTesting = "YES"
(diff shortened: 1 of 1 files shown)


Widcket:
@stevehobbsdev just updated the description and added some useful links. Thanks for pointing that out.

PR #344, opened 2020-01-27
Improve OIDC compliance: integrate IDTV

Widcket:
### Changes

This update improves the SDK support for **OpenID Connect (OIDC)**. In particular, it hooks the ID Token validation logic in the authentication flows.

#### What’s being added in this PR

- Tests for the `jwks` endpoint.

#### What’s being changed in this PR

- Removed the `decodeJwt(_:)` method from `CredentialsManager.swift` and replaced it with  calls to `JWTDecode` logic.
- Changed the signature of the (internal) method `validate(idToken:context:signatureValidator:claimsValidator:callback:)` to use `for` as the external name for the parameter `context`, for the sake of API expressiveness.
- Changed the signature of the initialisers of the (internal) structs `ImplicitGrant` and `PKCE` in order to pass the parameters needed to perform ID Token validation.
- Updated existing tests to account for ID Token validation.

#### Public surface area

- **Additions:** the methods `leeway(_:)` and `maxAge(_:)` in the WebAuth builder.
- **Changes:** none.

### Testing

### Checklist

### DIFF
--- a/Auth0/CredentialsManager.swift
+++ b/Auth0/CredentialsManager.swift
@@ -22,6 +22,7 @@
 
 import Foundation
 import SimpleKeychain
+import JWTDecode
 #if os(iOS)
 import LocalAuthentication
 #endif
@@ -129,7 +130,7 @@ public struct CredentialsManager {
     }
 
     /// Retrieve credentials from keychain and yield new credentials using refreshToken if accessToken has expired
-    /// otherwise the retrieved credentails will be returned as they have not expired. Renewed credentials will be 
+    /// otherwise the retrieved credentails will be returned as they have not expired. Renewed credentials will be
     /// stored in the keychain.
     ///
     ///
@@ -194,40 +195,14 @@ public struct CredentialsManager {
     }
 
     func hasExpired(_ credentials: Credentials) -> Bool {
-
         if let expiresIn = credentials.expiresIn {
             if expiresIn < Date() { return true }
         }
 
-        if let token = credentials.idToken,
-            let tokenDecoded = decodeJwt(token),
-            let exp = tokenDecoded["exp"] as? Double {
-            if Date(timeIntervalSince1970: exp) < Date() { return true }
+        if let token = credentials.idToken, let jwt = try? decode(jwt: token) {
+            return jwt.expired
         }
 
         return false
     }
-}
-
-// To avoid collision with JWTDecode's decode function. This one will be removed in the last PR
-func decodeJwt(_ jwt: String) -> [String: Any]? {
-    let parts = jwt.components(separatedBy: ".")
-    guard parts.count == 3 else { return nil }
-    var base64 = parts[1]
-        .replacingOccurrences(of: "-", with: "+")
-        .replacingOccurrences(of: "_", with: "/")
-    let length = Double(base64.lengthOfBytes(using: String.Encoding.utf8))
-    let requiredLength = 4 * ceil(length / 4.0)
-    let paddingLength = requiredLength - length
-    if paddingLength > 0 {
-        let padding = "".padding(toLength: Int(paddingLength), withPad: "=", startingAt: 0)
-        base64 += padding
-    }
-
-    guard
-        let bodyData = Data(base64Encoded: base64, options: .ignoreUnknownCharacters)
-        else { return nil }
-
-    let json = try? JSONSerialization.jsonObject(with: bodyData, options: [])
-    return json as? [String: Any]
 }
--- a/Auth0/IDTokenValidatorContext.swift
+++ b/Auth0/IDTokenValidatorContext.swift
@@ -34,22 +34,22 @@ struct IDTokenValidatorContext: IDTokenSignatureValidatorContext, IDTokenClaimsV
          audience: String,
          jwksRequest: Request<JWKS, AuthenticationError>,
          leeway: Int,
-         nonce: String?,
-         maxAge: Int?) {
+         maxAge: Int?,
+         nonce: String?) {
         self.issuer = issuer
         self.audience = audience
         self.jwksRequest = jwksRequest
         self.leeway = leeway
-        self.nonce = nonce
         self.maxAge = maxAge
+        self.nonce = nonce
     }
 
-    init(authentication: Authentication, leeway: Int, nonce: String?, maxAge: Int?) {
+    init(authentication: Authentication, leeway: Int, maxAge: Int?, nonce: String?) {
         self.init(issuer: "\(authentication.url.absoluteString)/",
                   audience: authentication.clientId,
                   jwksRequest: authentication.jwks(),
                   leeway: leeway,
-                  nonce: nonce,
-                  maxAge: maxAge)
+                  maxAge: maxAge,
+                  nonce: nonce)
     }
 }
--- a/Auth0/IDTokenValidatorContext.swift
+++ b/Auth0/IDTokenValidatorContext.swift
@@ -45,7 +45,7 @@ struct IDTokenValidatorContext: IDTokenSignatureValidatorContext, IDTokenClaimsV
     }
 
     init(authentication: Authentication, leeway: Int, nonce: String?, maxAge: Int?) {
-        self.init(issuer: authentication.url.host!,
+        self.init(issuer: "\(authentication.url.absoluteString)/",
                   audience: authentication.clientId,
                   jwksRequest: authentication.jwks(),
                   leeway: leeway,
--- a/Auth0/OAuth2Grant.swift
+++ b/Auth0/OAuth2Grant.swift
@@ -21,6 +21,7 @@
 // THE SOFTWARE.
 
 import Foundation
+import JWTDecode
 
 protocol OAuth2Grant {
     var defaults: [String: String] { get }
@@ -30,11 +31,21 @@ protocol OAuth2Grant {
 
 struct ImplicitGrant: OAuth2Grant {
 
+    let authentication: Authentication
     let defaults: [String: String]
     let responseType: [ResponseType]
+    let leeway: Int
+    let maxAge: Int?
 
-    init(responseType: [ResponseType] = [.token], nonce: String? = nil) {
+    init(authentication: Authentication,
+         responseType: [ResponseType] = [.token],
+         leeway: Int,
+         nonce: String? = nil,
+         maxAge: Int? = nil) {
+        self.authentication = authentication
         self.responseType = responseType
+        self.leeway = leeway
+        self.maxAge = maxAge
         if let nonce = nonce {
             self.defaults = ["nonce": nonce]
         } else {
@@ -43,15 +54,18 @@ struct ImplicitGrant: OAuth2Grant {
     }
 
     func credentials(from values: [String: String], callback: @escaping (Result<Credentials>) -> Void) {
-        guard validate(responseType: self.responseType, token: values["id_token"], nonce: self.defaults["nonce"]) else {
-            return callback(.failure(error: WebAuthError.invalidIdTokenNonce))
+        let responseType = self.responseType
+        let validatorContext = IDTokenValidatorContext(authentication: authentication,
+                                                       leeway: leeway,
+                                                       nonce: self.defaults["nonce"],
+                                                       maxAge: maxAge)
+        validate(idToken: values["id_token"], for: responseType, with: validatorContext) { error in
+            if let error = error { return callback(.failure(error: error)) }
+            guard !responseType.contains(.token) || values["access_token"] != nil else {
+                return callback(.failure(error: WebAuthError.missingAccessToken))
+            }
+            callback(.success(result: Credentials(json: values as [String: Any])))
         }
-
-        guard !responseType.contains(.token) || values["access_token"] != nil else {
-            return callback(.failure(error: WebAuthError.missingAccessToken))
-        }
-
-        callback(.success(result: Credentials(json: values as [String: Any])))
     }
 
     func values(fromComponents components: URLComponents) -> [String: String] {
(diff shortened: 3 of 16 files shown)


Auth0/IDTokenValidatorContext.swift
```diff
@@ -45,7 +45,7 @@ struct IDTokenValidatorContext: IDTokenSignatureValidatorContext, IDTokenClaimsV
     }
 
     init(authentication: Authentication, leeway: Int, nonce: String?, maxAge: Int?) {
-        self.init(issuer: authentication.url.host!,
+        self.init(issuer: "\(authentication.url.absoluteString)/",
```
joshcanhelp: **Question:** Can you tell me a bit about what `authentication.url.absoluteString` provides?

Auth0/OAuth2Grant.swift
```diff
@@ -30,11 +31,21 @@ protocol OAuth2Grant {
 
 struct ImplicitGrant: OAuth2Grant {
 
+    let authentication: Authentication
     let defaults: [String: String]
     let responseType: [ResponseType]
-
-    init(responseType: [ResponseType] = [.token], nonce: String? = nil) {
+    let leeway: Int
```
joshcanhelp: Shouldn't this be optional as well, with a default set?

Auth0/OAuth2Grant.swift
```diff
@@ -30,11 +31,21 @@ protocol OAuth2Grant {
...
+    let maxAge: Int?
+
+    init(authentication: Authentication,
+         responseType: [ResponseType] = [.token],
```
joshcanhelp: **Question:** Can you tell me what this is doing? Very unique syntax for me!

Auth0/OAuth2Grant.swift
```diff
@@ -43,15 +54,18 @@ struct ImplicitGrant: OAuth2Grant {
...
 
     func credentials(from values: [String: String], callback: @escaping (Result<Credentials>) -> Void) {
-        guard validate(responseType: self.responseType, token: values["id_token"], nonce: self.defaults["nonce"]) else {
-            return callback(.failure(error: WebAuthError.invalidIdTokenNonce))
+        let responseType = self.responseType
+        let validatorContext = IDTokenValidatorContext(authentication: authentication,
+                                                       leeway: leeway,
+                                                       nonce: self.defaults["nonce"],
+                                                       maxAge: maxAge)
+        validate(idToken: values["id_token"], for: responseType, with: validatorContext) { error in
```
joshcanhelp: I'm having a tough time understanding what this block is doing. Can you explain? Is it failing if there is no access token in the response? I see that was the behavior before as well but seems funny to fail without one. Definitely might be something I'm misunderstanding.

Auth0/OAuth2Grant.swift
```diff
@@ -67,50 +81,99 @@ struct PKCE: OAuth2Grant {
...
+    let maxAge: Int?
+
+    init(authentication: Authentication,
+         redirectURL: URL,
+         generator: A0SHA256ChallengeGenerator = A0SHA256ChallengeGenerator(),
+         reponseType: [ResponseType] = [.code],
+         nonce: String? = nil,
+         leeway: Int,
+         maxAge: Int? = nil) {
+        self.init(authentication: authentication,
```
joshcanhelp: **Nit:** Indent?

Auth0/OAuth2Grant.swift
```diff
@@ -67,50 +81,99 @@ struct PKCE: OAuth2Grant {
...
+        let maxAge = self.maxAge
+        let isHybridFlow = responseType.contains(.idToken)
+        let verifier = self.verifier
+        let redirectUrlString = self.redirectURL.absoluteString
+        let clientId = authentication.clientId
+        let validatorContext = IDTokenValidatorContext(authentication: authentication,
+                                                       leeway: leeway,
+                                                       nonce: nonce,
+                                                       maxAge: maxAge)
+        validate(idToken: idToken, for: responseType, with: validatorContext) { error in
```
joshcanhelp: **Question:** Is this validating the ID token before proceeding with the code exchange?

Auth0/OAuth2Grant.swift
```diff
@@ -67,50 +81,99 @@ struct PKCE: OAuth2Grant {
...
+            if let error = error { return callback(.failure(error: error)) }
+            authentication
+                .tokenExchange(withCode: code, codeVerifier: verifier, redirectURI: redirectUrlString)
+                .start { result in
+                    switch result {
+                    case .failure(let error as AuthenticationError) where error.description == "Unauthorized":
+                        // Special case for PKCE when the correct method for token endpoint authentication is not set (it should be None)
+                        let webAuthError = WebAuthError.pkceNotAllowed("Unable to complete authentication with PKCE. PKCE support can be enabled by setting Application Type to 'Native' and Token Endpoint Authentication Method to 'None' for this app at 'https://manage.auth0.com/#/applications/\(clientId)/settings'.")
```
joshcanhelp: Is this level of error detail pretty common in this library? I'm a little worried about this level of UI description in library could. If the label changes in the dashboard, will we know/remember to change this?

joshcanhelp: Also ... "Application Type to 'Native'" doesn't really do anything, I would stick to the token authentication part (if this description is staying).

Auth0/OAuth2Grant.swift
```diff
@@ -67,50 +81,99 @@ struct PKCE: OAuth2Grant {
...
+        let redirectUrlString = self.redirectURL.absoluteString
+        let clientId = authentication.clientId
+        let validatorContext = IDTokenValidatorContext(authentication: authentication,
+                                                       leeway: leeway,
+                                                       nonce: nonce,
+                                                       maxAge: maxAge)
+        validate(idToken: idToken, for: responseType, with: validatorContext) { error in
+            if let error = error { return callback(.failure(error: error)) }
+            authentication
+                .tokenExchange(withCode: code, codeVerifier: verifier, redirectURI: redirectUrlString)
```
joshcanhelp: I assume `tokenExchange` is an existing method? Since we will/do have token exchange in this library, `codeExchange` or similar would be more helpful/accurate.

Auth0/OAuth2Grant.swift
```diff
@@ -67,50 +81,99 @@ struct PKCE: OAuth2Grant {
...
+                .start { result in
+                    switch result {
+                    case .failure(let error as AuthenticationError) where error.description == "Unauthorized":
+                        // Special case for PKCE when the correct method for token endpoint authentication is not set (it should be None)
+                        let webAuthError = WebAuthError.pkceNotAllowed("Unable to complete authentication with PKCE. PKCE support can be enabled by setting Application Type to 'Native' and Token Endpoint Authentication Method to 'None' for this app at 'https://manage.auth0.com/#/applications/\(clientId)/settings'.")
+                        return callback(.failure(error: webAuthError))
+                    case .failure(let error): return callback(.failure(error: error))
+                    case .success(let credentials):
+                        guard !isHybridFlow else {
```
joshcanhelp: **Nit:** `guard !isHybridFlow else` is confusing ... can you switch this around to be more clear? Like `if this is the hybrid flow then do this thing and return`

Auth0/OAuth2Grant.swift
```diff
@@ -67,50 +81,99 @@ struct PKCE: OAuth2Grant {
...
+                            let newCredentials = Credentials(accessToken: credentials.accessToken,
+                                                             tokenType: credentials.tokenType,
+                                                             idToken: idToken,
+                                                             refreshToken: credentials.refreshToken,
+                                                             expiresIn: credentials.expiresIn,
+                                                             scope: credentials.scope)
+                            return callback(.success(result: newCredentials))
+                        }
+                        // Code flow
```
joshcanhelp: **Nit:** Technically** all of this is "code flow" of some kind.

Auth0/SafariWebAuth.swift
```diff
@@ -196,12 +211,21 @@ class SafariWebAuth: WebAuth {
...
+        if self.responseType.contains([.code]) { // both Hybrid and Code flow
             authentication.logger = self.logger
-            return PKCE(authentication: authentication, redirectURL: redirectURL, reponseType: self.responseType, nonce: self.nonce)
+            return PKCE(authentication: authentication,
+                        redirectURL: redirectURL,
+                        reponseType: self.responseType,
+                        nonce: self.nonce,
+                        leeway: self.leeway,
+                        maxAge: self.maxAge)
         } else {
```
joshcanhelp: **Nit:** `else` not required

Auth0/OAuth2Grant.swift
```diff
@@ -21,6 +21,7 @@
 // THE SOFTWARE.
```
Widcket: Please check the business logic in this file @lbalmaceda.

Auth0/SafariWebAuth.swift
```diff
@@ -196,12 +211,21 @@ class SafariWebAuth: WebAuth {
...
-            return ImplicitGrant(responseType: self.responseType, nonce: self.nonce)
+            return ImplicitGrant(authentication: authentication,
```
joshcanhelp: **Question:** Does an implicit grant in this SDK use fragment response type or query?

Auth0/CredentialsManager.swift
```diff
@@ -194,40 +195,14 @@ public struct CredentialsManager {
...
         if let expiresIn = credentials.expiresIn {
             if expiresIn < Date() { return true }
         }
 
-        if let token = credentials.idToken,
-            let tokenDecoded = decodeJwt(token),
-            let exp = tokenDecoded["exp"] as? Double {
-            if Date(timeIntervalSince1970: exp) < Date() { return true }
+        if let token = credentials.idToken, let jwt = try? decode(jwt: token) {
+            return jwt.expired
```
Widcket: See the implementation in https://github.com/auth0/JWTDecode.swift/blob/master/JWTDecode/JWTDecode.swift#L66.

Auth0Tests/OAuth2GrantSpec.swift
```diff
@@ -221,11 +215,25 @@ class OAuth2GrantSpec: QuickSpec {
                 }
             }
 
-            it("shoud fail credentials no nonce") {
-                pkce = PKCE(authentication: authentication, redirectURL: redirectURL, verifier: verifier, challenge: challenge, method: method, responseType: response)
+            it("should fail with no nonce") {
```
joshcanhelp: What does not have a nonce, the token or something else? The token here is going to come back as malformed. Is there a specific error or error message you could look for to make sure it's failing the way you expect?

Auth0/OAuth2Grant.swift
Widcket: If it has a default value, then it shouldn't be optional (because it will never be nil).

Auth0/OAuth2Grant.swift
Widcket: This is declaring a parameter of type `[ResponseType]`, which means 'array of `ResponseType`', and setting a default value of `[.token]`.

Auth0/IDTokenValidatorContext.swift
Widcket: It provides the absolute string for the `URL`. See https://developer.apple.com/documentation/foundation/url/1779984-absolutestring.

Auth0/OAuth2Grant.swift
Widcket: Yes, it's an existing (and public) method. Changing it would be a breaking change.

Auth0/OAuth2Grant.swift
Widcket: That's an existing case (check the red part of the diff). Not sure it would be a good idea to revisit it now.

Auth0/OAuth2Grant.swift
Widcket: How would you rephrase the comment so it's more accurate and useful?

Widcket: Please feel free to add suggestions.

Auth0Tests/OAuth2GrantSpec.swift
Widcket: The validator would not have a nonce to compare, in this case. Though the token should be a valid one. Thanks, will rephrase it and add a valid token.

Auth0/SafariWebAuth.swift
Widcket: Will remove it. Thanks for pointing it out.

Auth0/OAuth2Grant.swift
Widcket: That's precisely what the `guard` clause is for.

Auth0/OAuth2Grant.swift
Widcket: Yes, exactly.

Auth0/OAuth2Grant.swift
Widcket: It's 4 spaces to the right of `init`, so it's correct.

Auth0/OAuth2Grant.swift
lbalmaceda: Yes, this is currently failing for `response_type = id_token` for example. Shouldn't fail, it should accept all implicit flow scenarios, which are 3 in total: the combination of `id_token` and `token`.

I suggest moving to an `&&` instead and negate the condition, something like this:
```js
if response_type.contains("token") && !values.access_token -> fail
```
Anyway, something to fix on a separate PR as this was existing behavior.

with @Widcket now. cc @cocojoe

Auth0/OAuth2Grant.swift
```diff
@@ -67,50 +81,99 @@ struct PKCE: OAuth2Grant {
...
-
-    init(authentication: Authentication, redirectURL: URL, generator: A0SHA256ChallengeGenerator = A0SHA256ChallengeGenerator(), reponseType: [ResponseType] = [.code], nonce: String? = nil) {
-        self.init(authentication: authentication, redirectURL: redirectURL, verifier: generator.verifier, challenge: generator.challenge, method: generator.method, responseType: reponseType, nonce: nonce)
+    let leeway: Int
+    let maxAge: Int?
+
+    init(authentication: Authentication,
+         redirectURL: URL,
+         generator: A0SHA256ChallengeGenerator = A0SHA256ChallengeGenerator(),
+         reponseType: [ResponseType] = [.code],
```
lbalmaceda: typo 🧐

Auth0/OAuth2Grant.swift
```diff
@@ -67,50 +81,99 @@ struct PKCE: OAuth2Grant {
...
+         nonce: String? = nil,
+         leeway: Int,
```
lbalmaceda: please, reorder the optional params to the end of the method

Auth0/OAuth2Grant.swift
```diff
@@ -67,50 +81,99 @@ struct PKCE: OAuth2Grant {
...
     }
 
-    init(authentication: Authentication, redirectURL: URL, verifier: String, challenge: String, method: String, responseType: [ResponseType], nonce: String? = nil) {
+    init(authentication: Authentication,
+         redirectURL: URL,
+         verifier: String,
+         challenge: String,
+         method: String,
+         responseType: [ResponseType],
+         nonce: String? = nil,
```
lbalmaceda: optional param ordering

Auth0/OAuth2Grant.swift
```diff
@@ -67,50 +81,99 @@ struct PKCE: OAuth2Grant {
...
-                } else {
-                    callback(result)
-                }
+        let idToken = values["id_token"]
+        let responseType = self.responseType
+        let authentication = self.authentication
+        let leeway = self.leeway
+        let nonce = self.defaults["nonce"]
+        let maxAge = self.maxAge
+        let isHybridFlow = responseType.contains(.idToken)
```
lbalmaceda: hybrid flow examples:
- token id_token code
- token code
- id_token code

You could rename to `frontChannelIdTokenExpected`

Auth0/OAuth2Grant.swift
lbalmaceda: Let's negate this conditional and swap the IF's bodies. This way, if front channel is expected, validate and return. "else" (no need to add else) return the received credentials

Auth0/OAuth2Grant.swift
```diff
@@ -119,15 +182,16 @@ struct PKCE: OAuth2Grant {
...
-private func validate(responseType: [ResponseType], token: String?, nonce: String?) -> Bool {
-    guard responseType.contains(.idToken) else { return true }
-    guard
-        let expectedNonce = nonce,
-        let token = token
-        else { return false }
-    let claims = decodeJwt(token)
-    let actualNonce = claims?["nonce"] as? String
-    return actualNonce == expectedNonce
+private func validate(idToken: String?,
```
lbalmaceda: add a comment: this method will skip the validation if the response type does not contain "id_token"

Auth0/SafariWebAuth.swift
lbalmaceda: According to the matrix, nope. It doesn't support form_post.

Auth0/WebAuth.swift
```diff
@@ -162,8 +162,8 @@ public protocol WebAuth: Trackable, Loggable {
     /// - Returns: the same WebAuth instance to allow method chaining
     func responseType(_ response: [ResponseType]) -> Self
 
-    /// Add nonce paramater for authentication, this is a requirement for
-    /// when response type .id_token is specified.
+    /// Add `nonce` paramater for authentication, this is a requirement
```
lbalmaceda: typo. parameter

Auth0/WebAuth.swift
```diff
@@ -176,6 +176,20 @@ public protocol WebAuth: Trackable, Loggable {
     /// - Returns: the same WebAuth instance to allow method chaining
     func audience(_ audience: String) -> Self
 
+    /// Add a leeway amount for ID Token validation.
+    /// This value represents the clock skew for the validation of date claims e.g. `exp`.
+    ///
+    /// - Parameter leeway: number of milliseconds. Defaults to `60000`.
```
lbalmaceda:
```suggestion
    /// - Parameter leeway: number of milliseconds. Defaults to `60000` (1 minute).
```

Auth0Tests/OAuth2GrantSpec.swift
```diff
@@ -221,11 +215,25 @@ class OAuth2GrantSpec: QuickSpec {
...
+                waitUntil { done in
+                    pkce.credentials(from: values) {
+                        expect($0).to(beFailure())
+                        done()
+                    }
+                }
+            }
+            
+            it("should fail with an invalid id_token") {
+                pkce = PKCE(authentication: authentication, redirectURL: redirectURL, verifier: verifier, challenge: challenge, method: method, responseType: response, nonce: nil, leeway: leeway)
```
lbalmaceda: pass the default `nonce` value you defined on the top of the file

Auth0Tests/OAuth2GrantSpec.swift
```diff
@@ -221,11 +215,25 @@ class OAuth2GrantSpec: QuickSpec {
...
                 let token = UUID().uuidString
                 let code = UUID().uuidString
-                let values = ["code": code, "id_token" : idToken]
+                let values = ["code": code, "id_token": generateJWT(iss: nil).string, "nonce": nonce]
```
lbalmaceda: pass the default `nonce` value you defined on the top of the file to the generator

Auth0/SafariWebAuth.swift
Widcket: https://github.com/auth0/Auth0.swift/blob/04bc1a39875afaa0f724db5bcc502335c6d1896f/Auth0/OAuth2Grant.swift#L72
Looks like it uses fragment response.

Auth0/OAuth2Grant.swift
Widcket: Will be fixed on a separate PR. Thanks for raising this.

Auth0Tests/OAuth2GrantSpec.swift
Widcket: Done in ca28451be12118c8928745aead45feaa63e669e1.

Auth0Tests/OAuth2GrantSpec.swift
Widcket: Done in ca28451be12118c8928745aead45feaa63e669e1.

Auth0/WebAuth.swift
Widcket: Fixed in ca28451be12118c8928745aead45feaa63e669e1.

Auth0/OAuth2Grant.swift
Widcket: Added in ca28451be12118c8928745aead45feaa63e669e1.

Auth0/OAuth2Grant.swift
Widcket: Changed in ca28451be12118c8928745aead45feaa63e669e1.

Auth0/OAuth2Grant.swift
Widcket: Changed in ca28451be12118c8928745aead45feaa63e669e1.

Auth0/OAuth2Grant.swift
Widcket: Changed in ca28451be12118c8928745aead45feaa63e669e1.

Auth0/OAuth2Grant.swift
Widcket: Fixed in ca28451be12118c8928745aead45feaa63e669e1.

Auth0Tests/OAuth2GrantSpec.swift
Widcket: Fixed in ca28451be12118c8928745aead45feaa63e669e1.

Auth0/SafariWebAuth.swift
Widcket: Removed in ca28451be12118c8928745aead45feaa63e669e1.

Auth0/OAuth2Grant.swift
Widcket: Changed in ca28451be12118c8928745aead45feaa63e669e1.

Widcket:
@joshcanhelp can you please also check the domain logic in this entire file? 
It contains the implementations of OAuth2 grants (implicit and code flow).
https://github.com/auth0/Auth0.swift/blob/ca28451be12118c8928745aead45feaa63e669e1/Auth0/OAuth2Grant.swift

Auth0/OAuth2Grant.swift
joshcanhelp: Didn't see that, TY

Auth0/SafariWebAuth.swift
```diff
@@ -196,13 +212,21 @@ class SafariWebAuth: WebAuth {
...
+                        responseType: self.responseType,
+                        leeway: self.leeway,
+                        maxAge: self.maxAge,
+                        nonce: self.nonce)
         }
+        return ImplicitGrant(authentication: authentication,
+                             responseType: self.responseType,
+                             leeway: self.leeway,
+                             nonce: self.nonce,
+                             maxAge: self.maxAge)
```
lbalmaceda: I guess these are named params and that's why you can call them in any order, right? Would be good anyway to keep them on the same calling order everywhere. e.g. `leeway`, `nonce`, `maxAge`. I saw something similar on the `OAuth2Grant` constructor and later on that same file when you instantiate `IDTokenValidatorContext`. 
Not a blocker, but a nice to have!

Widcket: These are named params, but they can't be called in any order. Only the ones with default values can be omitted. I made the order consistent in all cases in 721afba4d18a9453de055348a5e580602d885fb3. Thanks for raising it.

PR #345, opened 2020-01-28
Execute the ID Token validation on a worker thread

Widcket:
### Changes

This update improves the SDK support for **OpenID Connect (OIDC)**. In particular, it runs the validation logic in a worker thread.

#### What’s being changed in this PR

- The validation logic now runs on a global queue with `.userInitiated` [priority](https://developer.apple.com/documentation/dispatch/dispatchqos/1780759-userinitiated), the highest priority after `.userInteractive` (which runs on the main thread). The result callback is still called on the main thread.

#### Public surface area

- **Additions:** none.
- **Changes:** none.

### Testing

### Checklist

### DIFF
--- a/Auth0/IDTokenValidator.swift
+++ b/Auth0/IDTokenValidator.swift
@@ -45,9 +45,14 @@ struct IDTokenValidator: JWTAsyncValidator {
     }
 
     func validate(_ jwt: JWT, callback: @escaping (LocalizedError?) -> Void) {
-        signatureValidator.validate(jwt) { error in
-            if let error = error { return callback(error) }
-            callback(self.claimsValidator.validate(jwt))
+        DispatchQueue.global(qos: .userInitiated).async {
+            self.signatureValidator.validate(jwt) { error in
+                if let error = error { return callback(error) }
+                let result = self.claimsValidator.validate(jwt)
+                DispatchQueue.main.async {
+                    callback(result)
+                }
+            }
         }
     }
 }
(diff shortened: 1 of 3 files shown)


Auth0/IDTokenValidator.swift
```diff
@@ -45,9 +45,14 @@ struct IDTokenValidator: JWTAsyncValidator {
     }
 
     func validate(_ jwt: JWT, callback: @escaping (LocalizedError?) -> Void) {
-        signatureValidator.validate(jwt) { error in
-            if let error = error { return callback(error) }
-            callback(self.claimsValidator.validate(jwt))
+        DispatchQueue.global(qos: .userInitiated).async {
+            self.signatureValidator.validate(jwt) { error in
```
cocojoe: Any potential issues around retain cycles?

Widcket: We actually want to keep a strong reference to `self`, since we later need it to execute the claims validator. Using `[weak self]` would not increase the ref count and therefore risks deallocation before the claims validator gets called (specially since there's a network request involved for signature validation, that could really delay that moment), because the validator itself is not referenced anywhere. Using `[unowned self]` would pose the same problem plus a runtime crash. 

AFAIK a strong reference to `self` here would not cause a strong reference cycle because the result callback is merely passed around before ultimately being executed, and is not referenced anywhere.

Widcket: @hzalaz thoughts?

Widcket: And since the validator is a struct, it's actually a value type and not a reference type. It will be copied.

PR #351, opened 2020-02-18
Added redirect URL method to WebAuth builder [SDK-1299]

Widcket:
### Changes

- Added `redirectURL()` method to `WebAuth` to allow the user to specify a custom redirect url.
- Removed some compilation conditions that targeted versions of Swift < 3.2 (this library [currently supports](https://github.com/auth0/Auth0.swift/blob/master/Auth0.podspec#L105) Swift 4+).
- Added proper target membership to two Objective-C test files that didn't belong to any target.

### Testing

### Checklist

### DIFF
--- a/Auth0/SafariWebAuth.swift
+++ b/Auth0/SafariWebAuth.swift
@@ -43,6 +43,17 @@ class SafariWebAuth: WebAuth {
     var nonce: String?
     var leeway: Int = 60 * 1000 // Default leeway is 60 seconds
     var maxAge: Int?
+
+    lazy var redirectURL: URL? = {
+        let bundleIdentifier = Bundle.main.bundleIdentifier ?? SafariWebAuth.NoBundleIdentifier
+        var components = URLComponents(url: self.url, resolvingAgainstBaseURL: true)
+        components?.scheme = self.universalLink ? "https" : bundleIdentifier
+        return components?.url?
+            .appendingPathComponent("ios")
+            .appendingPathComponent(bundleIdentifier)
+            .appendingPathComponent("callback")
+    }()
+
     private var authenticationSession = true
     private var safariPresentationStyle = UIModalPresentationStyle.fullScreen
 
@@ -93,6 +104,11 @@ class SafariWebAuth: WebAuth {
         return self
     }
 
+    func redirectURL(_ redirectURL: URL) -> Self {
+        self.redirectURL = redirectURL
+        return self
+    }
+
     func nonce(_ nonce: String) -> Self {
         self.nonce = nonce
         return self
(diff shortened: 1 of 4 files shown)


Auth0Tests/WebAuthSpec.swift
```diff
@@ -267,6 +267,10 @@ class WebAuthSpec: QuickSpec {
                 expect(newWebAuth().useUniversalLink().redirectURL?.absoluteString) == "https://\(Domain)/ios/\(bundleId)/callback"
             }
 
+            it("should build with a custom url") {
+                expect(newWebAuth().redirectURL(RedirectURL).redirectURL) == RedirectURL
```
lbalmaceda: where is this value defined, should we hardcode here something different instead?

Auth0/SafariWebAuth.swift
```diff
@@ -229,19 +234,8 @@ class SafariWebAuth: WebAuth {
...
-        components?.scheme = self.universalLink ? "https" : bundleIdentifier
-        return components?.url?
-            .appendingPathComponent("ios")
-            .appendingPathComponent(bundleIdentifier)
-            .appendingPathComponent("callback")
-    }
-
     func clearSession(federated: Bool, callback: @escaping (Bool) -> Void) {
         let logoutURL = federated ? URL(string: "/v2/logout?federated", relativeTo: self.url)! : URL(string: "/v2/logout", relativeTo: self.url)!
-        #if swift(>=3.2)
```
lbalmaceda: I assume these version checks were here for a reason, and now the "else" cases for swift versions lower than 3.2 are gone on both login and logout. That looks like breaking change to me. What is the reason to remove them?

Auth0/SafariWebAuth.swift
```diff
@@ -43,6 +43,17 @@ class SafariWebAuth: WebAuth {
     var nonce: String?
     var leeway: Int = 60 * 1000 // Default leeway is 60 seconds
     var maxAge: Int?
+
+    lazy var redirectURL: URL? = {
```
lbalmaceda: lazy means it will only get populated when required, right? e.g., if the user has set up a custom redirectURL value before launching the authentication, this lambda won't ever get called. Am I right?

Auth0Tests/WebAuthSpec.swift
Widcket: That value is defined at the top of that file, and is used through the tests in said file.

Auth0/SafariWebAuth.swift
Widcket: We're no longer supporting Swift 3.x.

Auth0/SafariWebAuth.swift
Widcket: Exactly. From the [Swift docs](https://docs.swift.org/swift-book/LanguageGuide/Properties.html):

> A lazy stored property is a property whose initial value is not calculated until the first time it is used. You indicate a lazy stored property by writing the lazy modifier before its declaration.

PR #358, opened 2020-02-29
Added Facebook token exchange method [SDK-1401]

Widcket:
#### Added
- Added a public method `tokenExchange(withFacebookSessionAccessToken:profile:scope:audience:)` to exchange Facebook's [session info access token](https://developers.facebook.com/docs/facebook-login/access-tokens/session-info-access-token/) and profile data for Auth0 credentials.
- Added a default implementation for that method in the `Authentication` protocol, that sets default values for the optional parameters.
- Added unit tests for the new method.

#### Changed
- Moved `Auth0Authentication` private methods into a private extension to avoid a linter warning due to the length of the struct body.

#### Fixed
- Explicitly passing `nil` as the scope value is now respected in the Apple token exchange method. This is backwards-compatible, since `nil` was the previous default value in that method.
- Fixed a linter whitespace warning.

### Testing

### Checklist

### DIFF
--- a/Auth0/Authentication.swift
+++ b/Auth0/Authentication.swift
@@ -568,18 +568,52 @@ public protocol Authentication: Trackable, Loggable {
     ```
     Auth0
        .authentication(clientId: clientId, domain: "samples.auth0.com")
-       .tokenExchange(withAppleAuthorizationCode: authCode, scope: "openid profile offline_access", audience: "https://myapi.com/api, fullName: credentials.fullName)
+       .tokenExchange(withAppleAuthorizationCode: authCode,
+           scope: "openid profile email",
+           audience: "https://myapi.com/api",
+           fullName: credentials.fullName)
        .start { print($0) }
     ```
 
     - parameter authCode: Authorization Code retrieved from Apple Authorization
-    - parameter scope: requested scope value when authenticating the user. By default is 'openid profile offline_access'
+    - parameter scope: Requested scope value when authenticating the user. By default is `openid profile offline_access`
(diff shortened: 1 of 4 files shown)


Auth0/Authentication.swift
```diff
@@ -1133,18 +1167,55 @@ public extension Authentication {
...
     - parameter authCode: Authorization Code retrieved from Apple Authorization
-    - parameter scope: requested scope value when authenticating the user. By default is 'openid profile offline_access'
+    - parameter scope: Requested scope value when authenticating the user. By default is `openid profile offline_access`
     - parameter audience: API Identifier that the client is requesting access to
     - parameter fullName: The full name property returned with the Apple ID Credentials
 
     - returns: a request that will yield Auth0 user's credentials
     */
-    func tokenExchange(withAppleAuthorizationCode authCode: String, scope: String? = nil, audience: String? = nil, fullName: PersonNameComponents? = nil) -> Request<Credentials, AuthenticationError> {
+    func tokenExchange(withAppleAuthorizationCode authCode: String, scope: String? = "openid profile offline_access", audience: String? = nil, fullName: PersonNameComponents? = nil) -> Request<Credentials, AuthenticationError> {
```
lbalmaceda: Should we move this to a constant?

Widcket: Currently none of the default param values in the Authentication API client have been extracted to constants. However, for the next major I plan on reviewing and consolidating those default values, and maybe extracting them. But not right now.

PR #367, opened 2020-03-25
Decouple SimpleKeychain as being the required keychain wrapper

sam-w:
### Changes
`CredentialsManager` currently prescribes an instance of [`A0SimpleKeychain`](https://github.com/auth0/SimpleKeychain) as the wrapper around the iOS keychain. This isn't great for situations where `SimpleKeychain` is not usable, e.g. [the current issue with iOS 13.4 support](https://github.com/auth0/SimpleKeychain/issues/90).

It would be better if we de-coupled `Auth0.swift` from `SimpleKeychain` and allowed people to use an alternative if they need to.

It would also probably be benificial for regular Auth0.swift contributors when writing unit tests, as a mock keychain dependency could be injected rather than directly using the test simulator's keychain.

### Implementation

I've added a new `KeychainWrapper` protocol, which replaces `A0SimpleKeychain` as the item referenced in `CredentialsManager.storage`. To maintain ease-of-use and backwards compatibility, `A0SimpleKeychain` is still used as the default `storage`.

### References

- https://github.com/auth0/SimpleKeychain/issues/90
- https://stackoverflow.com/questions/56700680/keychain-query-always-returns-errsecitemnotfound-after-upgrading-to-ios-13
- https://forums.developer.apple.com/thread/120619

### Testing

- Functionality remains identical if code remains as-is, and hence unit tests run as per normal.
- To test with an alternative `KeychainWrapper`, adopt the protocol on your keychain manager of choice and use `CredentialsManager(..., storage: yourKeychainManager)`.
- I have not written unit tests for alternative keychain managers, as these should be unit tested in their own projects.

* [x] This change has been tested on the latest version of the platform/language.

### Checklist

### DIFF
--- a/Auth0/CredentialsManager.swift
+++ b/Auth0/CredentialsManager.swift
@@ -27,10 +27,19 @@ import JWTDecode
 import LocalAuthentication
 #endif
 
+public protocol KeychainWrapper {
+
+    func setData(_ data: Data, forKey: String) -> Bool
+    func data(forKey: String) -> Data?
+    func deleteEntry(forKey: String) -> Bool
+}
+
+extension A0SimpleKeychain: KeychainWrapper {}
+
 /// Credentials management utility
 public struct CredentialsManager {
 
-    private let storage: A0SimpleKeychain
+    private let storage: KeychainWrapper
     private let storeKey: String
     private let authentication: Authentication
     #if os(iOS)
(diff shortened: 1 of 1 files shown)


Widcket:
Hi @sam-w, thanks for your contribution. However, we've already released a workaround in [SimpleKeychain 0.11.1](https://github.com/auth0/SimpleKeychain/releases/tag/0.11.1), so this change will not be needed.

sam-w:
@Widcket @cocojoe I respectfully disagree with this PR being closed. Just because you've fixed one issue, doesn't mean we're not going to see another one in future. In fact, it should act as _more_ motivation to allow your customers to workaround future issues, I would have thought? (Plus y'all should be decoupling your dependencies as good developer practice anyway right? :P).

cocojoe:
@sam-w thanks for reporting this originally, to be clear the original keychain issue appears to be an Apple bug of some kind, also this was limited to the simulator and thankfully not production. This type of issue is frustrating and I'm sure at some point it will be fixed upstream. 

However, your point is noted around a Storage protocol and I'll capture this as feature suggestion for our next major.

Thanks

Widcket:
Hi @sam-w, Auth0.swift [2.0.0-beta.0](https://github.com/auth0/Auth0.swift/releases/tag/2.0.0-beta.0) is out now. It decouples SimpleKeychain from the Credentials Manager, allowing you to use a custom storage layer. See https://github.com/auth0/Auth0.swift/issues/637 for more information.

Widcket:
The [2.0.0](https://github.com/auth0/Auth0.swift/releases/tag/2.0.0) release is out now. Please check the [Migration Guide](https://github.com/auth0/Auth0.swift/blob/master/V2_MIGRATION_GUIDE.md) for more information on how to update your apps.

PR #369, opened 2020-03-26
Added support for Refresh Token Rotation [SDK-1420]

Widcket:
### Changes

This PR adds support for Refresh Token Rotation (RTR). When RTR is enabled, a new refresh token will be included in the response of a renew credentials operation. That refresh token is then added to the returned `Credentials` object instead of the old refresh token.

### Testing

### Checklist

### DIFF
--- a/Auth0/CredentialsManager.swift
+++ b/Auth0/CredentialsManager.swift
@@ -183,7 +183,7 @@ public struct CredentialsManager {
                 let newCredentials = Credentials(accessToken: credentials.accessToken,
                                                  tokenType: credentials.tokenType,
                                                  idToken: credentials.idToken,
-                                                 refreshToken: refreshToken,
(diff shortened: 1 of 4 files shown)


Auth0/CredentialsManager.swift
```diff
@@ -183,7 +183,7 @@ public struct CredentialsManager {
                 let newCredentials = Credentials(accessToken: credentials.accessToken,
                                                  tokenType: credentials.tokenType,
                                                  idToken: credentials.idToken,
-                                                 refreshToken: refreshToken,
+                                                 refreshToken: credentials.refreshToken ?? refreshToken,
```
lbalmaceda: would this act as an OR? i.e. if `credentials.refreshToken` is defined, use it. If not, use `refreshToken`

Widcket: Exactly, if `credentials.refreshToken` is not nil will use it.

PR #388, opened 2020-05-08
Added support for ephemeral sessions [SDK-1559]

Widcket:
### Changes

- Added a new public method `useEphemeralSession()` to WebAuth that sets `prefersEphemeralWebBrowserSession` to `true` on iOS 13+ and macOS. It has no effect on older versions of iOS. As it depends on `prefersEphemeralWebBrowserSession`, this method is only available for Swift 5.1+. **Usage of this method disables SSO.**
- Renamed the file `WebAuth.swift` to `WebAuthenticatable.swift`.
- Renamed a couple of parameters in internal methods.

### Testing

This change was tested manually by performing login and logout on macOS **10.15.4** and iOS simulators **10.3.1**, **11.4** and **13.3**.

### Checklist

### DIFF
--- a/Auth0.podspec
+++ b/Auth0.podspec
@@ -32,7 +32,7 @@ web_auth_files = [
   'Auth0/SessionCallbackTransaction.swift',
   'Auth0/SessionTransaction.swift',
   'Auth0/TransactionStore.swift',
-  'Auth0/WebAuth.swift',
+  'Auth0/WebAuthenticatable.swift',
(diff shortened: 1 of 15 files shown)


Widcket:
> Seems like a lot of refactoring going on for what is basically surfacing this `ephemeralSession` flag.
I always try to leave the codebase in a better shape than before. It's better to do that incrementally like this.

PR #390, opened 2020-05-19
Move NativeAuth.start() into the NativeAuth protocol declaration

ejensen:
### Changes

Moves `NativeAuth.start()` into the `NativeAuth` protocol declaration. This allows SDK clients to provide an implementation. Rather than restricting it to the default implementation defined in the protocol extension.

This enables clients to write their own implementation of `start`. For example, we need to call `authentication.login(facebookSessionAccessToken: ...)` in our implementation of the protocol, rather than `authentication.loginSocial(token:...)` that's called by the default protocol extension implementation.

This is a non-breaking change to the protocol.

### References

- Methods defined in a protocol and provided a default implementation are dynamically dispatched (can be overridden)
- Methods only defined in a protocol extension are always statically dispatched (can't be overridden)
- [Swift Evolution Board discussion on static vs dynamic dispatch of protocol extensions](https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20151207/001125.html)

### Testing

No additional testing is required since this is an inert change to a protocol

### Checklist

### DIFF
--- a/Auth0/NativeAuth.swift
+++ b/Auth0/NativeAuth.swift
@@ -78,6 +78,17 @@ public protocol NativeAuthTransaction: AuthTransaction {
      - parameter callback: callback with the IdP credentials on success or the cause of the error.
      */
     func auth(callback: @escaping Callback)
+
+    /**
+     Starts the Auth transaction by trying to authenticate the user with the IdP SDK first,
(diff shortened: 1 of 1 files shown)


Widcket:
@ejensen thanks for the contribution!

PR #395, opened 2020-06-24
Added Forced Renew of Credentials Manager

Tauseef-TL:
### Changes

This change adds the ability to force refresh the `CredentialsManager` 

**Classes Changes**
`CredentialsManager`

**Method Changes**
`public func renew(withScope: String? = nil, callback: @escaping (CredentialsManagerError?, Credentials?) -> Void)` [Added]

### Testing

This change has been tested manually and is currently in the production app.

### Checklist

### DIFF
--- a/Auth0/CredentialsManager.swift
+++ b/Auth0/CredentialsManager.swift
@@ -167,32 +167,39 @@ public struct CredentialsManager {
         self.retrieveCredentials(withScope: scope, callback: callback)
     }
     #endif
-    
-    /// Renew credentials regardless of expiry time. This will yield new credentials using the refreshToken regardless of whether the accessToken has expired.
+
+    /// Renew credentials regardless of expiry time. This will yield new credentials regardless of whether the accessToken has expired.
     /// Renewed credentials will be stored in the keychain.
     ///
     ///
     /// ```
-    /// credentialsManager.renew(refreshToken: "some_refresh_token") { error, credential in
+    /// credentialsManager.renew { error, credential in
     ///    guard error == nil else { return }
     ///    print(credential)
     /// }
     /// ```
     ///
     /// - Parameters:
-    ///   - refreshToken: refreshToken to request new credentials.
     ///   - scope: scopes to request for the new tokens. By default is nil which will ask for the same ones requested during original Auth
     ///   - callback: callback with the user's credentials or the cause of the error.
     /// - Important: This method only works for a refresh token obtained after auth with OAuth 2.0 API Authorization.
     /// - Note: [Auth0 Refresh Tokens Docs](https://auth0.com/docs/tokens/refresh-token)
-    public func renew(refreshToken: String, scope: String? = nil, callback: @escaping (CredentialsManagerError?, Credentials?) -> Void) {
+    public func renew(withScope scope: String? = nil, callback: @escaping (CredentialsManagerError?, Credentials?) -> Void) {
+        guard let data = self.storage.data(forKey: self.storeKey),
+            let credentials = NSKeyedUnarchiver.unarchiveObject(with: data) as? Credentials else {
+                return callback(.noCredentials, nil)
+        }
+        guard let refreshToken = credentials.refreshToken else {
+            return callback(.noRefreshToken, nil)
+        }
+
         self.authentication.renew(withRefreshToken: refreshToken, scope: scope).start {
             switch $0 {
             case .success(let credentials):
                 let newCredentials = Credentials(accessToken: credentials.accessToken,
                                                  tokenType: credentials.tokenType,
                                                  idToken: credentials.idToken,
-                                                 refreshToken: refreshToken,
+                                                 refreshToken: credentials.refreshToken ?? refreshToken,
                                                  expiresIn: credentials.expiresIn,
                                                  scope: credentials.scope)
                 _ = self.store(credentials: newCredentials)
--- a/Auth0/CredentialsManager.swift
+++ b/Auth0/CredentialsManager.swift
@@ -167,6 +167,41 @@ public struct CredentialsManager {
         self.retrieveCredentials(withScope: scope, callback: callback)
     }
     #endif
+    
+    /// Renew credentials regardless of expiry time. This will yield new credentials using the refreshToken regardless of whether the accessToken has expired.
+    /// Renewed credentials will be stored in the keychain.
+    ///
+    ///
+    /// ```
+    /// credentialsManager.renew(refreshToken: "some_refresh_token") { error, credential in
+    ///    guard error == nil else { return }
+    ///    print(credential)
+    /// }
+    /// ```
+    ///
+    /// - Parameters:
+    ///   - refreshToken: refreshToken to request new credentials.
+    ///   - scope: scopes to request for the new tokens. By default is nil which will ask for the same ones requested during original Auth
+    ///   - callback: callback with the user's credentials or the cause of the error.
+    /// - Important: This method only works for a refresh token obtained after auth with OAuth 2.0 API Authorization.
+    /// - Note: [Auth0 Refresh Tokens Docs](https://auth0.com/docs/tokens/refresh-token)
+    public func renew(refreshToken: String, scope: String? = nil, callback: @escaping (CredentialsManagerError?, Credentials?) -> Void) {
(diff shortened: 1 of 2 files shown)


danielphillips:
@cocojoe one of our changes as mentioned before.

cocojoe:
Hey @danielphillips, I think for the solution it's worth thinking exploring the case here, to ensure this is the right way to do it from an architectural perspective.  

You may need to refresh my memory here, I believe you know you are going to be passing the AT into an environment that means you can't call a refresh on the device, so you need it to be valid for at least X minutes?

danielphillips:
> Hey @danielphillips, I think for the solution it's worth thinking exploring the case here, to ensure this is the right way to do it from an architectural perspective.
> 
> You may need to refresh my memory here, I believe you know you were talking about passing the AT into an environment that means you can't call a refresh on the device, so it needs to be valid for at least X minutes?

That's right, we have some feature on our website which requires authentication, we'd like to avoid users opening the web view and needing to login again. We'd also like to avoid the user being logged out mid way through whatever it is they're doing on the website, in the case where the access token is near expiring.

So the idea here is, what if we could get a brand new, fresh access token when we need to open a web page, then we'd be sure that the user has reasonable time to complete the task at hand without being punted over to a login screen.

cocojoe:
@danielphillips How are you set up that you need to use the WebAuthentication again? The Refresh Token rotation should be invisible as it's an API call and you don't need to prompt for the WebAuthentication.

danielphillips:
I think you're asking why don't we execute something on the website to trigger a refresh? Or have the website reach back to the app to do the token refresh?

We are only sending the access token into the website, not the refresh token. The desire architecturally was not to deal with complexities of refreshing the token on the website, so once the token expires you are taken to the web login page. 

Is that what you was asking?

cocojoe:
Yes, I'm wondering if a leeway would be more appropriate, my concern is once you introduce forced refresh behaviour, developers may incorrectly use this as the default path and run into rate limit issues etc

Is there a leeway that would work here? e.g. Refresh 1 hour before expiry?

danielphillips:
Our expiry is set to 10 minutes for app tokens. There was no desire technically to build a two way communication to allow refreshes.

cocojoe:
OK let me discuss internally and see what's best here.

cocojoe:
@danielphillips just to clarify when I talked about a leeway, you could pass in say 600 seconds, so a 10-minute leeway so if the access token was going to expire in 10 or less minutes when you ask for credentials, it would auto-force the renewal so you are confident it will be valid for the time you need. 

Would that work?

danielphillips:
What API do I pass a leeway to? We use the CredentialsManager is it on that?

cocojoe:
I explored a bit more in detail with the team, I think there is another use case here that helps justify a straight force method. If you want to down scope an AT then it would need to be refreshed to produce the correct AT.

For example, you may want to reduce the scope for a restricted area.

In that case that does make sense. However, the implementation needs a bit of work so I'll review this PR and feedback there.

Auth0/CredentialsManager.swift
```diff
@@ -167,6 +167,41 @@ public struct CredentialsManager {
...
+    ///
+    /// ```
+    /// credentialsManager.renew(refreshToken: "some_refresh_token") { error, credential in
+    ///    guard error == nil else { return }
+    ///    print(credential)
+    /// }
+    /// ```
+    ///
+    /// - Parameters:
+    ///   - refreshToken: refreshToken to request new credentials.
```
cocojoe, line 183: What is the purpose of passing in a refresh token? If you're going outside of the credentials manager I would suggest simply calling renew directly.

https://github.com/auth0/Auth0.swift/blob/master/Auth0/Auth0Authentication.swift#L338

Auth0/CredentialsManager.swift
```diff
@@ -167,6 +167,41 @@ public struct CredentialsManager {
...
+    ///   - scope: scopes to request for the new tokens. By default is nil which will ask for the same ones requested during original Auth
+    ///   - callback: callback with the user's credentials or the cause of the error.
+    /// - Important: This method only works for a refresh token obtained after auth with OAuth 2.0 API Authorization.
+    /// - Note: [Auth0 Refresh Tokens Docs](https://auth0.com/docs/tokens/refresh-token)
+    public func renew(refreshToken: String, scope: String? = nil, callback: @escaping (CredentialsManagerError?, Credentials?) -> Void) {
```
cocojoe, line 188: I don't think refreshToken should be here, the signature should then start `withScope scope: String? = nil` which is inline with the previous method.

Auth0/CredentialsManager.swift
```diff
@@ -167,6 +167,41 @@ public struct CredentialsManager {
...
+        self.authentication.renew(withRefreshToken: refreshToken, scope: scope).start {
+            switch $0 {
+            case .success(let credentials):
+                let newCredentials = Credentials(accessToken: credentials.accessToken,
+                                                 tokenType: credentials.tokenType,
+                                                 idToken: credentials.idToken,
+                                                 refreshToken: refreshToken,
+                                                 expiresIn: credentials.expiresIn,
+                                                 scope: credentials.scope)
+                _ = self.store(credentials: newCredentials)
```
cocojoe, line 198: This should refactor to just storing the credentials object as no manipulation should be required.

cocojoe: Actually it would be better to match the `retrieveCredentials` method.  If you're using Refresh Token Rotation you will get issued a new RT each time. However, if you're using non rotating then you want to fallback to the existing refreshToken e.g.

```swift
let newCredentials = Credentials(accessToken: credentials.accessToken,
                                                 tokenType: credentials.tokenType,
                                                 idToken: credentials.idToken,
                                                 refreshToken: credentials.refreshToken ?? refreshToken,
                                                 expiresIn: credentials.expiresIn,
                                                 scope: credentials.scope)
                _ = self.store(credentials: newCredentials)
```

Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -454,6 +454,46 @@ class CredentialsManagerSpec: QuickSpec {
                         }
                     }
                 }
+                
+                it("should yield new credentials when credentials expire in the future") {
```
cocojoe, line 458: The tests should ideally be testing for 1 case, in this case it doesn't really matter if the credentials have expired or not, as you are manually forcing a refresh. The signal here should be does the renew method on its own return new credentials.

Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -454,6 +454,46 @@ class CredentialsManagerSpec: QuickSpec {
...
+                                expect(storedCredentials?.accessToken) == newCredentials?.accessToken
+                                expect(storedCredentials?.refreshToken) == newCredentials?.refreshToken
+                                expect(storedCredentials?.idToken) == newCredentials?.idToken
+                                done()
+                            }
+                        }
+                    }
+                }
+                
+                it("should yield new credentials when credentials have expired") {
```
cocojoe, line 478: Similar to first, it doesn't matter if it's expired or not as the credential manager is not doing this automagically, you are forcing the refresh.

cocojoe:
@lbalmaceda would you mind taking a look as well please as we will then want to add to Android as well and it's what we talked about on Slack.  There other part around scopes is something we can address ourselves in another PR. The focus here is on the manual renewal.

Widcket:
Hi @Tauseef-TL and @danielphillips, apologies for the delay on this. Aside from the use case at hand —renewing the AT when it won’t remain valid for long— we’ve identified an additional use case requiring a forced AT renewal on demand. When a new AT is requested with a reduced scope than originally granted, it should be renewed immediately even if the old AT hasn’t expired yet. But in the current CredentialsManager implementation, it won’t actually be renewed unless the old AT has expired.

After careful consideration, we’ve decided not to move forward with this PR and implement an alternative solution that covers both use cases. We’ll be adding an additional parameter `minTTL` to `credentials(withScope:callback:)` to force renew the AT unless it remains valid for a certain amount of time, and we’ll be using the existing `scope` parameter to do the same when requesting a downscoped AT. 

Does this solution cover your use case?

Widcket:
The implementation is now available on the branch `feature/forced-renewal`.

danielphillips:
Thank you we will take a look at the other PR now.

PR #398, opened 2020-07-14
[DO NOT USE] Experimental Changes for Xcode 12 Beta

cocojoe:
Provide an alternative path for Signature validation in MacOS >= 10.12
Some Fixes for Tests related to Quick/Nimble changes.
Private Dependency Bumps

Using a forked version of Quick/Nimble that I patched for now. I'm sure these will break Xcode 11 CI Tests 😄 

There is a bug in Xcode 12 as you need to remove i386 & armv7* architectures, this effected Nimble at least when using it as a dependency, these changes could break Xcode 11.


cocojoe:
@Widcket Would be worth testing this branch in Lock.swift to see what happens.

cocojoe:
Not relevant now

PR #399, opened 2020-07-21
Added TTL renewal of credentials in CredentialsManager [SDK-1818]

Widcket:
### Changes

- Added a new `minTTL` parameter to the `credentials(withScope:callback:)` method to force renew the Access Token unless it remains valid for a certain amount of time, and to ensure the minimum lifetime of the renewed Access Token. This guarantees that the retrieved Access Token will always remain valid for at least `minTTL`.

```swift
credentialsManager.credentials(minTTL: 600) { error, newCredentials in // 600 seconds
    // newCredentials.accessToken will be valid for at least 10 minutes
}
```

- Added a new `minTTL` parameter to the `hasValid()` method to check that the Access Token will remain valid for at least a certain amount of time.

```swift
if credentialsManager.hasValid(minTTL: 600) {
    // The Access Token will remain valid for at least 10 minutes
}
```

- Used the existing `scope` parameter to force renew the Access Token when requesting a reduced scope than originally granted.

```swift
credentialsManager.credentials(withScope: "openid profile offline_access") { error, newCredentials in
    // newCredentials.accessToken will be limited to the new scope
}
```

### References

Replaces #395 

### Testing

### Checklist

### DIFF
--- a/Auth0/CredentialsManager.swift
+++ b/Auth0/CredentialsManager.swift
@@ -119,7 +119,8 @@ public struct CredentialsManager {
 
     /// Checks if a non-expired set of credentials are stored
     ///
-    /// - Returns: if there are valid and non-expired credentials stored
+    /// - Parameter minTTL: minimum lifetime in seconds the access token must have left.
+    /// - Returns: if there are valid and non-expired credentials stored.
     public func hasValid(minTTL: Int = 0) -> Bool {
         guard let data = self.storage.data(forKey: self.storeKey),
             let credentials = NSKeyedUnarchiver.unarchiveObject(with: data) as? Credentials,
--- a/Auth0/CredentialsManager.swift
+++ b/Auth0/CredentialsManager.swift
@@ -120,13 +120,11 @@ public struct CredentialsManager {
     /// Checks if a non-expired set of credentials are stored
     ///
     /// - Returns: if there are valid and non-expired credentials stored
-    public func hasValid() -> Bool {
-        guard
-            let data = self.storage.data(forKey: self.storeKey),
+    public func hasValid(minTTL: Int = 0) -> Bool {
+        guard let data = self.storage.data(forKey: self.storeKey),
             let credentials = NSKeyedUnarchiver.unarchiveObject(with: data) as? Credentials,
-            credentials.accessToken != nil
-            else { return false }
-        return !self.hasExpired(credentials) || credentials.refreshToken != nil
+            credentials.accessToken != nil else { return false }
+        return (!self.hasExpired(credentials) && !self.willExpire(credentials, within: minTTL)) || credentials.refreshToken != nil
     }
 
     /// Retrieve credentials from keychain and yield new credentials using `refreshToken` if `accessToken` has expired
@@ -171,7 +169,7 @@ public struct CredentialsManager {
     private func retrieveCredentials(withScope scope: String?, minTTL: Int, callback: @escaping (CredentialsManagerError?, Credentials?) -> Void) {
         guard let data = self.storage.data(forKey: self.storeKey),
             let credentials = NSKeyedUnarchiver.unarchiveObject(with: data) as? Credentials else { return callback(.noCredentials, nil) }
-        guard credentials.expiresIn != nil else { return callback(.noCredentials, nil) }
+        guard let expiresIn = credentials.expiresIn else { return callback(.noCredentials, nil) }
         guard self.willExpire(credentials, within: minTTL) ||
             self.hasExpired(credentials) ||
             self.hasScopeChanged(credentials, than: scope) else { return callback(nil, credentials) }
@@ -187,8 +185,11 @@ public struct CredentialsManager {
                                                  expiresIn: credentials.expiresIn,
                                                  scope: credentials.scope)
                 if self.willExpire(newCredentials, within: minTTL) {
+                    let accessTokenLifetime = Int(expiresIn.timeIntervalSinceNow / 1000)
                     // TODO: On the next major add a new case to CredentialsManagerError
-                    let error = NSError(domain: "The lifetime of the renewed Access Token is less than minTTL", code: -99999, userInfo: nil)
+                    let error = NSError(domain: "The lifetime of the renewed Access Token (\(accessTokenLifetime)s) is less than minTTL requested (\(minTTL)s). Increase the 'Token Expiration' setting of your Auth0 API in the dashboard or request a lower minTTL",
+                        code: -99999,
+                        userInfo: nil)
                     callback(.failedRefresh(error), nil)
                 } else {
                     _ = self.store(credentials: newCredentials)
--- a/Auth0/CredentialsManager.swift
+++ b/Auth0/CredentialsManager.swift
@@ -129,10 +129,9 @@ public struct CredentialsManager {
         return !self.hasExpired(credentials) || credentials.refreshToken != nil
     }
 
-    /// Retrieve credentials from keychain and yield new credentials using refreshToken if accessToken has expired
+    /// Retrieve credentials from keychain and yield new credentials using `refreshToken` if `accessToken` has expired
     /// otherwise the retrieved credentails will be returned as they have not expired. Renewed credentials will be
     /// stored in the keychain.
-    ///
     ///
     /// ```
     /// credentialsManager.credentials {
@@ -142,12 +141,13 @@ public struct CredentialsManager {
     /// ```
     ///
     /// - Parameters:
-    ///   - scope: scopes to request for the new tokens. By default is nil which will ask for the same ones requested during original Auth
+    ///   - scope: scopes to request for the new tokens. By default is nil which will ask for the same ones requested during original Auth.
+    ///   - minTTL: minimum time in milliseconds the access token must remain valid to avoid being renewed.
     ///   - callback: callback with the user's credentials or the cause of the error.
     /// - Important: This method only works for a refresh token obtained after auth with OAuth 2.0 API Authorization.
-    /// - Note: [Auth0 Refresh Tokens Docs](https://auth0.com/docs/tokens/refresh-token)
+    /// - Note: [Auth0 Refresh Tokens Docs](https://auth0.com/docs/tokens/concepts/refresh-tokens)
     #if WEB_AUTH_PLATFORM
-    public func credentials(withScope scope: String? = nil, callback: @escaping (CredentialsManagerError?, Credentials?) -> Void) {
+    public func credentials(withScope scope: String? = nil, minTTL: Int = 0, callback: @escaping (CredentialsManagerError?, Credentials?) -> Void) {
         guard self.hasValid() else { return callback(.noCredentials, nil) }
         if #available(iOS 9.0, macOS 10.15, *), let bioAuth = self.bioAuth {
             guard bioAuth.available else { return callback(.touchFailed(LAError(LAError.touchIDNotAvailable)), nil) }
@@ -155,26 +155,26 @@ public struct CredentialsManager {
                 guard $0 == nil else {
                     return callback(.touchFailed($0!), nil)
                 }
-                self.retrieveCredentials(withScope: scope, callback: callback)
+                self.retrieveCredentials(withScope: scope, minTTL: minTTL, callback: callback)
             }
         } else {
-            self.retrieveCredentials(withScope: scope, callback: callback)
+            self.retrieveCredentials(withScope: scope, minTTL: minTTL, callback: callback)
         }
     }
     #else
-    public func credentials(withScope scope: String? = nil, callback: @escaping (CredentialsManagerError?, Credentials?) -> Void) {
+    public func credentials(withScope scope: String? = nil, minTTL: Int = 0, callback: @escaping (CredentialsManagerError?, Credentials?) -> Void) {
         guard self.hasValid() else { return callback(.noCredentials, nil) }
-        self.retrieveCredentials(withScope: scope, callback: callback)
+        self.retrieveCredentials(withScope: scope, minTTL: minTTL, callback: callback)
     }
     #endif
 
-    private func retrieveCredentials(withScope scope: String? = nil, callback: @escaping (CredentialsManagerError?, Credentials?) -> Void) {
-        guard
-            let data = self.storage.data(forKey: self.storeKey),
-            let credentials = NSKeyedUnarchiver.unarchiveObject(with: data) as? Credentials
-            else { return callback(.noCredentials, nil) }
+    private func retrieveCredentials(withScope scope: String?, minTTL: Int, callback: @escaping (CredentialsManagerError?, Credentials?) -> Void) {
+        guard let data = self.storage.data(forKey: self.storeKey),
+            let credentials = NSKeyedUnarchiver.unarchiveObject(with: data) as? Credentials else { return callback(.noCredentials, nil) }
         guard credentials.expiresIn != nil else { return callback(.noCredentials, nil) }
-        guard self.hasExpired(credentials) else { return callback(nil, credentials) }
+        guard self.willExpire(credentials, within: minTTL) ||
+            self.hasExpired(credentials) ||
+            self.hasDifferentScope(credentials, than: scope) else { return callback(nil, credentials) }
         guard let refreshToken = credentials.refreshToken else { return callback(.noRefreshToken, nil) }
 
         self.authentication.renew(withRefreshToken: refreshToken, scope: scope).start {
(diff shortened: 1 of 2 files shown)


Auth0/CredentialsManager.swift
```diff
@@ -142,39 +141,40 @@ public struct CredentialsManager {
...
-            let credentials = NSKeyedUnarchiver.unarchiveObject(with: data) as? Credentials
-            else { return callback(.noCredentials, nil) }
+    private func retrieveCredentials(withScope scope: String?, minTTL: Int, callback: @escaping (CredentialsManagerError?, Credentials?) -> Void) {
+        guard let data = self.storage.data(forKey: self.storeKey),
+            let credentials = NSKeyedUnarchiver.unarchiveObject(with: data) as? Credentials else { return callback(.noCredentials, nil) }
         guard credentials.expiresIn != nil else { return callback(.noCredentials, nil) }
-        guard self.hasExpired(credentials) else { return callback(nil, credentials) }
+        guard self.willExpire(credentials, within: minTTL) ||
+            self.hasExpired(credentials) ||
+            self.hasDifferentScope(credentials, than: scope) else { return callback(nil, credentials) }
```
cocojoe, line 177: I think `self.hasScopeChanged` is more explicit.

Auth0/CredentialsManager.swift
```diff
@@ -186,14 +186,28 @@ public struct CredentialsManager {
                                                  refreshToken: credentials.refreshToken ?? refreshToken,
                                                  expiresIn: credentials.expiresIn,
                                                  scope: credentials.scope)
-                _ = self.store(credentials: newCredentials)
-                callback(nil, newCredentials)
+                if self.willExpire(newCredentials, within: minTTL) {
+                    // TODO: On the next major add a new case to CredentialsManagerError
```
cocojoe, line 190: Why do we need a new major to add another error enumeration? It would be a BC if you removed an error as this would cause existing code compilation to fail but adding one doesn't?

Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -454,8 +475,85 @@ class CredentialsManagerSpec: QuickSpec {
...
             }
-            
+
+            context("forced renew") {
+
+                it("should not yield a new access token without a new scope") {
+                    credentials = Credentials(accessToken: AccessToken, refreshToken: RefreshToken, expiresIn: Date(timeIntervalSinceNow: ExpiresIn), scope: "openid profile")
+                    _ = credentialsManager.store(credentials: credentials)
+                    waitUntil(timeout: 2) { done in
+                        credentialsManager.credentials(withScope: nil, minTTL: 0) { error = $0; newCredentials = $1
```
cocojoe, line 487: let's also do one without specifying param values to ensure defaults are working as expected.

Auth0/CredentialsManager.swift
Widcket: Because existing switch statements in customer's code will no longer be exhaustive.

Auth0Tests/CredentialsManagerSpec.swift
Widcket: Added in https://github.com/auth0/Auth0.swift/pull/399/commits/76f4575cf672ad7792eae20d4c64361a0de4a504.

Auth0/CredentialsManager.swift
Widcket: Thanks, changed in https://github.com/auth0/Auth0.swift/pull/399/commits/76f4575cf672ad7792eae20d4c64361a0de4a504.

Auth0/CredentialsManager.swift
```diff
@@ -142,39 +141,40 @@ public struct CredentialsManager {
     /// ```
     ///
     /// - Parameters:
-    ///   - scope: scopes to request for the new tokens. By default is nil which will ask for the same ones requested during original Auth
+    ///   - scope: scopes to request for the new tokens. By default is nil which will ask for the same ones requested during original Auth.
+    ///   - minTTL: minimum time in milliseconds the access token must remain valid to avoid being renewed.
```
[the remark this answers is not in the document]
Widcket: Changed in https://github.com/auth0/Auth0.swift/pull/399/commits/76f4575cf672ad7792eae20d4c64361a0de4a504.

Auth0/CredentialsManager.swift
```diff
@@ -142,12 +142,12 @@ public struct CredentialsManager {
...
     /// - Parameters:
     ///   - scope: scopes to request for the new tokens. By default is nil which will ask for the same ones requested during original Auth.
-    ///   - minTTL: minimum time in milliseconds the access token must remain valid to avoid being renewed.
+    ///   - minTTL: minimum time in seconds the access token must remain valid to avoid being renewed.
     ///   - callback: callback with the user's credentials or the cause of the error.
     /// - Important: This method only works for a refresh token obtained after auth with OAuth 2.0 API Authorization.
     /// - Note: [Auth0 Refresh Tokens Docs](https://auth0.com/docs/tokens/concepts/refresh-tokens)
     #if WEB_AUTH_PLATFORM
-    public func credentials(withScope scope: String? = nil, minTTL: Int = 0, callback: @escaping (CredentialsManagerError?, Credentials?) -> Void) {
+    public func credentials(withScope scope: String? = nil, minTTL: Float = 0, callback: @escaping (CredentialsManagerError?, Credentials?) -> Void) {
```
cocojoe, line 150: What is the purpose of the float as we're using seconds?

Widcket: Since seconds are a lot less granular than milliseconds, floats are needed to allow fractions of seconds, instead of whole seconds only.

cocojoe: What would be the use case that would require milliseconds? Let's look at a use case from the original issue of requiring a token with 10 minutes left e.g. 600 seconds.

Widcket: You're right, I'll change it back to Int.

Widcket: Changed in https://github.com/auth0/Auth0.swift/pull/399/commits/b02ee119124e4b78d2339314ded5421517b3a987.

cocojoe:
@danielphillips @Tauseef-TL We had a think about your use case and wanted to have a consistent story across the native SDKs, starting with iOS. Would be great to get your feedback on the functionality added in this PR (snippet included) and ensure it also solves your use case.

cocojoe:
@lbalmaceda please can you review to ensure it aligns with your expectations around the discovery and as a heads up for Android. Thx

Auth0/CredentialsManager.swift
```diff
@@ -186,14 +186,28 @@ public struct CredentialsManager {
                                                  refreshToken: credentials.refreshToken ?? refreshToken,
                                                  expiresIn: credentials.expiresIn,
                                                  scope: credentials.scope)
-                _ = self.store(credentials: newCredentials)
-                callback(nil, newCredentials)
+                if self.willExpire(newCredentials, within: minTTL) {
+                    // TODO: On the next major add a new case to CredentialsManagerError
+                    let error = NSError(domain: "The lifetime of the renewed Access Token is less than minTTL", code: -99999, userInfo: nil)
```
lbalmaceda, line 191: let's not use magic numbers and ensure the message gives a hint on where the error could be resolved (dashboard)

Auth0/CredentialsManager.swift
```diff
@@ -186,14 +186,28 @@ public struct CredentialsManager {
...
+                    _ = self.store(credentials: newCredentials)
+                    callback(nil, newCredentials)
+                }
             case .failure(let error):
                 callback(.failedRefresh(error), nil)
             }
         }
     }
 
+    func willExpire(_ credentials: Credentials, within ttl: Int) -> Bool {
```
lbalmaceda, line 203: Adding a new method for this can be avoided. There's already a `hasExpired` method that behaves exactly like this but without considering the `minTtl`. Let's do that instead.

Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -287,10 +295,25 @@ class CredentialsManagerSpec: QuickSpec {
...
-            
+
         }
-        
-        describe("validity with id token") {
+
+        describe("scope") {
+
+            it("should return true when the scope has changed") {
+                let credentials = Credentials(scope: "openid email")
```
lbalmaceda, line 304: since tests are also documentation, I'd swap these scope values so we don't show that a token could have its scope increased.

Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -454,8 +477,99 @@ class CredentialsManagerSpec: QuickSpec {
...
+                        credentialsManager.credentials(withScope: "openid profile", minTTL: 0) { error, newCredentials in
+                            expect(error).to(beNil())
+                            expect(newCredentials?.accessToken) == AccessToken
+                            done()
+                        }
+                    }
+                }
+
+                it("should yield a new access token with a new scope") {
+                    credentials = Credentials(accessToken: AccessToken, refreshToken: RefreshToken, expiresIn: Date(timeIntervalSinceNow: ExpiresIn), scope: "openid profile")
```
lbalmaceda, line 522: similar to the comment above, swap the scope values.

Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -454,8 +477,99 @@ class CredentialsManagerSpec: QuickSpec {
...
+                        }
+                    }
+                }
+
+                it("should fail to yield a renewed access token with a min ttl greater than its expiry") {
+                    credentials = Credentials(accessToken: AccessToken, refreshToken: RefreshToken, expiresIn: Date(timeIntervalSinceNow: ExpiresIn))
+                    _ = credentialsManager.store(credentials: credentials)
+                    waitUntil(timeout: 2) { done in
+                        credentialsManager.credentials(withScope: nil, minTTL: 100) { error, newCredentials in
+                            expect(error).toNot(beNil())
```
lbalmaceda, line 564: let's cover the error message as well

Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -454,8 +477,99 @@ class CredentialsManagerSpec: QuickSpec {
...
+                            done()
+                        }
+                    }
+                }
+
+                it("should fail to yield a renewed access token with a min ttl greater than its expiry") {
+                    credentials = Credentials(accessToken: AccessToken, refreshToken: RefreshToken, expiresIn: Date(timeIntervalSinceNow: ExpiresIn))
+                    _ = credentialsManager.store(credentials: credentials)
+                    waitUntil(timeout: 2) { done in
+                        credentialsManager.credentials(withScope: nil, minTTL: 100) { error, newCredentials in
```
lbalmaceda, line 563: since `minTtl` for this scenario should be greater than `ExpiresIn`, let's reference this variable and add something to it instead of hardcoding the `100`.

Auth0/CredentialsManager.swift
lbalmaceda: Checking again, this method looks package-private, not public. Then you might want to add the `minTtl` argument to the public one and invoke this -> `hasValid()`

Auth0/CredentialsManager.swift
Widcket: It's a number to mean "no number" or "error without error code". I took it from other `NSError` instances on some tests.
Regarding the error message, can you please provide an alternative?

Auth0/CredentialsManager.swift
Widcket: It's package-private so it can be unit-tested. It performs a distinct check, so it may be best for it to be in a separate method.

Auth0Tests/CredentialsManagerSpec.swift
Widcket: Sorry, I don't follow, can you please expand?

Auth0Tests/CredentialsManagerSpec.swift
Widcket: I think it's acceptable in a one-off unlike `ExpiresIn`, which is used everywhere in this file.

Tauseef-TL:
@cocojoe From initial review, it appears that this should cover our usecases/requirements. I will need to spike this with our codebase in order to fully verify this.

Auth0/CredentialsManager.swift
lbalmaceda: > It's a number to mean "no number" or "error without error code"

by keeping it like that it's a "magic number". Give it a name and it's no longer one. If it's used in other places as well, then extract it into a public constant or enum and reference that instead. That way, the name describes the intent of this value.

> provide an alternative?

You can point out the alternatives to resolve this issue. For example.

"The lifetime of the renewed Access Token is less than minTTL requested. Increase the 'Token Expiration' setting of your Auth0 API in the dashboard or request a lower minTTL."

Auth0Tests/CredentialsManagerSpec.swift
lbalmaceda: swap the scope values
```java
let credentials = Credentials(scope: "openid email profile")
expect(credentialsManager.hasScopeChanged(credentials, than: "openid email")).to(beTrue())
```

lbalmaceda:
@Widcket I cannot comment on the diff because this method hasn't changed, but note that it also needs to accept the `minTTL` value since this is the only way the developer has to check for existing non-expired credentials **that require a minTTL** without having to take them from the keystore or refreshing them (what the method you changed does).

https://github.com/auth0/Auth0.swift/blob/b02ee119124e4b78d2339314ded5421517b3a987/Auth0/CredentialsManager.swift#L123

Auth0Tests/CredentialsManagerSpec.swift
cocojoe: Much like code, okay if it's a one-off use. If this 100 value is showing up more than once it should be a const. I read this as a suggestion vs a required change.

Auth0Tests/CredentialsManagerSpec.swift
cocojoe: I believe here Lucho is saying, you can only down scope e.g reduce scopes you can't increase them. So,  let's not show up scoping behaviour in tests. Unless there it's a test to show that it doesn't work.

Auth0/CredentialsManager.swift
cocojoe: So are you suggesting that the error should output the values of the token expiry in seconds and the minTTL requirement? I think that would make it much easier to debug for sure.

lbalmaceda: Sure, that can help. 

In this error scenario, the issue is that the token issued lasts for the next, say, 10 seconds (expiresIn); and the dev requested a token that lasts at least 20 seconds (minTTL). 

So there are two possible solutions: 
a) request a token that lasts less than 10 seconds, the lifetime defined in the dashboard (expiresIn).
b) update the dashboard settings accordingly for the server to start issuing tokens that last at least 20 seconds (minTTL), the minimum required by the dev.

Whatever we log here as the exception message should give a hint or guide to those solutions.

Widcket:
> @Widcket I cannot comment on the diff because this method hasn't changed, but note that it also needs to accept the `minTTL` value since this is the only way the developer has to check for existing non-expired credentials **that require a minTTL** without having to take them from the keystore or refreshing them (what the method you changed does).
> 
> https://github.com/auth0/Auth0.swift/blob/b02ee119124e4b78d2339314ded5421517b3a987/Auth0/CredentialsManager.swift#L123

Good catch, I added the `minTTL` param to the `hasValid` method in https://github.com/auth0/Auth0.swift/pull/399/commits/548ee4472061e2ac1974c36320b6ba2b4004f2fc.

Auth0/CredentialsManager.swift
Widcket: I updated the error message with @lbalmaceda suggestion and added the token expiry in seconds and minTTL for easier debugging in https://github.com/auth0/Auth0.swift/pull/399/commits/04ffa990458113d1f2963b18bb82d2fc82d062c2.
Regarding the error code, it's a one-off because that particular code is not used anywhere else on the SDK target, only on the test target, and we shouldn't really be using `NSError` anyway. We're only using it because adding a new error case would be a breaking change.

Auth0Tests/CredentialsManagerSpec.swift
Widcket: Thanks, that makes sense. Changed in https://github.com/auth0/Auth0.swift/pull/399/commits/04ffa990458113d1f2963b18bb82d2fc82d062c2.

Auth0Tests/CredentialsManagerSpec.swift
Widcket: Thanks, changed in https://github.com/auth0/Auth0.swift/pull/399/commits/04ffa990458113d1f2963b18bb82d2fc82d062c2.

Auth0Tests/CredentialsManagerSpec.swift
Widcket: Thanks, changed in https://github.com/auth0/Auth0.swift/pull/399/commits/548ee4472061e2ac1974c36320b6ba2b4004f2fc.

Widcket:
This is now out on [v1.27.0](https://github.com/auth0/Auth0.swift/releases/tag/1.27.0).

Issue #406, opened 2020-08-17
Add support for web based universal login for macOS 10.11 to 10.14 


MichaelHudgins:
### Describe the problem you'd like to have solved

The support of macOS 10.11 to macOS 10.14 for web auth flow.

### Describe the ideal solution

The ability to use the already preset 10.15 auth flow for previous operating systems

## Alternatives and current work-arounds

I have added a way to achieve this in older operating systems by launching the users default browser.  PR to follow 

### Additional context
This was discussed as not something auth0 wanted to support / put the effort in for previous os support.  I ended up needing to add this support for my own projects so I decided to purpose a change to add these to the official repo. 

Is related to   #375  I will submit a PR to follow with the purposed changes

MichaelHudgins:
See #407

Widcket:
Closing as we won't be pursuing this now, as per https://github.com/auth0/Auth0.swift/pull/407.

Issue #375, opened 2020-04-03
Feature request: Universal Login for MacOS


MichaelHudgins:
### Describe the problem you'd like to have solved

> The ability to use universal login with MacOS applications

### Describe the ideal solution

> The offical cocoa pod located at [https://cocoapods.org/pods/Auth0] to support universal login for MacOS

### Additional context

> As a developer implementing Auth0 into my web, windows, macos, and ios applications it would be most desireable if universal login could be used for all logins.  After looking through documentation it appears MacOS is not a supported platform for universal login.

cocojoe:
@darkyen could be your time to shine 😄

Widcket:
Hi @MichaelHudgins, the implementation is now up in the `macos-web-auth/implementation` branch. Can you please try it out?

MichaelHudgins:
Thanks for the feature, I’ll have a good chance to integrate and test this out on Monday.

MichaelHudgins:
@Widcket I got a chance to test out the branch this morning and everything is looking good.  I was able to get the test app working with my credentials returning back all tokens as expected.  Thanks for the hard work!  I saw the unit tests are targeting 10.12, is this the version that the webauth flow is expected to target?

Widcket:
@MichaelHudgins great! The webauth flow targets [macOS 10.15+](https://developer.apple.com/documentation/authenticationservices/aswebauthenticationsession) (Catalina).

MichaelHudgins:
@Widcket a question on handling older operating systems, would a custom app scheme be a viable alternative `ASWebAuthenticationSession` ?  I realize there is the possibility that a malicious actor could attempt to register the same scheme but in that case I believe PKCE should prevent a compromised redirect from being used.  

I have the option of splitting the ui for old users and new users but unfortunately 10.15 has not had as rapid adoption in our industry and older macOS needs support.

Widcket:
@MichaelHudgins unfortunately we do not have plans for supporting older versions of macOS.

PR #407, opened 2020-08-17
Added webauth support for macOS 10.11 to 10.14

MichaelHudgins:
This adds webauth support for macOS 10.11 to 10.14.  For these operating systems auth will open via the user's default browser to perform the web based auth and use an older method of listening to a callback.

### Changes

_Auth0.xcodeproj/project.pbxproj:_ Changed to include added AuthenticationLegacySession.swift
_AuthenticationLegacySession.swift_: Session subclass to handle this legacy auth flow
_DesktopWebAuth.swift_: Modification to this class to handle the 10.11 to 10.14 flow
_OAuth2Mac/AppDelegate.swift_: Modification of the delegate to show how to use this flow for older operating systems
_OAuth2Mac/Info.plist_: CFBundleTypeRole changed to editor as it appears required for some operating systems to register the new callback
_README.md_: updated to show how to use these changes 
### References

See
#406 
#375 

### Testing

This test does not add unit test coverage.  The control path for it is only followed if the operating system is prior to 10.15 
It has been tested on the latest OS but the control path does not get utilized normally as it targets specifically older operating systems.   To test the flow on operating systems 10.15 and further please change line 58 in DesktopWebAuth.swift `if #available(macOS 10.15, *){` to `if #available(macOS 10.16, *){`

### Checklist


MichaelHudgins:
The basis of this change is to directly open the authorization url in the user's browser if their operating system is too old for the new AuthenticationServices flow.  
To my knowledge the primary security implication is that the callback is not as safe as the newer one because it can be more easily forged by another programming running on the system that registers the callback back before the desired application can, it is my understanding that this should be fully mitigated by the proper use PKCE of this library.

Widcket:
@hzalaz can you please take a look?

cocojoe:
Thanks @MichaelHudgins for this PR. We explored this in the past and wasn't an option we wanted to go with due to potential security concerns. I will raise this into a product conversation internally to see if we want to revisit.  However, for now this is not something we want to add to this library.

PR #410, opened 2020-08-20
Use 'connection' parameter for webauth

lbalmaceda:
### Changes
I was working on this on my side as well and couldn't find any docs that backed that this property exists for /authorize. It seems to be called "connection" all the time. Even when it refers to the connection or realm (where to go and find the username/password pair)

### References

https://auth0.com/docs/api/authentication#login

### Testing

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -432,10 +432,10 @@ Auth0
                     DispatchQueue.main.async {
                         Auth0
                             .webAuth()
-                            .parameters(["connection": realm,
-                                        "login_hint": email])
-                                        // ☝🏼 So the user doesn't have to type it again
+                            .connection(realm)
                             .scope(scope)
+                            .parameters(["login_hint": email])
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -432,7 +432,7 @@ Auth0
                     DispatchQueue.main.async {
                         Auth0
                             .webAuth()
-                            .parameters(["realm": realm,
+                            .parameters(["connection": realm,
```
Widcket, line 435: The thing is, there is already a builder method for "connection". Let me test the snippet with that method and get back to you.

Widcket: I tested the snippet with the `connection` method, and updated the snippet in https://github.com/auth0/Auth0.swift/pull/410/commits/412b4c0be58cd22dd2d2ccf4a14c331ac1e7db8b.

Issue #396, opened 2020-07-07
macCatalyst support


piercifani:
### Describe the problem you'd like to have solved

We're building a macCatalyst version of our application, but some of the symbols used inside Auth0 are not available for macCatalyst. Specifically, `SFAuthenticationError` symbols are deprecated since iOS 13 and they should be migrated to `ASWebAuthenticationSessionErrorCode`.
 
### Describe the ideal solution

macCatalyst should be a supported platform, specially since Apple Silicon based Macs are right around the corner and run iOS apps as is. Intel based Macs should also get this treatment with Catalyst.

## Alternatives and current work-arounds

Remove Auth0 from our codebase and build login manually.

### Additional context

Auth0 is integrated with CocoaPods

cocojoe:
Thanks @piercifani we are evaluating this internally 👍  I'll leave this open for now.

Widcket:
Hi @piercifani, I tested with a sample app and it compiles successfully with Xcode Version 11.6 (11E708). [SimpleKeychain](https://github.com/auth0/SimpleKeychain) compiles with a couple of warnings, though, and those will be addressed.

Widcket:
Auth0.swift also produces a few Catalyst-related compiler warnings, and we'll be addressing these as well.

<img width="1012" alt="Screen Shot 2020-08-11 at 21 30 50" src="https://user-images.githubusercontent.com/5055789/89962064-03d87800-dc1a-11ea-8865-86061f002242.png">

piercifani:
Hey! 

I just updated to Auth0 version 1.27.0 and I'm still getting this error when compiling:

<img width="1490" alt="image" src="https://user-images.githubusercontent.com/869981/90002180-d7584680-dc81-11ea-8ae7-fe08281ddeeb.png">

Widcket:
@piercifani are you using Xcode 12 and/or macOS Big Sur?

piercifani:
Yes, latest beta 5 in Big Sur

Widcket:
Thanks, we'll dig more into this.

piercifani:
The issue is that `SFAuthenticationSession` is not available on Catalyst, so you have to use `ASWebAuthenticationSession`. Thankfully it's just a rename

Widcket:
This is now fixed in [1.30.0](https://github.com/auth0/Auth0.swift/releases/tag/1.30.0).

PR #421, opened 2020-10-04
Updated Quick and Nimble

Widcket:
### Changes

This PR updates the test dependencies [Quick](https://github.com/Quick/Quick/releases/tag/v3.0.0) and [Nimble](https://github.com/Quick/Nimble/releases/tag/v9.0.0) to the latest majors, making it possible to test Auth0.swift with Xcode 12. This update will also allow to start addressing the [compilation](https://github.com/auth0/Auth0.swift/issues/397) [issues](https://github.com/auth0/Auth0.swift/issues/396) on BigSur.

**The new majors of Quick and Nimble no longer support Swift versions < 5.2. This means that the _tests_ can no longer be run on Xcode versions < 11.4. The CircleCI config has been updated accordingly.**

### Testing

### Checklist


cocojoe:
So this PR will be a breaking change in that Carthage at the very least will no longer work in Xcode versions below 12?
Do CocoaPods still work, I don't think they do any testing?

Although this is necessary for Xcode 12 and Carthage. This really needs to be called out explicitly in the README that for Xcode <12 use Minor X for Xcode >=12 this minor is needed.

Widcket:
@cocojoe Quick and Nimble are dropping support for Xcode < 11.4, so it would still work with Xcode 11.4+. So no, the minimum supported Xcode does not shift to 12.
Why would the samples have to be updated? They use Cocoapods.

Widcket:
@cocojoe besides, the README currently states that the supported Xcode versions are 11.x. So the only thing that would actually change is that to use Carthage you would have to use Xcode 11.4+, instead of 11.0-11.3. And that just because the library uses Carthage to manage its own dependencies. The Podspec does not even mention Quick and Nimble.
But, I agree that the above (about Carthage) should be added to the README.

cocojoe:
This needs a heavy call out and clear context, if the minimum listed is 11.x then give the circumstances I think it's okay. We just need to be very explicit to avoid any potential confusion.

Widcket:
Correction, it says "Xcode 10.x / 11.x" However, you can't submit apps to the AppStore that were built with Xcode 10.x anymore, so it's effectively EOL'd.

Widcket:
And with the current major being 4 years old, Xcode 10 is not the first Xcode version to stop being supported by this library, I suppose.

Widcket:
@cocojoe I've added a notice to the README.

README.md
```diff
@@ -5,13 +5,14 @@
...
 
 Swift toolkit that lets you communicate efficiently with many of the [Auth0 API](https://auth0.com/docs/api/info) functions and enables you to seamlessly integrate the Auth0 login.
 
-## Important Notice
-[Behaviour changes in iOS 13](https://github.com/auth0/Auth0.swift/pull/297) related to Web Authentication require that developers using Xcode 11 with this library **must** compile using Swift 5.x. This *should* be the default setting applied when updating, unless it has been manually set. However, we recommend checking that this value is set correctly.
+## Important Notices
+- [The latest majors of Quick and Nimble](https://github.com/auth0/Auth0.swift/pull/421) require that developers using Carthage **must** use Xcode 11.4+ only, starting from Auth0.swift [1.29.0](https://github.com/auth0/Auth0.swift/releases/tag/1.29.0).
+- [Behaviour changes in iOS 13](https://github.com/auth0/Auth0.swift/pull/297) related to Web Authentication require that developers using Xcode 11 with this library **must** compile using Swift 5.x. This *should* be the default setting applied when updating, unless it has been manually set. However, we recommend checking that this value is set correctly.
```
cocojoe, line 15: Should we create a historical notices section and move this part to then? iOS 13 feels way less relevant now.

cocojoe (approved):

Issue #306, opened 2019-09-03
Swift package manager support


jessemx109:
### Description

Description of the bug or feature request and why it's a problem. Consider including:

Can you please add support for including the framework through swift package manager

cocojoe:
Hi @jessemx109 it's in our backlog to add 👍

piercifani:
Thanks, we also need this!

jeffc-dev:
Also looking forward to SwiftPM support, hopefully sooner rather than later (as Xcode 11 is now out and SwiftPM is the "officially supported" way of handling dependencies).

Thanks!

rynecheow:
Hi @jeffc-dev - Do we have plans to work on this? I can help if there's a plan. Currently I am stuck as SPM is unable to build Objective-C code (SimpleKeychain).

jeffc-dev:
I have no near-term plans (a bit swamped with other work at the moment), but would be interested in testing if someone comes up with something.

pedrorgr:
+1 on looking forward to this

ryangittings:
+1

amkomatz:
+1 We really could use this

El-Fitz:
+1 Really looking forward to this!

Widcket:
Thanks everyone for the input, but we won’t be moving forward with this ATM. SPM still does not support a way to exclude files by platform, and we have some Objective-C files we’d have to create a new target for (or migrate them to Swift and drop support for Swift < 4.2).
We’ll be happy to revisit this once SPM supports excluding files by platform.

piercifani:
I'm sorry @Widcket , but you can exclude files per platform easily on ObjC+SPM like this:

```objc
#if TARGET_OS_IOS
/// Your platform specific classes
#endif
```

You can check it out here https://github.com/theleftbit/BSWInterfaceKit/blob/develop/Sources/BSWInterfaceKitObjC/include/BSWZoomTransition.h#L8

Would this change your mind?

Widcket:
We've considered workarounds for every issue, but we'd need to gate 25+ files, and that's not viable for us.

sam-w:
@Widcket @cocojoe that's a real shame to hear. I'd like to see this reconsidered at some point.

CypherPoet:
Just an idea, but perhaps an issue with some kind of "Help Wanted" tag can be created for this. 

I anticipate that demand for SPM support is only going to grow, and that seems like some nice low-hanging fruit for an aspiring contributor.

[...]

PR #425, opened 2020-10-16
Added Swift Package Manager support [SDK-2014]

Widcket:
### Changes

This PR adds support for the Swift Package Manager (SPM).
The `Package.swift` file defines 3 targets: `Auth0` for the Swift code, `Auth0ObjectiveC` for the Objective-C code and a test target for the tests. All the WebAuth files were guarded with a conditional compilation condition in the Swift code, and with a preprocessor macro in the Objective-C code.
Also, a separate implementation of RSA signature generation and verification was added for macOS 10.12+. `SecKeyCreateSignature` was used instead of `SecKeyRawSign`, and `SecKeyVerifySignature` instead of `SecKeyRawVerify`.
The issues with Catalyst on Big Sur were addressed as well.

No public APIs were added or modified.

### References

- Fixes https://github.com/auth0/Auth0.swift/issues/306
- Fixes https://github.com/auth0/Auth0.swift/issues/396
- Fixes https://github.com/auth0/Auth0.swift/issues/397

### Testing

The changes were tested manually as follows:

- New apps were created for iOS, macOS, tvOS and watchOS
- Auth0.swift was added to these apps
- On iOS and macOS, a WebAuth login/logout was performed; also with Catalyst as well
- On tvOS and watchOS, a call to the JWKS endpoint was performed using the Authentication API client

Those tests were executed with:

- SPM and Xcode 12 on macOS Catalina + BigSur
- CocoaPods and Xcode 12 on macOS Catalina + BigSur
- Carthage and Xcode 11 on macOS Catalina

Also, `swift test` was added to the **build-macOS-Swift-5.2** CI job.

### Checklist


Auth0.podspec
```diff
@@ -1,12 +1,12 @@
...
-  'Auth0/A0RSA.h',
-  'Auth0/A0RSA.m',
-  'Auth0/A0SHA.h',
-  'Auth0/A0SHA.m',
+  'Auth0/ObjectiveC/A0ChallengeGenerator.h',
+  'Auth0/ObjectiveC/A0ChallengeGenerator.m',
+  'Auth0/ObjectiveC/A0RSA.h',
+  'Auth0/ObjectiveC/A0RSA.m',
+  'Auth0/ObjectiveC/A0SHA.h',
+  'Auth0/ObjectiveC/A0SHA.m',
```
Widcket, line 9: The ObjectiveC files were moved into a folder, to be able to create a SPM target with them.

Auth0.podspec
```diff
@@ -75,22 +75,24 @@ Pod::Spec.new do |s|
...
+  s.ios.source_files = 'Auth0/*.{swift,h,m}', 'Auth0/ObjectiveC/*.{h,m}'
   s.ios.frameworks = 'UIKit', 'SafariServices', 'LocalAuthentication'
   s.ios.weak_framework = 'AuthenticationServices'
   s.ios.dependency 'SimpleKeychain'
   s.ios.dependency 'JWTDecode'
   s.ios.exclude_files = macos_files
   s.ios.pod_target_xcconfig = {
-    'SWIFT_ACTIVE_COMPILATION_CONDITIONS' => 'WEB_AUTH_PLATFORM'
+    'SWIFT_ACTIVE_COMPILATION_CONDITIONS' => 'WEB_AUTH_PLATFORM',
+    'GCC_PREPROCESSOR_DEFINITIONS' => '$(inherited) WEB_AUTH_PLATFORM=1'
```
Widcket, line 86: The GCC preprocessor is necessary for `WEB_AUTH_PLATFORM ` to be available in ObjectiveC code.

Auth0.xcodeproj/project.pbxproj
```diff
@@ -2639,8 +2654,12 @@
 					"$(inherited)",
 					"$(PROJECT_DIR)/Carthage/Build/iOS",
 				);
+				GCC_PREPROCESSOR_DEFINITIONS = (
+					"$(inherited)",
+					"WEB_AUTH_PLATFORM=1",
+				);
 				INFOPLIST_FILE = Auth0Tests/Info.plist;
-				IPHONEOS_DEPLOYMENT_TARGET = 10.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 9.0;
```
Widcket, line 2662: Since SPM defines the same platform versions for all targets, the test targets were changed here to match the same minimum platform versions as the library. Previously, the test targets had higher requirements to avoid having to guard the code everywhere a newer API was used (generally crypto APIs). So the code guards were added as well.

Auth0/MobileWebAuth.swift
```diff
@@ -114,12 +115,16 @@ final class MobileWebAuth: BaseWebAuth, WebAuth {
                                           handler: handler,
                                           callback: callback)
             }
+            #if !targetEnvironment(macCatalyst)
```
Widcket, line 118: Catalyst does not support SafariServices, only AuthenticationServices.

Auth0/ObjectiveC/A0RSA.m
```diff
@@ -38,6 +39,17 @@ - (nullable instancetype)initWithKey: (SecKeyRef)key {
...
+    NSData * signature;
+    CFErrorRef error = NULL;
+
+    signature = CFBridgingRelease(SecKeyCreateSignature(self.key,
+                                                        kSecKeyAlgorithmRSASignatureDigestPKCS1v15SHA256,
+                                                        (__bridge CFDataRef) plainData,
+                                                        &error));
+
+    return signature;
+    #else
```
Widcket, line 52: `SecKeyRawSign ` was never officially supported on macOS and on the latest versions it was removed. So here we add an implementation with a mac0S-supported API.

Auth0/ObjectiveC/A0RSA.m
```diff
@@ -56,16 +68,30 @@ - (NSData *)sign:(NSData *)plainData {
...
+    BOOL result = NO;
+    CFErrorRef error = NULL;
+
+    result = SecKeyVerifySignature(self.key,
+                                   kSecKeyAlgorithmRSASignatureDigestPKCS1v15SHA256,
+                                   (__bridge CFDataRef) plainData,
+                                   (__bridge CFDataRef) signature,
+                                   &error);
+    return result;
+    #else
```
Widcket, line 85: Same as above.

Auth0/ObjectiveC/A0RSA.m
cocojoe: Could argue this should be a separate PR dealing with this migration for macOS support in Xcode 12. Please mention in changes.

Widcket: I added it to the changes section. This was not changed arbitrarily, but because otherwise the library would not compile for macOS and I wouldn't be able to test it with Xcode 12.

cocojoe (approved):
This is mostly a huge amount of gating, at a glance looks okay the real validation is in the manual testing results you posted for each of the dependency managers.

PR #433, opened 2020-11-11
Use Swift's built-in Result type

ejensen:
### Changes

Replaces the conflicting public `Result` type with the `Result` type built into the [Swift Standard Library](https://developer.apple.com/documentation/swift/swift_standard_library)

This helps prevents type conflicts while minimizing changes to the existing Auth0 interface.
This can be taken a step further by removing the shims that convert the `.success(result:)` to `.success()` and `.failure(error:)` to `.failure()` as done in https://github.com/auth0/Auth0.swift/compare/master...kiva:swift-result-type-no-shims

### References

- [Swift's Result type documentation](https://developer.apple.com/documentation/swift/result)

### Testing

### Checklist

### DIFF
--- a/Auth0/Result.swift
+++ b/Auth0/Result.swift
@@ -24,11 +24,15 @@ import Foundation
 
 /**
  Result object for Auth0 APIs requests
-
- - Success: request completed successfuly with it's response body
- - Failure: request failed with a specific error
  */
-public enum Result<T> {
-    case success(result: T)
-    case failure(error: Error)
+public typealias Result<T> = Swift.Result<T, Error>
+
+extension Result {
+    static func success(result: Success) -> Self {
+        return .success(result)
+    }
+
+    static func failure(error: Failure) -> Self {
+        return .failure(error)
+    }
 }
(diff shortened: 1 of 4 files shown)


Auth0/Result.swift
```diff
@@ -24,11 +24,15 @@ import Foundation
...
-
- - Success: request completed successfuly with it's response body
- - Failure: request failed with a specific error
  */
-public enum Result<T> {
-    case success(result: T)
-    case failure(error: Error)
+public typealias Result<T> = Swift.Result<T, Error>
+
+extension Result {
```
ejensen, line 30: These shims can be removed by convert the `.success(result:)` to `.success()` and `.failure(error:)` to `.failure()` as done in https://github.com/auth0/Auth0.swift/compare/master...kiva:swift-result-type-no-shims

Widcket:
@ejensen thanks for the PR. Unfortunately this is a breaking change, as the current major still supports Swift 4.x. However, a new major is being planned, and we intend to use the Swift 5 Result type in it.

Issue #437, opened 2020-12-29
Provide a synchronous way of retrieving the currently stored Credentials 


JanC:
### Describe the problem you'd like to have solved

Hey,
as a SDK user, I'd like to retrieve the currently stored `Credentials` in a **synchronous** way so that I can access some metadata stored in `Credentials.idToken` regardless of the credentials validity

### Describe the ideal solution

In the same way `Auth0.CredentialsManager` provides a way to store the credentials, it could provide a sync method to retrieve them:

```swift
func credentials() -> Credentials? {
    guard let data = self.storage.data(forKey: self.storeKey),
          let credentials = NSKeyedUnarchiver.unarchiveObject(with: data) as? Credentials else {
        return nil
    }
    return credentials
}
```

## Alternatives and current work-arounds

A workaround is to use directly the `A0SimpleKeychain` without relying on the `Auth0.CredentialsManager` and use the same `storeKey` as the one passed in the init of `Auth0.CredentialsManager`:

```swift
private extension A0SimpleKeychain {
    func credentials(_ storeKey: String) -> Credentials? {
        guard let data = self.data(forKey: storeKey),
              let credentials = NSKeyedUnarchiver.unarchiveObject(with: data) as? Credentials else {
            return nil
        }
        return credentials
    }
}
```

However, this unnecessarily exposes the `A0SimpleKeychain` as well as the way the credentials are persisted using `NSKeyedUnarchiver`. So if the `Auth0.CredentialsManager` way of archiving the credentials would change, the above code would likely fail.

Widcket:
Hi @JanC, we'll look into adding a way to retrieve the user profile separately from the credentials in the next major. This is unlikely to happen for the current 1.x. Will close this for now.

JanC:
thanks for your answer @Widcket 

> This is unlikely to happen for the current 1.x.

Even if I make a PR ? :) 

cheers

Widcket:
We'll discuss this internally and come back to you.

Widcket:
@JanC thanks for the offer, we'll happily take a PR that implements this. Might probably want to return an instance of [UserInfo](https://github.com/auth0/Auth0.swift/blob/master/Auth0/UserInfo.swift). Please ping if you have any questions or want to discuss something further.

JanC:
ok, I have a branch somewhere with pretty much the implementation I posted in my original comment. I'll clean it up + write tests and will submit a PR. I'll look at the UserInfo

PR #459, opened 2021-02-25
[437] Provide a synchronous way of retrieving the currently stored Credentials

JanC:
### Changes
This PR adds a synchronous way of retrieving the store credentials. Please see https://github.com/auth0/Auth0.swift/issues/437 for the motivation and use case.

An example of the usage of the added method would be

```swift
let credentialsManager = CredentialsManager(authentication: Auth0.authentication())

if let jwt = try? credentialsManager
    .credentials
    .flatMap(\.idToken)
    .map(decode(jwt:)) {
    // do something with a locally stored JWT
    print(jwt)
}
```

@Widcket [mentioned ](https://github.com/auth0/Auth0.swift/issues/437#issuecomment-785977881) that it might be worth it to return the `UserInfo` instead of the `Credentials`. However, I wonder if that would be useful as this new method is purely related to the credentials.

### Testing

### Checklist


Auth0/CredentialsManager.swift
```diff
@@ -172,6 +172,16 @@ public struct CredentialsManager {
     }
     #endif
 
+    /// Retrieve credentials from keychain without any validation of the expiration.
+    /// No attempts to renew the credentials will be made so it's up to to caller to make the expiration date checks if they are relevant.
+    public var credentials: Credentials? {
```
shaps80, line 177: Sorry, I know I'm not a contributor to this repo. But as a user of this library and a library author myself, I'd recommend a better name here that's more indicative of its unsafe nature. Unsafe might be the wrong term, but hopefully you get my meaning 😬

Widcket:
We’ll be tackling the API in detail next week.

Issue #485, opened 2021-07-09
Support various LAPolicy to enable fallback to pin-code when validating Biometric


krodak:
### Describe the problem you'd like to have solved

Currently, if Face Id / Touch Id fails and iOS displays "Enter passcode" UI, tapping on **`Enter password`** triggers instant biometrics failure with `CredentialsManagerError.touchFailed(Error)`

<img src="https://user-images.githubusercontent.com/1955364/125054000-051e5900-e0a6-11eb-86f2-9eba704bfda6.jpg" width="320">

This happens because Auth0 favours `LAPolicy.deviceOwnerAuthenticationWithBiometrics`  evaluation policy within `BioAuthentication`. 

### Describe the ideal solution

We should be able to customise `BioAuthentication` with either `LAPolicy.deviceOwnerAuthenticationWithBiometrics` or `LAPolicy.deviceOwnerAuthentication` as an evaluation policy

As per Apple's documentation **`LAPolicy.deviceOwnerAuthentication`**
> /// Device owner is going to be authenticated by biometry or device passcode.
    ///
    /// @discussion Biometric or passcode authentication is required. If the biometry is available, enrolled and
    ///             not locked out, users are asked for it first. Otherwise they are asked to enter device
    ///             passcode. If passcode is not enabled, policy evaluation will fail with LAErrorPasscodeNotSet.
    ///
    ///             Biometric authentication dialog behaves similarly as the one used by
    ///             LAPolicyDeviceOwnerAuthenticationWithBiometrics. However, instead of "Enter Password"
    ///             button there is "Enter Passcode" button which, when tapped, switches the authentication
    ///             method and allows users to enter device passcode.
    ///
    ///             Passcode authentication will get locked after 6 unsuccessful attempts with progressively
    ///             increased backoff delay.

Diving into code, this would mean that we are able to have `validateBiometric` as follow:

```swift
 @available(iOS 9.0, macOS 10.15, *)
    func validateBiometric(callback: @escaping (Error?) -> Void) {
        self.authContext.evaluatePolicy(LAPolicy.deviceOwnerAuthentication, localizedReason: self.title) {
            guard $1 == nil else { return callback($1) }
            callback($0 ? nil : LAError(LAError.authenticationFailed))
        }
    }
```

Visualisation of expected flow:

https://user-images.githubusercontent.com/1955364/125054187-39921500-e0a6-11eb-97f9-d9cb40b478d7.MOV

### Alternate solution

[...]

PR #486, opened 2021-07-14
[FR-485] Add support for custom LAPolicy when evaluating biometry aut…

krodak:
### Changes

Currently, if Face Id / Touch Id fails and iOS displays "Enter passcode" UI, tapping on **`Enter password`** triggers instant biometrics failure with `CredentialsManagerError.touchFailed(Error)`

This happens because Auth0 favours `LAPolicy.deviceOwnerAuthenticationWithBiometrics`  evaluation policy within `BioAuthentication`. 

#### Updated classes:
- `BioAuthentication `
- `CredentialsManager`

### References

Feature request #485 

### Testing

More on why I haven't added tests below 🙏 

### Checklist

### Issues
To be honest I couldn't build test target using Carthage and xcframeworks with current setup. I've tried `carthage update` to work around some of the initial issues, that helped building `Nimble` for main target, but still I couldn't build test target. 

I've pushed my idea for adding testability for included changes, but I see that input parameters testing might not be a part of convention, so please let me know whether I should revert those. 🙏  I admit I haven't tested those locally, but I see they are passing on CI. I'd be happy to retest locally, if I could get more information on what commands I should run to be able to run test target after fresh repo checkout.

### DIFF
--- a/Auth0/CredentialsManager.swift
+++ b/Auth0/CredentialsManager.swift
@@ -71,20 +71,23 @@ public struct CredentialsManager {
     ///   - fallbackTitle: fallback message to display in TouchID prompt after a failed match
     #if WEB_AUTH_PLATFORM
     @available(*, deprecated, message: "see enableBiometrics(withTitle title:, cancelTitle:, fallbackTitle:)")
+    @available(iOS 9.0, macOS 10.15, *)
     public mutating func enableTouchAuth(withTitle title: String, cancelTitle: String? = nil, fallbackTitle: String? = nil) {
         self.enableBiometrics(withTitle: title, cancelTitle: cancelTitle, fallbackTitle: fallbackTitle)
     }
     #endif
 
+    #if WEB_AUTH_PLATFORM
     /// Enable Biometric Authentication for additional security during credentials retrieval
     ///
     /// - Parameters:
     ///   - title: main message to display when Touch ID is used
     ///   - cancelTitle: cancel message to display when Touch ID is used (iOS 10+)
     ///   - fallbackTitle: fallback message to display when Touch ID is used after a failed match
-    #if WEB_AUTH_PLATFORM
-    public mutating func enableBiometrics(withTitle title: String, cancelTitle: String? = nil, fallbackTitle: String? = nil) {
-        self.bioAuth = BioAuthentication(authContext: LAContext(), title: title, cancelTitle: cancelTitle, fallbackTitle: fallbackTitle)
+    ///   - evaluationPolicy: policy to be used for authentication policy evaluation
+    @available(iOS 9.0, macOS 10.15, *)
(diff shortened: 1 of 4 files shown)


Auth0/CredentialsManager.swift
```diff
@@ -76,15 +76,16 @@ public struct CredentialsManager {
     }
     #endif
 
+    #if WEB_AUTH_PLATFORM
```
krodak, line 79: **Explanation for unrelated change:**
By moving this above comment section, comment is now visible if we investigate function's signature 🙏

Auth0/CredentialsManager.swift
```diff
@@ -76,15 +76,17 @@ public struct CredentialsManager {
...
     ///
     /// - Parameters:
     ///   - title: main message to display when Touch ID is used
     ///   - cancelTitle: cancel message to display when Touch ID is used (iOS 10+)
     ///   - fallbackTitle: fallback message to display when Touch ID is used after a failed match
-    #if WEB_AUTH_PLATFORM
-    public mutating func enableBiometrics(withTitle title: String, cancelTitle: String? = nil, fallbackTitle: String? = nil) {
-        self.bioAuth = BioAuthentication(authContext: LAContext(), title: title, cancelTitle: cancelTitle, fallbackTitle: fallbackTitle)
+    ///   - evaluationPolicy: policy to be used for authentication policy evaluation
+    @available(iOS 9.0, macOS 10.15, *)
```
krodak, line 87: This would be required if we want to pass `LAPolicy`. But given that most of functionality of `BioAuthentication` struct is limited to these system versions, I think this should be ok.

Alternative would be to just have separate method with additional parameter, let me know if I should add it, if current solution is not sufficient.

README.md
```diff
@@ -337,6 +337,12 @@ You can enable an additional level of user authentication before retrieving cred
 credentialsManager.enableBiometrics(withTitle: "Touch to Login")
 ```
 
+If needed, you are able to specify specific `LAPolicy` to be used - i.e. you might want to support FaceID, but allow fallback to  pin code.
```
Widcket, line 340:
```suggestion
If needed, you are able to specify specific `LAPolicy` to be used - i.e. you might want to support FaceID, but allow fallback to pin code.
```

Auth0/CredentialsManager.swift
Widcket: The current solution is fine.

Widcket (approved):
Great, thanks!

krodak:
@Widcket could I ask how after `Auth0` releases new patch versions? Do you think there is a chance to have a patch release that would include this PR? 🙏

Widcket:
@krodak I'll put out a new release later today.

Widcket:
This is out now in version `1.35.0`.

PR #487, opened 2021-07-15
Expose recovery code [SDK-2661]

Widcket:
### Changes

The [documentation](https://auth0.com/docs/api/authentication#verify-with-recovery-code) for the **Verify with Recovery Code** endpoint states that the response might include a new `recovery_code` that the app should display to the user. The `Credentials` class currently does not include that field, so it is not exposed to the developer.

Screen Shot 2021-07-15 at 17 50 41

### Testing

Unit tests were added, and the change was also tested manually on an iOS simulator:

Screen Shot 2021-07-15 at 17 46 10

### Checklist

### DIFF
--- a/Auth0/Credentials.swift
+++ b/Auth0/Credentials.swift
@@ -38,18 +38,21 @@ public class Credentials: NSObject, JSONObjectPayload, NSSecureCoding {
     @objc public let expiresIn: Date?
     /// If the API allows you to request new access tokens and the scope `offline_access` was included on Auth
     @objc public let refreshToken: String?
-    // Token that details the user identity after authentication
+    /// Token that details the user identity after authentication
     @objc public let idToken: String?
-    // Granted scopes, only populated when a requested scope or scopes was not granted and Auth is OIDC Conformant
+    /// Granted scopes, only populated when a requested scope or scopes was not granted and Auth is OIDC Conformant
     @objc public let scope: String?
+    /// MFA recovery code that the application must display to the end-user to be stored securely for future use
+    @objc public let recoveryCode: String?
 
-    @objc public init(accessToken: String? = nil, tokenType: String? = nil, idToken: String? = nil, refreshToken: String? = nil, expiresIn: Date? = nil, scope: String? = nil) {
+    @objc public init(accessToken: String? = nil, tokenType: String? = nil, idToken: String? = nil, refreshToken: String? = nil, expiresIn: Date? = nil, scope: String? = nil, recoveryCode: String? = nil) {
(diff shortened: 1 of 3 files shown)


evansims (approved):

Auth0/Credentials.swift
```diff
@@ -38,18 +38,21 @@ public class Credentials: NSObject, JSONObjectPayload, NSSecureCoding {
...
+    /// Token that details the user identity after authentication
     @objc public let idToken: String?
-    // Granted scopes, only populated when a requested scope or scopes was not granted and Auth is OIDC Conformant
+    /// Granted scopes, only populated when a requested scope or scopes was not granted and Auth is OIDC Conformant
     @objc public let scope: String?
+    /// MFA recovery code that the application must display to the end-user to be stored securely for future use
+    @objc public let recoveryCode: String?
 
-    @objc public init(accessToken: String? = nil, tokenType: String? = nil, idToken: String? = nil, refreshToken: String? = nil, expiresIn: Date? = nil, scope: String? = nil) {
+    @objc public init(accessToken: String? = nil, tokenType: String? = nil, idToken: String? = nil, refreshToken: String? = nil, expiresIn: Date? = nil, scope: String? = nil, recoveryCode: String? = nil) {
```
lbalmaceda, line 48: @Widcket Given that this class and constructor is public. If someone is instantiating a `Credentials` object today without using named parameters, wouldn't this change be a breaking change?

Widcket: No, check the tests. When the value is not provided, it ends up as `nil`.

lbalmaceda: @Widcket The tests in `CredentialsSpec.swift` all use the constructor that takes a `json` input.

https://github.com/auth0/Auth0.swift/pull/487/files/8bc15053675b00025de5f162f40b9241f8ed82c6#diff-77a2df3b8a702b22e6181fc1f662e39a2012290011ad69caf52cb08bbb8c04c3R42

If not that file, what is the file I should be checking that makes use of the alternate constructor that takes all the arguments?

Widcket: The JSON constructor is calling the one with the parameters: https://github.com/auth0/Auth0.swift/blob/master/Auth0/Credentials.swift#L68

> If someone is instantiating a Credentials object today without using named parameters

The parameter is expressed as `recoveryCode: String? = nil`, so there's no instantiating it without named parameters. The parameters are there, even if you do `Credentials()`:

Screen Shot 2021-07-16 at 11 12 47

PR #529, opened 2021-11-02
Call the login callback block with failure if state is missing from callback URL after logging in

acoomans:
### Changes

- changes the login callback behavior: it is now called even if the returned callbackURL has a missing state in its query string
- added convenience function to tell if state is missing

### References

- support ticket: https://support.auth0.com/tickets/00504734

### Testing

Not tested

### Checklist

### DIFF
--- a/Auth0/AuthenticationError.swift
+++ b/Auth0/AuthenticationError.swift
@@ -147,6 +147,11 @@ public class AuthenticationError: Auth0Error, CustomStringConvertible {
         return self.code == "requires_verification"
     }
 
+    /// When state is missing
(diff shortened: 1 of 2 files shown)


acoomans:
Note that this does not solve the root problem, which is that we're not getting the `state` parameter in the `callbackURL`, but at least we get an failure with an error

Widcket:
Hi @acoomans, thanks for raising this. 
We cannot accept this PR, which seems to be bringing into the SDK a workaround for a situation that appears to be specific to your app. Besides that, the tests are failing for iOS and macOS, and the changes themselves are not tested.

PR #531, opened 2021-11-03
Drop support for Swift < 5.3 [SDK-2896]

Widcket:
### Changes

The minimum Swift version supported was raised to 5.3.

### References

### Testing

### Checklist

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.2
+// swift-tools-version:5.3
 
 import PackageDescription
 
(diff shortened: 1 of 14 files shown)


Package.swift
```diff
@@ -29,7 +29,7 @@ let package = Package(
             name: "Auth0", 
             dependencies: ["SimpleKeychain", "JWTDecode", "Auth0ObjectiveC"], 
             path: "Auth0",
-            exclude: ["ObjectiveC"],
+            exclude: ["ObjectiveC", "Info.plist", "Info-tvOS.plist"],
```
Widcket, line 32: Swift tools version 5.3+ require declaring (bundling) any resource files. If there are non-code files that are not being bundled, it will produce a warning. So we need to exclude those files.
See https://developer.apple.com/documentation/swift_packages/bundling_resources_with_a_swift_package

adamjmcgrath (approved):

PR #533, opened 2021-11-04
Make some `Credentials` properties non-optional [SDK-2900]

Widcket:
### Changes

The following properties of the Credentials class were made non-optional, using the empty string ("”) as the default value in the initializer:

- `accessToken`
- `tokenType`
- `expiresIn`
- `idToken`

`scope`, `refreshToken`, and `recoveryCode` remain optional.

Consequently, the ID Token presence check was removed from the ID Token validation logic.

### References

### Testing

### Checklist

### DIFF
--- a/Auth0/CredentialsManager.swift
+++ b/Auth0/CredentialsManager.swift
@@ -235,10 +235,7 @@ public struct CredentialsManager {
     }
 
     func hasExpired(_ credentials: Credentials) -> Bool {
-        if credentials.expiresIn < Date() { return true }
-        if let jwt = try? decode(jwt: credentials.idToken) { return jwt.expired }
-
-        return false
+        return credentials.expiresIn < Date()
     }
 
     func hasScopeChanged(_ credentials: Credentials, from scope: String?) -> Bool {
--- a/Auth0/CredentialsManager.swift
+++ b/Auth0/CredentialsManager.swift
@@ -57,8 +57,7 @@ public struct CredentialsManager {
     /// - Important: Access to this property will not be protected by Biometric Authentication.
     public var user: UserInfo? {
         guard let credentials = retrieveCredentials(),
-            let idToken = credentials.idToken,
-            let jwt = try? decode(jwt: idToken) else { return nil }
+              let jwt = try? decode(jwt: credentials.idToken) else { return nil }
 
         return UserInfo(json: jwt.body)
     }
@@ -196,8 +195,7 @@ public struct CredentialsManager {
     }
 
     private func retrieveCredentials(withScope scope: String?, minTTL: Int, parameters: [String: Any] = [:], callback: @escaping (CredentialsManagerError?, Credentials?) -> Void) {
-        guard let credentials = retrieveCredentials(),
-              let expiresIn = credentials.expiresIn else { return callback(.noCredentials, nil) }
+        guard let credentials = retrieveCredentials() else { return callback(.noCredentials, nil) }
         guard self.hasExpired(credentials) ||
                 self.willExpire(credentials, within: minTTL) ||
                 self.hasScopeChanged(credentials, from: scope) else { return callback(nil, credentials) }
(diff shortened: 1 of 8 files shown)


Auth0Tests/CredentialsSpec.swift
```diff
@@ -39,8 +39,7 @@ class CredentialsSpec: QuickSpec {
         describe("init from json") {
 
             it("should have all tokens and token_type") {
-                let credentials = Credentials(json: ["access_token": AccessToken, "token_type": Bearer, "id_token": IdToken, "refresh_token": RefreshToken, "expires_in" : expiresIn, "scope": Scope, "recovery_code": RecoveryCode])
-                expect(credentials).toNot(beNil())
```
Widcket, line 43: When created from JSON, the instance of `Credentials` cannot be `nil`. That assertion is superflous.

Auth0Tests/CredentialsSpec.swift
```diff
@@ -50,27 +49,17 @@ class CredentialsSpec: QuickSpec {
...
-                expect(credentials.idToken).to(beNil())
-                expect(credentials.expiresIn).to(beNil())
+                expect(credentials.idToken) == IdToken
+                expect(credentials.refreshToken).to(beNil())
+                expect(credentials.expiresIn).to(beCloseTo(Date(), within: 5))
                 expect(credentials.scope).to(beNil())
                 expect(credentials.recoveryCode).to(beNil())
             }
 
-            it("should have id_token") {
```
Widcket, line 64: All properties (nullable and non-nullable) are tested in the two previous test cases.

Auth0/OAuth2Grant.swift
```diff
@@ -154,20 +154,20 @@ struct PKCE: OAuth2Grant {
...
         let clientId = authentication.clientId
-        let isFrontChannelIdTokenExpected = responseType.contains(.idToken)
         let validatorContext = IDTokenValidatorContext(authentication: authentication,
                                                        issuer: self.issuer,
                                                        leeway: self.leeway,
                                                        maxAge: self.maxAge,
                                                        nonce: self.defaults["nonce"],
                                                        organization: self.organization)
-        validateFrontChannelIDToken(idToken: idToken, for: responseType, with: validatorContext) { error in
+        validateFrontChannelIDToken(idToken: frontChannelIdToken, for: responseType, with: validatorContext) { error in
```
Widcket, line 170: This method (and everything related to front-channel ID Tokens) will be removed in a future PR, as we'll only support `code`.

Auth0/IDTokenValidator.swift
```diff
@@ -59,23 +59,20 @@ struct IDTokenValidator: JWTAsyncValidator {
 }
 
 enum IDTokenDecodingError: LocalizedError {
-    case missingToken
     case cannotDecode
 
     var errorDescription: String? {
         switch self {
-        case .missingToken: return "ID token is required but missing"
```
Widcket, line 67: There's no point in validating the ID Token presence if there will always be an ID Token in the response.

Auth0/CredentialsManager.swift
```diff
@@ -233,21 +231,12 @@ public struct CredentialsManager {
...
     func hasExpired(_ credentials: Credentials) -> Bool {
-        if let expiresIn = credentials.expiresIn {
-            if expiresIn < Date() { return true }
-        }
-
-        if let token = credentials.idToken, let jwt = try? decode(jwt: token) {
-            return jwt.expired
-        }
+        if credentials.expiresIn < Date() { return true }
+        if let jwt = try? decode(jwt: credentials.idToken) { return jwt.expired }
```
Widcket, line 239: @adamjmcgrath should we keep taking into account the ID Token expiration to determine the validity of the credentials?

adamjmcgrath: Yep 👍  - we should only use the ID Token `exp` when validating the ID Token

Widcket: Ok, I'll remove that check then

adamjmcgrath (approved):

PR #540, opened 2021-11-09
Use `openid profile email` as the default scope value [SDK-2926]

Widcket:
### Changes

The library was using an assortment of different values as the default value of the `scope` parameter in Authentication client methods and Web Auth. The default scope value used across the library was changed to `openid profile email`, with the exception of the `renew(withRefreshToken:scope:)` method, in which it remains `nil`.

As the `scope` parameter in Authentication client methods have a default value, it is not necessary for that parameter to be optional. So it was changed to be non-optional.

Finally, in a couple of Authentication client methods, the `scope` and `audience` parameters switched places for the sake of consistency with the rest of the methods.

### Testing

### Checklist


Auth0/Auth0Authentication.swift
```diff
@@ -19,11 +19,11 @@ struct Auth0Authentication: Authentication {
         self.telemetry = telemetry
     }
 
-    func login(email username: String, code otp: String, audience: String?, scope: String?) -> Request<Credentials, AuthenticationError> {
```
Widcket, line 22: With a default value, there's no point in having the `scope` parameter be optional.

adamjmcgrath (approved):

PR #544, opened 2021-11-11
Allow relative urls

todd-toast:
### Changes

We need our auth0 requests to hit an url that is prefixed by a specific path. For example: `https://samples.auth0.com/authentication/`

Any requests made by the auth0 would need to append to this path. For example: `https://samples.auth0.com/authentication/oauth/token`

Prior to this change any base path would get ignored when making a request. We have confirmed that the Android SDK already handles this properly.

The main issue is the:

    URL(string: path, relativeTo: self.url)

where path always contained a leading "/". This caused the path that was already on the url object to get ignored.

The main function that was added was the constructURL function.

```swift
    private func constructURL(for path: String) -> URL {
        let relativeURLPrefix = relativeToBaseURL && !path.hasPrefix("/") ? "/": ""
        return URL(string: relativeURLPrefix + path, relativeTo: self.url)!
    }
```

In case of any backwards compatibility issues, I opted to create this function and configuration option `relativeToBaseURL` in order to not break code when updating.

This PR is a suggestion and I put it up because we really need this ability to provide a baseURL, but fully understand that it might not be in line with api that Auth0 would like.

### References

- support ticket
- community post
- StackOverflow post
- support forum thread

### Testing

* [ ] This change has been tested on the latest version of the platform/language or why no

I have tested this with our project and confirmed that it solves the issue.

### Checklist

### DIFF
--- a/Auth0/Auth0.swift
+++ b/Auth0/Auth0.swift
@@ -35,8 +35,8 @@ import Foundation
 
  - returns: Auth0 Authentication API
  */
-public func authentication(clientId: String, domain: String, session: URLSession = .shared) -> Authentication {
-    return Auth0Authentication(clientId: clientId, url: .a0_url(domain), session: session)
+public func authentication(clientId: String, domain: String, relativeToBaseURL: Bool = true, session: URLSession = .shared) -> Authentication {
+    return Auth0Authentication(clientId: clientId, url: .a0_url(domain), relativeToBaseURL: relativeToBaseURL, session: session)
 }
 
 /**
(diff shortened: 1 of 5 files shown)


chanceattoast (approved):

Widcket:
Closing as we'll be tackling this on the new major (as per https://github.com/auth0/Auth0.swift/issues/545#issuecomment-971132014).

Widcket:
👋🏼  @todd-toast we're planning to release a beta of this new major version around the third week of January (tentative ETA). That beta will include support for subpaths in the Auth0 domain. It'd be great to have your feedback on it. Please let us know if you'd be interested on trying it out.

todd-toast:
Hi @Widcket. We're definitely interested in trying it out and providing feedback when the time comes. Thank you!

PR #547, opened 2021-11-12
Migrate existing Objective-C code to Swift [SDK-2905]

adamjmcgrath:
### Changes

Migrate the following files to Swift:

- `A0ChallengeGenerator.m` (moved to `ChallengeGenerator.swift`)
- `A0RSA.m` (moved into `JWTAlgorithm.swift`)
- `A0SHA.m` (deleted)

Removed the references to Objective-C code from the `Podspec`, and the `Objective-C` target from the `Package.swift` file

### Testing

- Used the existing tests (added one for the length of a challenge verifier)
- Manually tested on the Dummy App

### Checklist


adamjmcgrath:
@Widcket fi added 658c32c

- remove `h, m` from podspec
- delete `Auth0/Auth0.h`
- Remove `cSettings` from `Package.swift`
- `ChallengeGenerator.swift` made floating funcs private
- remove `GCC_PREPROCESSOR_DEFINITIONS` from the podspec and  pbxproj

Auth0/JWTAlgorithm.swift
```diff
@@ -22,10 +18,16 @@ enum JWTAlgorithm: String {
             !signature.isEmpty else { return false }
         switch self {
         case .rs256:
-            guard let publicKey = jwk.rsaPublicKey, let rsa = A0RSA(key: publicKey) else { return false }
-            let sha256 = A0SHA()
-            return rsa.verify(sha256.hash(data), signature: signature)
+            var verificationError: Unmanaged<CFError>?
+            guard
+                SecKeyVerifySignature(
```
Widcket, line 23: No need for the `guard` clause, just return the call.

Auth0/JWTAlgorithm.swift
```diff
@@ -22,10 +18,16 @@ enum JWTAlgorithm: String {
             !signature.isEmpty else { return false }
         switch self {
         case .rs256:
-            guard let publicKey = jwk.rsaPublicKey, let rsa = A0RSA(key: publicKey) else { return false }
-            let sha256 = A0SHA()
-            return rsa.verify(sha256.hash(data), signature: signature)
+            var verificationError: Unmanaged<CFError>?
```
Widcket, line 21: No need for passing an error instance, if we're not going to use it. Just pass `nil` to the respective parameter of `SecKeyVerifySignature` and remove the error variable.

Auth0Tests/Generators.swift
```diff
@@ -121,10 +121,14 @@ func generateJWT(alg: String = JWTAlgorithm.rs256.rawValue,
     
     if let signature = signature {
         signaturePart = signature
+    } else {
+        var signingError: Unmanaged<CFError>?
```
Widcket, line 125: Same as above, remove the error variable if it is not going to be used and just pass `nil` instead.

Widcket (changes requested):
Good work! Just a couple of comments about `CFError`s.

Auth0/JWTAlgorithm.swift
```diff
@@ -1,9 +1,5 @@
-#if WEB_AUTH_PLATFORM
```
Widcket, line 1: Why remove `WEB_AUTH_PLATFORM`? This is only used in macOS and iOS.

adamjmcgrath: huh, good point, I think I thought it was obj c related and so axed it - I'll put it back.

Also, should I add the same condition around `ChallengeGenerator.swift`?

Widcket: Yes, it should have the same condition. It prevents it being compiled on watchOS and tvOS.

Widcket (approved):
LGTM

PR #548, opened 2021-11-13
Make credentials renewal thread-safe [SDK-2903]

Widcket:
### Changes

Currently, when Refresh Token Rotation is enabled, a race condition can occur when calling the `credentials(withScope:minTTL:parameters:callback:)` method of the `CredentialsManager` class repeatedly on the same thread, or concurrently form different threads. This will cause more than one renewal request to be fired, but only the first request will succeed as the old refresh token will get invalidated.

This PR uses a `DispatchGroup` along with a queue to make the credentials renewal logic run serially instead of concurrently, while also ensuring that:
- If a renewal request succeeds, the next requests will not be fired and instead the new credentials will be returned.
- If a renewal request fails, the next request will be fired.

### Testing

### Checklist

### DIFF
--- a/Auth0/CredentialsManager.swift
+++ b/Auth0/CredentialsManager.swift
@@ -11,7 +11,7 @@ public struct CredentialsManager {
     private let storage: A0SimpleKeychain
     private let storeKey: String
     private let authentication: Authentication
-    private let dispatchQueue = DispatchQueue(label: "com.auth0.credentialsmanager.queue", attributes: .concurrent)
+    private let dispatchQueue = DispatchQueue(label: "com.auth0.credentialsmanager.serial")
     private let dispatchGroup = DispatchGroup()
     #if WEB_AUTH_PLATFORM
     private var bioAuth: BioAuthentication?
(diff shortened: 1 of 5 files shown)


Auth0/CredentialsManager.swift
```diff
@@ -158,40 +157,60 @@ public struct CredentialsManager {
...
-                        callback(nil, newCredentials)
                     }
-                case .failure(let error):
-                    callback(.failedRefresh(error), nil)
                 }
-            }
+        }
+
+        if DispatchQueue.getSpecific(key: queueKey) != queueLabel {
+            self.dispatchGroup.wait()
```
Widcket, line 212: This blocks the queue until `dispatchGroup.leave()` is called, and thus making the whole operation run serially.

Widcket, line 212: This blocks until `dispatchGroup.leave()` is called, and thus makes the whole operation run serially.

adamjmcgrath (approved):

PR #549, opened 2021-11-17
Replace `JSONObjectPayload` with `Codable` on `Credentials` [SDK-2958]

Widcket:
### Changes

This PR replaces the conformance of `Credentials` to `JSONObjectPayload` with conformance to `Codable` and uses the newer, non-deprecated methods of `NSKeyedArchiver` and `NSKeyedUnarchiver` to archive and unarchive `Credentials`.

Additionally, the `Credentials` class was made `final`.

Care was taken to ensure that credentials archived with Auth0.swift 1.x can be unarchived successfully, and the other way around (in case customers need to downgrade to 1.x). Otherwise, if apps are using the Credentials Manager of Auth0.swift 1.x and then get upgraded to 2.x, the stored credentials will not be readable, essentially logging out all the users.
To maintain compatibility, `@objc(A0Credentials)` was added back to `Credentials`, which still needs to inherit from `NSObject`.

Dropping `NSSecureCoding` in favor of `Codable` + the `Codable`-supporting methods of `NSKeyedArchiver` and `NSKeyedUnarchiver` has been ruled out for the sake of this compatibility after doing a spike testing it.

### Testing

The backward compatibility of the archived credentials was tested manually in a test app, by:
- Storing credentials with the Credentials Manager from Auth0.swift 1.38.0 and then reading them back using the Credentials Manager from this PR.
- Storing credentials with the Credentials Manager from this PR and then reading them back using the Credentials Manager from Auth0.swift 1.38.0.

These tests were performed on both iOS 15 (simulator) and macOS Big Sur.

### Checklist

### DIFF
--- a/Auth0/Credentials.swift
+++ b/Auth0/Credentials.swift
@@ -3,7 +3,7 @@ import Foundation
 /**
  User's credentials obtained from Auth0.
  */
-public final class Credentials: NSObject, NSSecureCoding {
+public class Credentials: NSObject, NSSecureCoding {
 
     /// Token used that allows calling to the requested APIs (audience sent on Auth)
     public let accessToken: String
--- a/Auth0/Credentials.swift
+++ b/Auth0/Credentials.swift
@@ -35,6 +35,7 @@ public final class Credentials: NSObject, NSSecureCoding {
         self.scope = scope
         self.recoveryCode = recoveryCode
     }
+
     // MARK: - NSSecureCoding
 
     convenience public init?(coder aDecoder: NSCoder) {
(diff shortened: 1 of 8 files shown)


Auth0/Credentials.swift
```diff
@@ -3,7 +3,8 @@ import Foundation
 /**
  User's credentials obtained from Auth0.
  */
-public class Credentials: NSObject, JSONObjectPayload, NSSecureCoding {
+@objc(A0Credentials)
```
Widcket, line 6: Re-added for backward compatibility of archived credentials.

adamjmcgrath (approved):

PR #551, opened 2021-11-18
Decouple Credentials Manager storage from SimpleKeychain [SDK-2936]

adamjmcgrath:
### Changes

Create a protocol that contains the storage methods used by the Credentials Manager and add an extension to A0SimpleKeychain to have it implement that protocol. Then ensure the storage parameter has that protocol as its type instead of A0SimpleKeychain.

Wasn't sure where to put the protocol and extension...

### Testing

### Checklist

### DIFF
--- a/Auth0/CredentialsManager.swift
+++ b/Auth0/CredentialsManager.swift
@@ -7,7 +7,7 @@ import LocalAuthentication
 
 /// Generic storage API for storing credentials
 public protocol CredentialsStorage {
-    /// Retreive a storage entry
+    /// Retrieve a storage entry
     ///
     /// - Parameters:
     ///   - forKey: The key to get from the store
--- a/Auth0/CredentialsManager.swift
+++ b/Auth0/CredentialsManager.swift
@@ -12,14 +12,14 @@ public protocol CredentialsStorage {
     /// - Parameters:
     ///   - forKey: The key to get from the store
     /// - Returns: The stored data
-    func data(forKey: String) -> Data?
+    func getEntry(forKey: String) -> Data?
     /// Set a storage entry
     ///
     /// - Parameters:
     ///   - _: The data to be stored
     ///   - forKey: The key to store it to
     /// - Returns: if credentials were stored
-    func setData(_: Data, forKey: String) -> Bool
+    func setEntry(_: Data, forKey: String) -> Bool
     /// Delete a storage entry
     ///
     /// - Parameters:
@@ -29,6 +29,12 @@ public protocol CredentialsStorage {
 }
 
 extension A0SimpleKeychain: CredentialsStorage {
+    public func getEntry(forKey: String) -> Data? {
+        return data(forKey: forKey)
+    }
+    public func setEntry(_ data: Data, forKey: String) -> Bool {
+        return setData(data, forKey: forKey)
+    }
 }
 
 /// Credentials management utility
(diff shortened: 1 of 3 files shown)


V2_MIGRATION_GUIDE.md
```diff
@@ -202,6 +202,31 @@ In the following methods the `scope` parameter became non-optional (with a defau
 
 The `multifactorChallenge(mfaToken:types:authenticatorId:)` method lost its `channel` parameter, which is no longer necessary.
 
+### Credentials Manager
+
+`CredentialsManager` now takes a `CredentialsStorage` protocol as it's storage argument rather than an instance of `SimpleKeyChain`.
```
Widcket, line 207:
```suggestion
`CredentialsManager` now takes a `CredentialsStorage` protocol as it's storage argument rather than an instance of `SimpleKeychain`.
```

Auth0/CredentialsManager.swift
```diff
@@ -5,10 +5,36 @@ import JWTDecode
...
 #endif
 
+/// Generic storage API for storing credentials
+public protocol CredentialsStorage {
+    /// Retrieve a storage entry
+    ///
+    /// - Parameters:
+    ///   - forKey: The key to get from the store
+    /// - Returns: The stored data
+    func data(forKey: String) -> Data?
```
Widcket, line 15: Following the [API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/), the methods can be renamed as follows:

```swift
func getEntry(forKey: String) -> Data?
func setEntry(_: Data, forKey: String) -> Bool
func deleteEntry(forKey: String) -> Bool
```

Usage:

```swift
guard let data = storage.getEntry(forKey: z) else { return }  // get entry for key z
guard storage.setEntry(y, forKey: z) else { return } //  set entry y for key z
guard storage.deleteEntry(forKey: z) else { return }  // delete entry for key z
```

> Include all the words needed to avoid ambiguity for a person reading code where the name is used.

> Compensate for weak type information to clarify a parameter’s role.
Especially when a parameter type is NSObject, Any, AnyObject, or a fundamental type such Int or String, type information and context at the point of use may not fully convey intent.
To restore clarity, precede each weakly typed parameter with a noun describing its role

> Prefer method and function names that make use sites form grammatical English phrases.

> Name functions and methods according to their side-effects
Those with side-effects should read as imperative verb phrases

That will require wrapping the SimpleKeychain methods in the extension.

Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -59,7 +59,44 @@ class CredentialsManagerSpec: QuickSpec {
                 expect(credentialsManager.clear()).to(beFalse())
             }
         }
-        
+
+        describe("custom storage") {
+
+            class CustomStore: CredentialsStorage {
+                var store: [String : Data] = [:]
```
Widcket, line 66:
```suggestion
                var store: [String: Data] = [:]
```

Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -59,7 +59,44 @@ class CredentialsManagerSpec: QuickSpec {
                 expect(credentialsManager.clear()).to(beFalse())
             }
         }
-        
+
+        describe("custom storage") {
+
+            class CustomStore: CredentialsStorage {
```
Widcket, line 65: Could you please move the class to the bottom of the file?

Widcket (changes requested):
The protocol methods need some changes to make sure they follow the API Guidelines.

Widcket (approved):
LGTM

Auth0/CredentialsManager.swift
```diff
@@ -5,10 +5,42 @@ import JWTDecode
 import LocalAuthentication
 #endif
 
+/// Generic storage API for storing credentials
+public protocol CredentialsStorage {
```
Widcket, line 9: It's fine for it to be on the same file as the Credentials Manager

PR #552, opened 2021-11-19
Migrate all errors to structs [SDK-2968]

Widcket:
### Changes

In Auth0.swift 1.x `WebAuthError` and `CredentialsManagerError`  are enums, and as such the consumers must switch exhaustively over all the cases. This means that adding new cases constitutes a breaking change. So this was the source of many `TODO:` comments, as new error cases needed to be added.

This PR migrates all errors to be structs, and in the case of `WebAuthError` and `CredentialsManagerError` in particular, all the former cases are now static properties, meaning that they can be switched over but that switch can never be exhaustive. So adding new error cases will not be a breaking change.

Additionally, the error handling DX was improved, as now all errors conform to `Equatable` and `CustomDebugStringConvertible`, and all errors are being properly wrapped. Before, Web Auth could produce either a `WebAuthError` or an `AuthenticationError`; now it will only produce a `WebAuthError`. Same for the Credentials Manager, which will only produce a `CredentialsManagerError`.

### Testing

### Checklist


V2_MIGRATION_GUIDE.md
```diff
@@ -119,22 +119,136 @@ The `useUniversalLink()` method was removed as well, as Universal Links [cannot
...
+}
+```
+
+##### After
+
+```swift
+switch error {
+    case .userCancelled: // handle error
+    // ...
+    default: // handle unkwown errors, e.g. errors added in future versions
```
adamjmcgrath, line 194:
```suggestion
    default: // handle unknown errors, e.g. errors added in future versions
```

adamjmcgrath (approved):

Issue #545, opened 2021-11-11
Allow prefixed paths


todd-toast:
**Please do not report security vulnerabilities here**. The [Responsible Disclosure Program](https://auth0.com/responsible-disclosure-policy) details the procedure for disclosing security issues.

**Thank you in advance for helping us to improve this library!** Your attention to detail here is greatly appreciated and will help us respond as quickly as possible. For general support or usage questions, use the [Auth0 Community](https://community.auth0.com/) or [Auth0 Support](https://support.auth0.com/). Finally, to avoid duplicates, please search existing Issues before submitting one here.

By submitting an Issue to this repository, you agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Describe the problem you'd like to have solved

We need our auth0 requests to hit an url that is prefixed by a specific path. For example: https://samples.auth0.com/authentication/

Any requests made by the auth0 would need to append to this path. For example: https://samples.auth0.com/authentication/oauth/token

Currently any base/prefixed path would get ignored when making a request. We have confirmed that the Android SDK already handles this properly.

I wasn't sure whether to file this as a bug or a feature request. In a way it could be both.

### Describe the ideal solution

The ideal solution would be to drop the leading `/` in URL(string, relativeTo)
For example on line 121:
```swift
let url = URL(string: "/oauth/token", relativeTo: self.url)!
```
The string part has a leading slash. Changing this to:
```swift
let url = URL(string: "oauth/token", relativeTo: self.url)!
```
Would fix this issue. However, I know that change would have the potential of breaking someone's current implementation of the library so another solution could be providing an additional option to allow a prefixed path.

## Alternatives and current work-arounds

I have made a proposal here #544. Another option could be passing in in a `URLConstructor` protocol. Then you could have a default one but you could also pass in a custom one. Another option could be providing the path explicitly as another parameter in the constructor.

### Additional context
From what I've found so far, there is no way around this at the moment. We are unable to provide a prefixed path until this is fixed or added.

Widcket:
Hi @todd-toast, thanks for reporting this.
I was able to reproduce the issue. As we're working on a new major version of Auth0.swift right now, we'll be tackling this issue there. So I'll be closing this issue (and the PR) as this was just added to our backlog for the new major.

Widcket:
[...]

PR #557, opened 2021-11-29
Add support for subpaths in Auth0 domain [SDK-2963]

Widcket:
### Changes

This PR adds support for subpaths in the Auth0 domain, by removing the leading slash from all the paths and adding a trailing slash to the base URL.

⚠️ **THIS PR CONTAINS BREAKING CHANGES**: an URL extension has been made internal (it was public), as it has no use for developers.

### References

Fixes https://github.com/auth0/Auth0.swift/issues/545

### Testing

Unit tests have been added, and the changes were also tested manually (for Web Auth only) with an iOS 15.0 emulator (iPhone 13) and macOS 11.6.1, using the respective playground apps.

Screen Shot 2021-11-29 at 18 43 21

Screen Shot 2021-11-29 at 18 04 08

### Checklist


Auth0/NSURL+Auth0.swift
```diff
@@ -1,20 +1,17 @@
 import Foundation
 
-public extension URL {
```
Widcket, line 3: This is the breaking change.

Auth0Tests/ClaimValidatorsSpec.swift
```diff
@@ -66,7 +66,7 @@ class ClaimValidatorsSpec: IDTokenValidatorBaseSpec {
         describe("iss validation") {
             
             var issValidator: IDTokenIssValidator!
-            let expectedIss = "\(URL.a0_url(domain).absoluteString)/"
+            let expectedIss = "\(URL.httpsURL(from: domain).absoluteString)/"
```
adamjmcgrath, line 69: Not that it's breaking any of your tests, but this will end in `//`

adamjmcgrath (approved):

Auth0Tests/ClaimValidatorsSpec.swift
Widcket: Good catch, I'll look into it.

Widcket: Indeed.

Screen Shot 2021-11-30 at 13 47 34

It does not break any tests because that constant is only used for checking presence and non-nil.

Widcket: I've removed the extra `/` in https://github.com/auth0/Auth0.swift/pull/557/commits/68be1ea56fb24dc4f01c7b8219dd5bccc941a568.

adamjmcgrath (approved):

PR #562, opened 2021-12-01
Add a way to pass custom headers directly to the request [SDK-2908]

adamjmcgrath:
### Changes

Add the `headers(_: [String: String])` method to the Request type, which allows the developer to pass arbitrary HTTP headers to a given request.

```swift
Auth0
    .authentication()
    .renew(withRefreshToken: refreshToken)
    .headers(["key": "value"])
    .start { result in
        switch result {
        case .success(let credentials): dump(credentials)
        case .failure(let error): print(error)
        }
    }
```

added a `headers: [String: String] = [:]` parameter to the `credentials(withScope:minTTL:parameters:callback:)` and the `revoke(:_)` public methods of the Credentials Manager which use the new `headers(_: [String: String])` method to add the headers to the Request.

### Testing

* [x] This change has been tested on the latest version of the platform/language

### Checklist

### DIFF
--- a/Auth0.xcodeproj/project.pbxproj
+++ b/Auth0.xcodeproj/project.pbxproj
@@ -289,6 +289,9 @@
 		970BC36C25C27095007A7745 /* Challenge.swift in Sources */ = {isa = PBXBuildFile; fileRef = 970BC36A25C27095007A7745 /* Challenge.swift */; };
 		970BC36D25C27095007A7745 /* Challenge.swift in Sources */ = {isa = PBXBuildFile; fileRef = 970BC36A25C27095007A7745 /* Challenge.swift */; };
 		970BC36E25C27095007A7745 /* Challenge.swift in Sources */ = {isa = PBXBuildFile; fileRef = 970BC36A25C27095007A7745 /* Challenge.swift */; };
+		D581CF772757D773007327D1 /* RequestSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = D581CF762757D773007327D1 /* RequestSpec.swift */; };
+		D581CF782757D773007327D1 /* RequestSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = D581CF762757D773007327D1 /* RequestSpec.swift */; };
+		D581CF792757D773007327D1 /* RequestSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = D581CF762757D773007327D1 /* RequestSpec.swift */; };
 		D5E9E317273ACCA5000CDB0A /* ChallengeGenerator.swift in Sources */ = {isa = PBXBuildFile; fileRef = D5E9E316273ACCA5000CDB0A /* ChallengeGenerator.swift */; };
 		D5E9E318273ACCA5000CDB0A /* ChallengeGenerator.swift in Sources */ = {isa = PBXBuildFile; fileRef = D5E9E316273ACCA5000CDB0A /* ChallengeGenerator.swift */; };
 /* End PBXBuildFile section */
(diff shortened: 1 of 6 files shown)


Auth0.xcodeproj/project.pbxproj
```diff
@@ -289,6 +289,9 @@
 		970BC36C25C27095007A7745 /* Challenge.swift in Sources */ = {isa = PBXBuildFile; fileRef = 970BC36A25C27095007A7745 /* Challenge.swift */; };
 		970BC36D25C27095007A7745 /* Challenge.swift in Sources */ = {isa = PBXBuildFile; fileRef = 970BC36A25C27095007A7745 /* Challenge.swift */; };
 		970BC36E25C27095007A7745 /* Challenge.swift in Sources */ = {isa = PBXBuildFile; fileRef = 970BC36A25C27095007A7745 /* Challenge.swift */; };
+		D581CF772757D773007327D1 /* RequestSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = D581CF762757D773007327D1 /* RequestSpec.swift */; };
```
adamjmcgrath, line 292: Not sure why there's 3 entries, I added and removed the spec file a couple of times so it could be that

Widcket: There are 3 entries because the file belongs to 3 targets:

Screen Shot 2021-12-01 at 16 23 44

Auth0/CredentialsManager.swift
```diff
@@ -113,7 +113,8 @@ public struct CredentialsManager {
     /// If no refresh token is available the endpoint is not called, the credentials are cleared, and the callback is invoked without an error.
     ///
     /// - Parameter callback: callback with an error if the refresh token could not be revoked
```
Widcket, line 115:
```suggestion
    /// - Parameters:
    /// - callback: callback with an error if the refresh token could not be revoked
```

Auth0/CredentialsManager.swift
```diff
@@ -196,7 +199,7 @@ public struct CredentialsManager {
         return credentials
     }
 
-    private func retrieveCredentials(withScope scope: String?, minTTL: Int, parameters: [String: Any] = [:], callback: @escaping (CredentialsManagerResult<Credentials>) -> Void) {
+    private func retrieveCredentials(withScope scope: String?, minTTL: Int, parameters: [String: Any] = [:], headers: [String: String] = [:], callback: @escaping (CredentialsManagerResult<Credentials>) -> Void) {
```
Widcket, line 202:
```suggestion
    private func retrieveCredentials(withScope scope: String?, minTTL: Int, parameters: [String: Any], headers: [String: String], callback: @escaping (CredentialsManagerResult<Credentials>) -> Void) {
```
There's no need for default values in a private method that will receive values from other methods that also have default values.

Auth0/CredentialsManager.swift
```diff
@@ -113,7 +113,8 @@ public struct CredentialsManager {
     /// If no refresh token is available the endpoint is not called, the credentials are cleared, and the callback is invoked without an error.
     ///
     /// - Parameter callback: callback with an error if the refresh token could not be revoked
-    public func revoke(_ callback: @escaping (CredentialsManagerError?) -> Void) {
+    ///   - headers: additional headers to add to a possible token revokation. The headers will be set via Request.headers.
+    public func revoke(_ callback: @escaping (CredentialsManagerError?) -> Void, headers: [String: String] = [:]) {
```
Widcket, line 117: The callback should be last.

Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -162,6 +162,17 @@ class CredentialsManagerSpec: QuickSpec {
                     }
                 }
             }
+            
+            it("should include custom headers") {
+                stub(condition: hasHeader("foo", value: "bar")) { _ in return revokeTokenResponse() }.name = "revoke success"
+                waitUntil(timeout: Timeout) { done in
+                    credentialsManager.revoke({ error in
+                        expect(error).to(beNil())
+                        expect(credentialsManager.hasValid()).to(beFalse())
```
Widcket, line 171: Is this assert necessary for this test case?

Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -484,6 +495,20 @@ class CredentialsManagerSpec: QuickSpec {
...
+
+                it("should include custom headers") {
+                    stub(condition: hasHeader("foo", value: "bar")) {
+                        _ in return authResponse(accessToken: NewAccessToken, idToken: NewIdToken, refreshToken: NewRefreshToken, expiresIn: ExpiresIn)
+                    }
+                    credentials = Credentials(accessToken: AccessToken, tokenType: TokenType, idToken: IdToken, refreshToken: RefreshToken, expiresIn: Date(timeIntervalSinceNow: -ExpiresIn))
+                    _ = credentialsManager.store(credentials: credentials)
+                    waitUntil(timeout: Timeout) { done in
+                        credentialsManager.credentials(headers: ["foo": "bar"]) { result in
+                            expect(result).to(haveCredentials(NewAccessToken, NewIdToken, NewRefreshToken))
```
Widcket, line 507: Same as above, this probably should be only asserting that the result was successful.

Widcket (changes requested):
Looking good, left a few comments.

Auth0Tests/CredentialsManagerSpec.swift
adamjmcgrath: not really, will remove 👍

Widcket (approved):
LGTM

PR #565, opened 2021-12-02
Do not use the message from the underlying error

Widcket:
### Changes

This PR changes the error message logic for `CredentialsManagerError` and `AuthenticationError` so that it will not automatically return the error message from the underlying error if there is one, because the developer is expected to handle the underlying error anyway, and having both the error and underlying error have the same message could be confusing.

### Testing

### Checklist


starman1398:
https://story.snapchat.com/o/W7_EDlXWTBiXAEEniNoMPwAAYcG9seWFmdmR2AXyEcd-JAXyEcdzBAAAAAA?share_id=QzNDM0Y1&locale=en_US

adamjmcgrath (approved):

PR #573, opened 2021-12-06
Remove parameterless `tokenExchange()` method

Widcket:
### Changes

Auth0.swift 1.x has a `tokenExchange(withParameters:)` method, besides `tokenExchange(withCode:codeVerifier:redirectURI:)`. Since we added `Request.parameters(_:)` and removed all the existing `parameters:` parameters, `tokenExchange(withParameters:)` ended up as `tokenExchange()`.
 
Auth0.Android does not have a similar parameterless method, [only](https://github.com/auth0/Auth0.Android/blob/main/auth0/src/main/java/com/auth0/android/authentication/AuthenticationAPIClient.kt#L679) `fun token(authorizationCode: String, codeVerifier: String, redirectUri: String )`, that is, the equivalent to `tokenExchange(withCode:codeVerifier:redirectURI:)`.

Since we already have `tokenExchange(withCode:codeVerifier:redirectURI:)`, and given the fact developers can add custom parameters with `Request.parameters(_:)`, it does not make sense to keep the parameterless `tokenExchange()` method. This PR removes it.

### Testing

### Checklist

### DIFF
--- a/Auth0/Auth0Authentication.swift
+++ b/Auth0/Auth0Authentication.swift
@@ -270,20 +270,6 @@ struct Auth0Authentication: Authentication {
                        telemetry: self.telemetry)
     }
 
-    func tokenExchange() -> Request<Credentials, AuthenticationError> {
-        let payload: [String: Any] = [
-            "client_id": self.clientId
-        ]
-        let token = URL(string: "oauth/token", relativeTo: self.url)!
-        return Request(session: session,
-                       url: token,
-                       method: "POST",
-                       handle: codable,
-                       parameters: payload,
-                       logger: self.logger,
-                       telemetry: self.telemetry)
-    }
-
     func tokenExchange(withCode code: String, codeVerifier: String, redirectURI: String) -> Request<Credentials, AuthenticationError> {
         return self.tokenExchange().parameters([
             "code": code,
@@ -340,6 +326,7 @@ struct Auth0Authentication: Authentication {
 // MARK: - Private Methods
 
 private extension Auth0Authentication {
+
     func login(username: String, otp: String, realm: String, audience: String?, scope: String) -> Request<Credentials, AuthenticationError> {
         let url = URL(string: "oauth/token", relativeTo: self.url)!
         var payload: [String: Any] = [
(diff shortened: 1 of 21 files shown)


Auth0/Auth0Authentication.swift
```diff
@@ -360,6 +347,20 @@ private extension Auth0Authentication {
                        telemetry: self.telemetry)
     }
 
+    func tokenExchange() -> Request<Credentials, AuthenticationError> {
```
adamjmcgrath, line 350: Not sure if you want to change these, but strictly speaking there's no "Token Exchange" going on with these defaults, you're just calling the "Get Token" endpoint https://auth0.com/docs/api/authentication#dynamic-application-client-registration

V2_MIGRATION_GUIDE.md
```diff
@@ -97,6 +97,20 @@ Use `createUser(email:username:password:connection:userMetadata:rootAttributes:`
 
 Use `userInfo(withAccessToken:)` instead.
 
+#### `tokenExchange(withParameters:)`
+
+Use `tokenExchange(withCode:codeVerifier:redirectURI:)` instead. To pass custom parameters, use the `parameters(_:)` method from `Request`:
```
adamjmcgrath, line 102: Similarly, this should really be called `codeExchange` - since you're doing Proof Key for *Code Exchange* (not sure what Android calls it)

Auth0/Auth0Authentication.swift
Widcket: Renamed the private method to just `token()` after the endpoint in https://github.com/auth0/Auth0.swift/pull/573/commits/8f5a6561e3739ce4f122668f5ef2135481789ce4.

V2_MIGRATION_GUIDE.md
Widcket: Good catch, renamed the method in https://github.com/auth0/Auth0.swift/pull/573/commits/8f5a6561e3739ce4f122668f5ef2135481789ce4.
Android calls it `token()`.

adamjmcgrath (approved):

PR #530, opened 2021-11-02
Remove `parameters` from Authentication API client methods [SDK-2902]

Widcket:
### Changes

⚠️ **THIS PR CONTAINS BREAKING CHANGES**

The following Authentication API Client methods lost the `parameters` parameter:
- `login(phoneNumber:code:audience:scope:)`
- `login(usernameOrEmail:password:realm:audience:scope:)`
- `loginDefaultDirectory(withUsername:password:audience:scope:)`
- `tokenExchange()`

That parameter was no longer necessary since `Request` has a `parameters()` [method](https://github.com/auth0/Auth0.swift/blob/master/Auth0/Request.swift#L103) that can be used to pass custom parameters to any Authentication API Client method.

### Testing

* [ ] This change adds unit test coverage
* [ ] This change has been tested on the latest version of the platform/language or why not

### Checklist

* [ ] I have read the [Auth0 general contribution guidelines](https://github.com/auth0/open-source-template/blob/master/GENERAL-CONTRIBUTING.md)
* [ ] All existing and new tests complete without errors
* [ ] All active GitHub checks have passed

PR #607, opened 2021-12-21
Remove remaining `parameters` parameter

Widcket:
### Changes

In the PR that removed the `parameters` parameter across the SDK, the one in the method `startPasswordless(email:type:connection:)` from the Authentication API client was skipped because that method had some conditional logic for sending those parameters. But since the `Request.parameters` method allows adding arbitrary parameters to any request (even that one), that logic does not hold anymore. So this PR removes the last remaining `parameters` parameter from the `startPasswordless(email:type:connection:)` method.

### References

See https://github.com/auth0/Auth0.swift/pull/530

### Testing

### Checklist

### DIFF
--- a/Auth0/Authentication.swift
+++ b/Auth0/Authentication.swift
@@ -246,24 +246,24 @@ public protocol Authentication: Trackable, Loggable {
          }
      ```
 
-     If you need to specify a scope:
+     You can also specify audience and scope:
 
      ```
      Auth0
          .authentication(clientId: clientId, domain: "samples.auth0.com")
          .login(appleAuthorizationCode: authCode,
                 fullName: credentials.fullName,
-                scope: "openid profile email offline_access",
-                audience: "https://myapi.com/api")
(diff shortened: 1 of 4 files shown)


Auth0/Authentication.swift
```diff
@@ -631,12 +631,8 @@ public extension Authentication {
         return self.signup(email: email, username: username, password: password, connection: connection, userMetadata: userMetadata, rootAttributes: rootAttributes)
     }
 
-    func signup(email: String, username: String? = nil, password: String, connection: String, userMetadata: [String: Any]? = nil) -> Request<DatabaseUser, AuthenticationError> {
```
Widcket, line 634: This overload is not necessary because there is one above that adds default values to all the optional params. This one was adding default values to all but one. But in Swift you can omit a parameter if it has a default value (as per the overload above). So this one is pointless.

adamjmcgrath (approved):

PR #657, opened 2022-02-15
Adds usage of ephemeralSession for logout session

sofacoder:
### Changes

This changes `AuthenticationServicesSessionCallback` to also allow to use an ephemeral session (if configured in the according `WebAuthenticatable`)

This helps for example to avoid the iOS pop up asking the user to allow accessing the auth0.com to sign in. Which no user understands during logout (for example).  

### References

There is no ticket or discussion yet, we just found out that this is an small change and tried it out.

### Testing

The change will executed during existing tests, but writing an explicit test for this is hard without changing the design (hence the used  `ASWebAuthenticationSession` is not exposed)

It can be tested in an demo app by doing for example the following (e.g. as a logout)

```
  .webAuth(clientId: AuthConfiguration.clientId, domain: AuthConfiguration.domain)
                        .useEphemeralSession()
                        .clearSession(federated: false) { success in
                         // ...
```

### Checklist

### DIFF
--- a/Auth0/AuthenticationServicesSessionCallback.swift
+++ b/Auth0/AuthenticationServicesSessionCallback.swift
@@ -26,7 +26,10 @@ import AuthenticationServices
 @available(iOS 12.0, macOS 10.15, *)
 final class AuthenticationServicesSessionCallback: SessionCallbackTransaction {
 
-    init(url: URL, schemeURL: URL, callback: @escaping (Bool) -> Void) {
+    init(url: URL,
+         schemeURL: URL,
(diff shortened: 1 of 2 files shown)


adamjmcgrath:
Hi @sofacoder - If you're using `useEphemeralSession()` to login, you do not need to call `clearSession()` to perform logout as there will be no cookies to remove. Just deleting the credentials will suffice.

adamjmcgrath:
Yep, there's no session to clear by `clearSession` if the session is ephemeral - closing, feel free to ping me if you want to continue the conversation

PR #701, opened 2022-06-03
CredentialManager: add hasValidRefreshToken() method

AndrewSB:
Version 2.0.0 of the SDK dropped checking for refresh tokens (https://github.com/auth0/Auth0.swift/pull/574)

Without this, there's no synchronous way to tell if the stored credential
can be refreshed, even if `isValid()` returns false.


Widcket:
Hi @AndrewSB , thank you for the report but we're missing some important information. Please re-open a new issue with a completed **PR template**. Delete sections that are not relevant, ask for help for sections that are not clear, and check all required checkboxes.

Please explain your use case for this feature in detail as well.

PR #708, opened 2022-06-23
Update README.md

adamjmcgrath:
Noticed this when I was doing something on Android

You can't use `x:users` scope on public clients, see https://auth0.com/docs/secure/tokens/access-tokens/get-management-api-tokens-for-single-page-applications#available-scopes-and-endpoints


Widcket (approved):
Thanks!

PR #746, opened 2022-10-28
Migrate API docs to use DocC

Widcket:

### 📋 Changes

This PR migrates all the API docs to use [Apple's DocC](https://developer.apple.com/documentation/docc). Previously, we used [Jazzy](https://github.com/realm/jazzy) to generate them. This involved:
- Updating the docs syntax to suit DocC
- Cleaning up and updating the API docs
- Setting up DocC
- Removing Jazzy


Auth0/AuthenticationError.swift
```diff
@@ -149,6 +137,8 @@ public struct AuthenticationError: Auth0APIError {
     /// - [timedOut](https://developer.apple.com/documentation/foundation/urlerror/2293002-timedout)
     /// - [internationalRoamingOff](https://developer.apple.com/documentation/foundation/urlerror/2292893-internationalroamingoff)
     /// - [callIsActive](https://developer.apple.com/documentation/foundation/urlerror/2293147-callisactive)
+    ///
+    /// The underlying `URLError` is available in the ``Auth0Error/cause-9wuyi`` property.
```
ewanharris, line 141: Is `9wuyi` expected? It occurs a few times throughout so I'm guessing so?

ewanharris (approved):
Changing review to approve so you're able to merge if the question in my comment isn't a problem

Auth0/AuthenticationError.swift
Widcket: Yes, it's a unique identifier to point to a specific `cause` property, because it's a protocol (interface) property with a default implementation. Without it, DocC doesn't know whether to link to the definition of the property, or to the default implementation.

Widcket: See https://developer.apple.com/documentation/xcode/formatting-your-documentation-content

Screen Shot 2022-10-28 at 12 40 51

ewanharris: Ah neat! Thank you for explaining!

PR #753, opened 2023-01-25
Added parameter for ability to have empty default parameters

furkankaynaryell:

### 📋 Changes
Auth0WebAuth class has this property: shouldStartWithDefaultParameters
Also in Auth0 class, where we initialise this class, we have this parameter, with the default value true, to not to break any current impleemntation.

### 🎯 Testing
Auth0WebAuth class has the method buildAuthorizeURL(). This method should contain the default parameters such as code_challenge, code_challenge_method, if shouldStartWithDefaultParameters set to true, or else, the URL must not contain those parameters.

### DIFF
--- a/Auth0/Auth0.swift
+++ b/Auth0/Auth0.swift
@@ -182,12 +182,13 @@ public func users(token: String, domain: String, session: URLSession = .shared) 
  - Parameters:
    - session: `URLSession` instance used for networking. Defaults to `URLSession.shared`.
    - bundle:  Bundle used to locate the `Auth0.plist` file. Defaults to `Bundle.main`.
+   - shouldStartWithDefaultParameters: The option for building the authorization URL without the default parameters.
  - Returns: Auth0 Web Auth client.
  - Warning: Calling this method without a valid `Auth0.plist` file will crash your application.
  */
-public func webAuth(session: URLSession = .shared, bundle: Bundle = Bundle.main) -> WebAuth {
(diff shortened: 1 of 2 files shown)


Widcket:
Hi @furkankaynaryell, thanks for raising this.

> code_challenge, code_challenge_method

These parameters cannot be changed because the SDK only supports the [authorization code flow with PKCE](https://auth0.com/docs/get-started/authentication-and-authorization-flow/authorization-code-flow-with-proof-key-for-code-exchange-pkce), and this will remain this way. Can you please tell us more about your use case?

furkankaynaryell:
> Hi @furkankaynaryell, thanks for raising this.
> 
> > code_challenge, code_challenge_method
> 
> These parameters cannot be changed because the SDK only supports the [authorization code flow with PKCE](https://auth0.com/docs/get-started/authentication-and-authorization-flow/authorization-code-flow-with-proof-key-for-code-exchange-pkce), and this will remain this way. Can you please tell us more about your use case?

Hi @Widcket, my case is that, I need the authorisation code to pass to an another SDK to initiate code flow. So I need to intervene just at the point when I call /authorize . Any possible ways to do it with the current implementation that you would suggest?

Widcket:
I'm afraid the SDK is not designed to do that. You would need to implement the Web Auth part yourself up to the point where you get the auth code –just a matter of opening the `/authorize` URL in the system browser, and handling the callback URL to extract the code. Note however that we don't recommend not using PKCE for logging in from mobile apps: https://auth0.com/blog/oauth-2-best-practices-for-native-apps/

PR #772, opened 2023-06-09
Add `renew()` to Credentials Manager [SDK-4300]

Widcket:

### 📋 Changes

This PR adds a new method `renew()` to the Credentials Manager, that allows to renew the stored credentials in a thread-safe way.

```swift
credentialsManager.renew { result in
    switch result {
    case .success(let credentials):
        print("Renewed credentials: \(credentials)")
    case .failure(let error):
        print("Failed with: \(error)")
    }
}
```

This new method also supports Combine and async/await.

### 🎯 Testing

Besides adding a slew of unit tests, this method was tested manually with Xcode 14.2 (14C18) using an iPhone simulator running iOS 16.2.

### DIFF
--- a/Auth0Tests/CredentialsManagerSpec.swift
+++ b/Auth0Tests/CredentialsManagerSpec.swift
@@ -528,17 +528,20 @@ class CredentialsManagerSpec: QuickSpec {
                         }
                     }
                 }
-                
+
                 it("should store new credentials") {
+                    let store = SimpleKeychain()
+                    credentialsManager = CredentialsManager(authentication: authentication, storage: store)
                     credentials = Credentials(accessToken: AccessToken, tokenType: TokenType, idToken: IdToken, refreshToken: RefreshToken, expiresIn: Date(timeIntervalSinceNow: -ExpiresIn))
                     _ = credentialsManager.store(credentials: credentials)
                     await waitUntil(timeout: Timeout) { done in
                         credentialsManager.credentials { result in
                             expect(result).to(beSuccessful())
-                            credentialsManager.credentials { result in
-                                expect(result).to(haveCredentials(NewAccessToken, NewIdToken, RefreshToken))
-                                done()
-                            }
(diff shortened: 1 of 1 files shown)


Auth0/CredentialsManager.swift
```diff
@@ -299,13 +300,55 @@ public struct CredentialsManager {
...
+    /// renewed credentials. Since this method is thread-safe and ``store(credentials:)`` is not, calling it anyway can
+    /// cause concurrency issues.
+    /// - Important: To ensure that no concurrent renewal requests get made, do not call this method from multiple
+    /// Credentials Manager instances. The Credentials Manager cannot synchronize requests across instances.
+    ///
+    /// ## See Also
+    ///
+    /// - [Refresh Tokens](https://auth0.com/docs/secure/tokens/refresh-tokens)
+    /// - [Authentication API Endpoint](https://auth0.com/docs/api/authentication#refresh-token)
+    public func renew(parameters: [String: Any] = [:], headers: [String: String] = [:], callback: @escaping (CredentialsManagerResult<Credentials>) -> Void) {
```
poovamraj, line 341: Should we pass scope here? Is it because we renew anyway if the scope is different?

poovamraj (approved):

Auth0/CredentialsManager.swift
Widcket: If a developer wants to downscope, we already have the `credentials()` method available, which will force-renew in that case. However, I'm not entirely convinced of the usefulness of this downscoping feature given that the CM does not support multiple sets of credentials, therefore it will overwrite the current credentials with the downscoped ones. So the AT with the full scopes will not be stored anymore. Something to consider for the future.

Widcket: And we already expose the auth API client method `renew()`, that supports downscoping as well (with no storage involved).

stevehobbsdev (approved):
LGTM!

PR #832, opened 2024-02-22
Add support for HTTPS callbacks [SDK-4749]

Widcket:

### 📋 Changes

This PR adds a new `useHTTPS()` method to `WebAuth`. It allows to use HTTPS redirect URLs when performing web-based authentication on iOS 17.4+ and macOS 14.4+:
- When this option is enabled and the platform version supports it, the SDK uses HTTPS.
- When this option is enabled but the platform version does not support it, the SDK falls back to using a custom scheme.
- When this option is not enabled, the SDK uses a custom scheme.

### 🎯 Testing

These changes have been tested manually with iOS 17.4 beta 3 and iOS 17.2, using Xcode 15.3 beta 3 (15E5194e).

### DIFF
--- a/Auth0/ASProvider.swift
+++ b/Auth0/ASProvider.swift
@@ -1,22 +1,29 @@
 #if WEB_AUTH_PLATFORM
 import AuthenticationServices
 
+typealias ASHandler = ASWebAuthenticationSession.CompletionHandler
+
 extension WebAuthentication {
 
-    static func asProvider(urlScheme: String, ephemeralSession: Bool = false) -> WebAuthProvider {
+    static func asProvider(redirectURL: URL, ephemeralSession: Bool = false) -> WebAuthProvider {
         return { url, callback in
-            let session = ASWebAuthenticationSession(url: url, callbackURLScheme: urlScheme) {
-                guard let callbackURL = $0, $1 == nil else {
-                    if let error = $1, case ASWebAuthenticationSessionError.canceledLogin = error {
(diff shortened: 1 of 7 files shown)


Auth0/ASProvider.swift
```diff
@@ -1,38 +1,68 @@
...
+        return {
+            guard let callbackURL = $0, $1 == nil else {
+                if let error = $1 as? NSError,
+                    error.userInfo.isEmpty,
+                    case ASWebAuthenticationSessionError.canceledLogin = error {
+                    return callback(.failure(WebAuthError(code: .userCancelled)))
+                } else if let error = $1 {
+                    return callback(.failure(WebAuthError(code: .other, cause: error)))
+                }
+
```
stevehobbsdev, line 51: Are these bogus code coverage problems? It seems to be complaining about a blank line not being covered.

stevehobbsdev (approved):
LGTM

Auth0/ASProvider.swift
Widcket: This is due to partial coverage in [this line](https://github.com/auth0/Auth0.swift/blob/2.5.0/Auth0/Request.swift#L57):

Screenshot 2024-02-29 at 11 06 46

Which isn't even relevant anymore, as with Xcode 15 `URLComponents` and `URL` now have the same parsing behavior: https://forums.swift.org/t/url-string-behavior-changed-with-xcode-15-0-beta-5/66570

PR #838, opened 2024-04-03
feat: self-bundle is a permitted resume callback

alexander-joinlane:

### 📋 Changes

When resuming the flow, some urls can be registered by the application (ex. www.my-app.com can be registered to open <My App>), but in some cases a redirect proxy must be used like in the case of B2B2C Auth0 consumers. Such a proxy could redirect directly into the app using the identifier syntax.

In accordance with "Getting Started" documentation which advises the use of "BUNDLE_IDENTIFIER.auth0://" schema urls we conditionally pass a direct and anticipated match.

More simply, if the outgoing redirect is a proxy, incoming redirects could go directly into the app.

### 📎 References

N/A

### 🎯 Testing

I was able to use xcode to test.

image


Auth0Tests/LoginTransactionSpec.swift
```diff
@@ -8,12 +8,16 @@ class LoginTransactionSpec: QuickSpec {
...
-        let loggerOutput = SpyOutput()
+        var userAgent: SpyUserAgent!
+        var handler: SpyGrant!
+        var loggerOutput: SpyOutput!
         let code = "123456"
 
         beforeEach {
+            userAgent = SpyUserAgent()
+            handler = SpyGrant()
+            loggerOutput = SpyOutput()
```
alexander-joinlane, line 19: note: reset mocks between test

Auth0Tests/LoginTransactionSpec.swift
```diff
@@ -50,6 +54,17 @@ class LoginTransactionSpec: QuickSpec {
                     expect(transaction.userAgent).to(beNil())
                 }
 
+                it("should succeed when using bundle schema") {
```
alexander-joinlane, line 57: note: roughly identical in implementation to "can handle"

desusai7:
Hi @alexander-joinlane, 

Thank you for raising this PR and bringing about this to our notice. We are tackling this as part of [#850 ](https://github.com/auth0/Auth0.swift/pull/850), hence closing this PR.

Issue #827, opened 2024-01-26
Auth0 crashes due to swift task continuation misuse


erichoracek:
### Checklist

- [X] The issue can be reproduced in the [Auth0.swift sample app](https://github.com/auth0-samples/auth0-ios-swift-sample/tree/master/Sample-01) (or N/A).
- [X] I have looked into the [Readme](https://github.com/auth0/Auth0.swift#readme), [Examples](https://github.com/auth0/Auth0.swift/blob/master/EXAMPLES.md), and [FAQ](https://github.com/auth0/Auth0.swift/blob/master/FAQ.md) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://auth0.github.io/Auth0.swift/documentation/auth0) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/Auth0.swift/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

We have seen two crashes in the Auth0 SDK with `SWIFT TASK CONTINUATION MISUSE`.

These crashes occur on the order of 1-20 times per day in our app in production, as reported by Sentry.

Here are the two relevant stack traces:

### Crash 1: Auth0WebAuth.swift:175
```
Exception Type: EXC_BREAKPOINT (SIGTRAP)
Crashed Thread: 0

Application Specific Information:
_Concurrency/CheckedContinuation.swift > Fatal error > SWIFT TASK CONTINUATION MISUSE: start() tried to resume its continuation more than once, throwing The user cancelled the Web Auth operation.!


Thread 0 Crashed:
0   libswiftCore.dylib              0x32e5923fc         _assertionFailure
1   libswift_Concurrency.dylib      0x3522f5a00         CheckedContinuation.resume
2   Redacted                        0x10490d628         [inlined] CheckedContinuation.resume<T>
3   Redacted                        0x10490d628         Auth0WebAuth.start (Auth0WebAuth.swift:306)
4   Redacted                        0x10490b05c         Auth0WebAuth.start (Auth0WebAuth.swift:175)
5   Redacted                        0x10490ffd4         thunk for closure
6   Redacted                        0x10490af78         thunk for closure
[...]

PR #848, opened 2024-05-02
fix: removed the possibility of concurrent webauth transactions to handle continuation misuse

desusai7:
### 📋 Changes

* Updated WebAuth to cancel any newly triggered WebAuth transactions when there exists an active WebAuth transaction already i.e ensuring that only single WebAuth transaction is active at any point of time.
* The above change would fix the issues in some special cases of race condition where continuations are misused by resuming twice.
* Updated unit tests for `login` & `logout` in WebAuthSpec to clear `TransactionStore` before each run to start clean and not get failed by the un-cleared transaction in memory by previous test runs.

### 📎 References

[auth0/Auth0.swift#827](https://github.com/auth0/Auth0.swift/issues/827)

### Checklist

### 🎯 Testing

Tested by trying to start the below defined two flows at exact same time via both callbacks & async/await approach and found that the code is working as per expectations **by cancelling the newly triggering web auth flow and ensuring the existing web auth flow executes completely.** 

1. WebAuth flow with `google-oauth2` connection using `ASWebAuthenticationSession`
2. WebAuth flow with `Universal Login` using `SafariViewController`


Auth0/WebAuthError.swift
```diff
@@ -79,6 +80,7 @@ extension WebAuthError {
         switch self.code {
         case .noBundleIdentifier: return "Unable to retrieve the bundle identifier from Bundle.main.bundleIdentifier,"
             + " or it could not be used to build a valid URL."
+        case .transactionActiveAlready: return "Failed to start this transaction, as there is an active transaction at the moment"
```
Widcket, line 83:
```suggestion
        case .transactionActiveAlready: return "Failed to start this transaction, as there is an active transaction at the moment."
```

Auth0/WebAuthError.swift
```diff
@@ -6,6 +6,7 @@ public struct WebAuthError: Auth0Error {
 
     enum Code: Equatable {
         case noBundleIdentifier
+        case transactionActiveAlready
```
Widcket, line 9: There are tests for `WebAuthError`, those should be updated as well: https://github.com/auth0/Auth0.swift/blob/master/Auth0Tests/WebAuthErrorSpec.swift

desusai7: Thanks for pointing it out, included tests for it now.

Auth0/WebAuthError.swift
desusai7: Updated the text description for the WebAuthError `.transactionActiveAlready` as per description

Auth0/WebAuthError.swift
Widcket: There's also a test for the error code: https://github.com/auth0/Auth0.swift/blob/master/Auth0Tests/WebAuthErrorSpec.swift#L27

Widcket (approved):

Issue #862, opened 2024-07-09
Getting "No credentials were found in the store." on production app


Kaur-Pushpinder:
### Checklist

- [X] The issue can be reproduced in the [Auth0.swift sample app](https://github.com/auth0-samples/auth0-ios-swift-sample/tree/master/Sample-01) (or N/A).
- [X] I have looked into the [Readme](https://github.com/auth0/Auth0.swift#readme), [Examples](https://github.com/auth0/Auth0.swift/blob/master/EXAMPLES.md), and [FAQ](https://github.com/auth0/Auth0.swift/blob/master/FAQ.md) and have not found a suitable solution or answer.
- [X] I have looked into the [API documentation](https://auth0.github.io/Auth0.swift/documentation/auth0) and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/Auth0.swift/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

This issue is randomly reproduced while renewing the token.

### Reproduction

This issue is randomly reproduced while renewing the token.
![IMG_0958](https://github.com/auth0/Auth0.swift/assets/139348758/6d357d7d-55e5-4eaa-b9f4-fad2b1fe6995)

### Auth0.swift version

2.7.2

### Platform

iOS

### Platform version(s)

17+

### Xcode version

15.3

### Package manager

Cocoapods

desusai7:
Hi @Kaur-Pushpinder, 

Can you help us with the steps to reproduce ?

And is the issue #863 also opened by someone from your team ?

Kaur-Pushpinder:
@desusai7 this is a random issue, there are no specific steps. just one info it is coming in renewAuth.

Kaur-Pushpinder:
@desusai7 any update on this?

Vaidios:
I am also experiencing similar issue, though I've been able to determine that in my case it breaks on `NSKeyedUnarchiver`
```
[path/ios.app]"
)}" UserInfo={NSDebugDescription=value for key 'root' was of unexpected class 'A0Credentials' (0x105529ff8) [/path/ios.app/Frameworks/Auth0.framework].
Allowed classes are:
 {(
[...]

PR #868, opened 2024-09-06
Replacing NSCoder with Swift's Codable

Vaidios:

### 📋 Changes

- Replaced occurrences of NSKeyedUnarchiver/NSKeyedArchiver with JSONDecoder/JSONEncoder
- Added encode function to Credentials type

### 📎 References

It doesn't seem like usage of this API is required in this context. Credentials type doesn't use polymorphism, which could be only handled by NSKeyedUnarchiver.

Thanks to these changes my code now properly runs inside a SPM package

Please let me know if there is something that I don't understand :)

https://forums.swift.org/t/should-i-stick-with-codable-or-switch-back-to-nscoding/61604
https://github.com/auth0/Auth0.swift/issues/862

### 🎯 Testing

This is not a breaking change, these code paths are covered by unit tests


Auth0/CredentialsManager.swift
```diff
@@ -362,7 +362,12 @@ public struct CredentialsManager {
 
     private func retrieveCredentials() -> Credentials? {
         guard let data = self.storage.getEntry(forKey: self.storeKey) else { return nil }
-        return try? NSKeyedUnarchiver.unarchivedObject(ofClass: Credentials.self, from: data)
+        do {
+            return try JSONDecoder().decode(Credentials.self, from: data)
+        } catch {
+            print(error.localizedDescription)
```
theolampert, line 368: @Vaidios I know that this error was already not being surfaced here, so this is not a comment on your PR but more this process in general. If an `NSCoder` failure is the cause of #862 there is an argument to be made that the problem would have been easier to debug had this function either throw here or have the `.noCredentials` error provide a more descriptive message. What are your thoughts @desusai7 ?

Vaidios: I totally agree, it would be much easier to find a root cause if that error would be propagated.
Maybe we should throw an error such as 'credentialsParsingFailure' instead of non-descriptive 'noCredentials'.

PS. For the time being, I will remove that do-catch with print to be optional try, just as before

Vaidios:
@theolampert sorry for tagging you here, I was just wondering if there is any button to ask for the review from maintainers and I clicked it without thinking much about it :E

desusai7:
Hi @theolampert, @Vaidios,

Thank you for raising a PR to address this, we will look into it and get back to you !

Vaidios:
Hey @desusai7, are there any updates on this?

shusuke0812:
Hello @desusai7 

We are encountering the same issue which gets "No credentials were found in the store." message.
Could you please provide any updates on when this issue might be resolved? Thank you.

Widcket:
Hi everyone,

We cannot accept this PR because `JSONDecoder` cannot read credentials encoded with `NSKeyedArchiver`. This means all the users of all Auth0 customers using this SDK would be logged out after updating the app and would have to log in again.

This is precisely the reason why the existing `NSKeyedUnarchiver/NSKeyedArchiver`-based solution was left as-is during the development of Auth0.swift v2.

@shusuke0812 please see https://github.com/auth0/Auth0.swift/issues/862#issuecomment-2583238245 for a way to log the underlying Keychain error.

PR #912, opened 2025-03-11
Add support for Multi-Resource Refresh Token (MRRT)

Widcket:

### 📋 Changes

This PR moves the Credentials Manager from a single credentials model to a multiple credentials one, supporting:
- 1 set of app credentials (the existing functionality)
- N sets of API-specific credentials

To this end, two new public methods were added to the Credentials Manager:

```swift
// Plus its combine + async/await wrappers

func apiCredentials(forAudience audience: String,
                    scope: String? = nil,
                    minTTL: Int = 0,
                    parameters: [String: Any] = [:],
                    headers: [String: String] = [:],
                    callback: @escaping (CredentialsManagerResult<APICredentials>) -> Void)
```

```swift
func clear(forAudience audience: String) -> Bool
```

The Authentication API client's `renew(withRefreshToken:scope:)` method also gained a new `audience` argument —that defaults to `nil`— allowing the exchange of a refresh token for API-specific credentials.

### 🎯 Testing

A complete test suite is included.


Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -1222,14 +1829,72 @@ class CredentialsManagerSpec: QuickSpec {
 
             }
 
+            context("api credentials") {
```
Widcket, line 1832: These are the tests for the async/await wrapper.

Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -1037,6 +1559,92 @@ class CredentialsManagerSpec: QuickSpec {
...
```
Widcket, line 1562: These are the tests for the Combine wrapper. Combine is Apple's framework for functional reactive programming.

Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -755,20 +825,470 @@ class CredentialsManagerSpec: QuickSpec {
...
+                                done()
+                            }
+                        }
+                    }
+                }
+            }
+
+        }
+
+        describe("retrieval of api credentials") {
```
Widcket, line 859: These are the tests of `credentialsManager.apiCredentials()`.

Auth0/CredentialsManager.swift
```diff
@@ -753,6 +1096,80 @@ public extension CredentialsManager {
         }
     }
 
+    /// Retrieves API credentials from the Keychain and automatically renews them using the refresh token if the access
```
Widcket, line 1099: This is the async/await wrapper.

Auth0/CredentialsManager.swift
```diff
@@ -587,6 +835,99 @@ public extension CredentialsManager {
         }.eraseToAnyPublisher()
     }
 
+    /// Retrieves API credentials from the Keychain and automatically renews them using the refresh token if the access
```
Widcket, line 838: This is the Combine wrapper.

Auth0/CredentialsManager.swift
```diff
@@ -399,49 +585,111 @@ public struct CredentialsManager {
...
                             callback(.failure(CredentialsManagerError(code: .renewFailed, cause: error)))
                         }
                     }
             }
 
-            self.dispatchGroup.wait()
+            dispatchGroup.wait()
+        }
+    }
+
```
Widcket, line 612: This is the meat of the new Credentials Manager logic.

Auth0/CredentialsManager.swift
```diff
@@ -399,49 +585,111 @@ public struct CredentialsManager {
...
+                self.authentication
+                    .renew(withRefreshToken: refreshToken, audience: audience, scope: scope)
+                    .parameters(parameters)
+                    .headers(headers)
+                    .start { result in
+                        switch result {
+                        case .success(let credentials):
+                            let newCredentials = Credentials(from: currentCredentials,
+                                                             idToken: credentials.idToken,
+                                                             refreshToken: credentials.refreshToken ?? refreshToken)
```
Widcket, line 651: The ID token and the refresh token (if present) are replaced in the app credentials, which are then saved with these updated values.

Auth0/CredentialsManager.swift
```diff
@@ -294,27 +310,177 @@ public struct CredentialsManager {
...
+        self.retrieveCredentials(scope: scope,
+                                 minTTL: minTTL,
+                                 parameters: parameters,
+                                 headers: headers,
+                                 forceRenewal: false,
+                                 callback: callback)
     }
     #endif
 
+    /// Retrieves API credentials from the Keychain and automatically renews them using the refresh token if the access
```
Widcket, line 407: This is the main, callback-based `apiCredentials()`.

Auth0/CredentialsManager.swift
```diff
@@ -294,27 +310,177 @@ public struct CredentialsManager {
...
+            self.retrieveCredentials(scope: scope,
+                                     minTTL: minTTL,
+                                     parameters: parameters,
+                                     headers: headers,
+                                     forceRenewal: false,
+                                     callback: callback)
         }
     }
     #else
+    /// Retrieves credentials from the Keychain and automatically renews them using the refresh token if the access
```
Widcket, line 335: This was a missing bit of documentation for tvOS and watchOS targets.

Auth0/CredentialsManager.swift
```diff
@@ -120,7 +121,21 @@ public struct CredentialsManager {
     ///
     /// - Returns: If the credentials were removed.
     public func clear() -> Bool {
-        return self.storage.deleteEntry(forKey: storeKey)
+        return self.storage.deleteEntry(forKey: self.storeKey)
+    }
+
+    /// Clears API credentials stored in the Keychain for a given audience value.
```
Widcket, line 127: This is the second new method.

Auth0/Auth0Authentication.swift
```diff
@@ -276,19 +276,28 @@ struct Auth0Authentication: Authentication {
...
             "client_id": self.clientId
         ]
-        payload["scope"] = scope
         let oauthToken = URL(string: "oauth/token", relativeTo: self.url)!
+
+        if let audience = audience {
+            // Make sure to always include the 'openid' scope if we're trying to get a new set of credentials for an
+            // API like My Account API. This way, we'll always get an ID token, matching the existing login behavior.
+            payload["scope"] = includeRequiredScope(in: scope)
+            payload["audience"] = audience
```
Widcket, line 291: We're enforcing the `openid` scope only when exchanging for API-specific creddentials.

pmathew92: Do we need to add any validation for the scenario where the user passes an audience but misses passing a scope  ?

Auth0/CredentialsManager.swift
pmathew92: Would this be easier if we have only one clear method with a default value of the `storeKey`
`public func clear(key:String = self. storeKey) {
return self.storage.deleteEntry(forKey: key)
}
`

Auth0/CredentialsManager.swift
```diff
@@ -357,38 +523,58 @@ public struct CredentialsManager {
...
+        return try? APICredentials(from: data)
+    }
+
+    // swiftlint:disable:next function_parameter_count
+    private func retrieveCredentials(scope: String?,
+                                     minTTL: Int,
                                      parameters: [String: Any],
                                      headers: [String: String],
-                                     forceRenewal: Bool = false,
+                                     forceRenewal: Bool,
```
pmathew92, line 557: Any particular reason to remove the default value from forceRenewal parameter ?

Widcket: So it's explicit at the call site.

Auth0/CredentialsManager.swift
Widcket: The default value would have to be a constant at build time, which `storeKey` is not.

pmathew92 (dismissed):
LGTM

NandanPrabhu (approved):

pmathew92 (approved):

PR #981, opened 2025-07-24
Added support for custom token exchange EA

cocojoe:
Added tests
Added claude helper

- [ ] I have added documentation for all new/changed functionality (or N/A)

### DIFF
--- a/Auth0/Auth0Authentication.swift
+++ b/Auth0/Auth0Authentication.swift
@@ -476,6 +476,10 @@ struct Auth0Authentication: Authentication {
                        telemetry: self.telemetry)
     }
 
+    func customTokenExchange(subjectTokenType: String, subjectToken: String, scope: String = "openid profile email", audience: String? = nil) -> Request<Credentials, AuthenticationError> {
+        return self.tokenExchange(subjectToken: subjectToken, subjectTokenType: subjectTokenType, scope: scope, audience: audience)
+    }
+
 }
 
 // MARK: - Private Methods
--- a/Auth0/Authentication.swift
+++ b/Auth0/Authentication.swift
@@ -923,6 +923,38 @@ public protocol Authentication: Trackable, Loggable {
     func codeExchange(withCode code: String, codeVerifier: String, redirectURI: String) -> Request<Credentials, AuthenticationError>
 
     /**
+     Exchanges a subject token for Auth0 tokens using the OAuth 2.0 Token Exchange spec.
+
+     - Important: Custom Token Exchange is an Early Access feature for Auth0 Enterprise customers. You must enable it for your application and configure a Custom Token Exchange Profile and Action in your Auth0 Dashboard. See: https://auth0.com/docs/authenticate/custom-token-exchange
+     - Security: You are responsible for implementing strong validation logic in your Auth0 Action to prevent spoofing or replay attacks. See Auth0's security guidance for recommendations.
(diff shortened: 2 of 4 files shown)


Auth0/Auth0Authentication.swift
```diff
@@ -476,6 +476,10 @@ struct Auth0Authentication: Authentication {
                        telemetry: self.telemetry)
     }
 
+    func customTokenExchange(subjectTokenType: String, subjectToken: String, scope: String = "openid profile email", audience: String? = nil) -> Request<Credentials, AuthenticationError> {
```
Widcket, line 479: This should be using the [defaultScope](https://github.com/auth0/Auth0.swift/blob/2.13.0/Auth0/Auth0.swift#L33) property instead of hardcoding it.

Auth0/Authentication.swift
```diff
@@ -1184,4 +1216,8 @@ public extension Authentication {
         return self.renew(withRefreshToken: refreshToken, audience: audience, scope: scope)
     }
 
+    func customTokenExchange(subjectTokenType: String, subjectToken: String, scope: String = "openid profile email", audience: String? = nil) -> Request<Credentials, AuthenticationError> {
```
Widcket: This should be using the defaultScope property instead of hardcoding it.

cocojoe: Agreed, good catch

Auth0/Auth0Authentication.swift
Widcket: Default param values don't go here, they go in the respective extension method in Authentication.swift.

Auth0Tests/AuthenticationSpec.swift
```diff
@@ -1634,5 +1634,71 @@ class AuthenticationSpec: QuickSpec {
...
+            }
+            it("should exchange a custom token for credentials") {
+                waitUntil(timeout: Timeout) { done in
+                    auth.customTokenExchange(subjectTokenType: subjectTokenType, subjectToken: subjectToken).start { result in
+                        expect(result).to(haveCredentials(AccessToken, IdToken))
+                        done()
+                    }
+                }
+            }
+            it("should exchange a custom token for credentials with audience") {
```
Widcket, line 1677: Missing a test case with a custom scope value.

Auth0/Auth0Authentication.swift
```diff
@@ -476,6 +476,10 @@ struct Auth0Authentication: Authentication {
                        telemetry: self.telemetry)
     }
 
+    func customTokenExchange(subjectTokenType: String, subjectToken: String, scope: String, audience: String? = nil) -> Request<Credentials, AuthenticationError> {
```
Widcket: `audience` should not have a default value here (`audience: String? = nil`). Default values for params go in the respective extension method in Authentication.swift

cocojoe: I mean other methods like `renew` do it. However, will fix

Widcket: That's probably a leftover from years ago.

Widcket:
@cocojoe have you tested this manually yet?

cocojoe:
No, I intend to. Mostly this was a first pass to see if AI added value and how close it was to revise guidance. I'm going to try a few different approaches. Will probably close this PR at some point.

PR #989, opened 2025-07-31
Added Custom Token Exchange (CTE) Support

cocojoe:
✅ Features Added:
customTokenExchange method in the Authentication protocol (Auth0/Authentication.swift:1102) 
Implementation in Auth0Authentication struct (Auth0/Auth0Authentication.swift:469) 
Default parameter extension (Auth0/Authentication.swift:1235) 
Comprehensive unit tests (Auth0Tests/AuthenticationSpec.swift:1007-1089)

✅ Key Implementation Details:

Endpoint: Uses /oauth/token with urn:ietf:params:oauth:grant-type:token-exchange grant type
Parameters: Accepts subjectTokenType, subjectToken, optional audience, and scope (defaults to openid profile email)
Scope Enforcement: Automatically includes openid scope when needed via the existing parameters() method
Error Handling: Properly returns AuthenticationError for failed exchanges
Documentation: Complete Swift documentation with usage examples

✅ Testing Coverage:

Default parameter usage
Custom scope values
Audience parameter
Invalid token error scenarios
All using the same testing patterns as existing authentication methods

### 🎯 Testing

Have manually tested this with a prod tenant, with custom scope/audience


cocojoe:
Not clear why codecov is failing, the PR adds 1.03% coverage.

cocojoe:
@Widcket, let me know your thoughts on the additional parameters, it was an interesting addition.Context:

> The additionalParameters was included to provide flexibility for future use cases like custom claims, resource parameters, or other RFC 8693-defined optional parameters, but it could be removed if the        
>   API should be kept minimal and focused on the core token exchange functionality.

Auth0/Authentication.swift
```diff
@@ -1080,6 +1080,57 @@ public protocol Authentication: Trackable, Loggable {
...
+       - additionalParameters: Additional parameters to include in the token exchange request. Defaults to empty dictionary.
+     - Returns: A request that will yield Auth0 user's credentials.
+
+     ## See Also
+
+     - [Authentication API Endpoint](https://auth0.com/docs/api/authentication/token-exchange)
+     - [Custom Token Exchange Documentation](https://auth0.com/docs/authenticate/custom-token-exchange)
+     - [RFC 8693: OAuth 2.0 Token Exchange](https://tools.ietf.org/html/rfc8693)
+     */
+    func customTokenExchange(subjectToken: String, subjectTokenType: String, audience: String?, scope: String, additionalParameters: [String: Any]) -> Request<Credentials, AuthenticationError>
```
Widcket: Custom parameters (`additionalParameters`) are already supported. The `Request` type has a `parameters()` method for this purpose: https://github.com/auth0/Auth0.swift/blob/master/Auth0/Request.swift#L97

Widcket: Since the method is in the `Request` type, it's available for all API client methods that return a `Request`.

cocojoe: Ah yes, that was it, had something in the back of my mind on this as couldn't see any other methods adding it and I was looking for 'additionalParameters' I'll revise that one.

I just liked that Claude was curious from the RFC😄

Auth0/Authentication.swift
```diff
@@ -1080,6 +1080,57 @@ public protocol Authentication: Trackable, Loggable {
      */
     func jwks() -> Request<JWKS, AuthenticationError>
 
+    /**
+     Performs a custom token exchange to obtain Auth0 credentials using an existing identity provider token.
+
```
Widcket, line 1096: Please add a notice about EA like https://github.com/auth0/Auth0.swift/blob/master/Auth0/Authentication.swift#L499-L503

cocojoe: Do we circle back and update these when they hit GA?

Widcket: We should, yes.

Widcket: Similar notices are also in EXAMPLES.md.

Auth0/Authentication.swift
cocojoe: Done

Auth0/Authentication.swift
cocojoe: That's fair then, if they were not being kept up to date I would say we should drop the release stage disclaimer. I'll add

cocojoe: done

NandanPrabhu:
@cocojoe please update examples.md file as part of documentation

cocojoe:
Added Examples

Issue #950, opened 2025-05-08
Crash in Auth0WebAuth.swift


963239327:
### Checklist

- [x] The issue can be reproduced in the [Auth0.swift sample app](https://github.com/auth0-samples/auth0-ios-swift-sample/tree/master/Sample-01) (or N/A).
- [x] I have looked into the [Readme](https://github.com/auth0/Auth0.swift#readme), [Examples](https://github.com/auth0/Auth0.swift/blob/master/EXAMPLES.md), and [FAQ](https://github.com/auth0/Auth0.swift/blob/master/FAQ.md) and have not found a suitable solution or answer.
- [x] I have looked into the [API documentation](https://auth0.github.io/Auth0.swift/documentation/auth0) and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/Auth0.swift/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

Stack Traces:
![Image](https://github.com/user-attachments/assets/7f581759-5b39-4079-857b-7a590512cb1d)

From Apple's document: `You must invoke the continuation’s resume method exactly once.`
![Image](https://github.com/user-attachments/assets/b8f9805b-a571-4aad-a405-950f46c7b893)

In line 208 and line 216, it will be called twice.
![Image](https://github.com/user-attachments/assets/57c22a35-1e8e-4bba-80f3-8edf642e0fda)

### Reproduction

N/A

### Auth0.swift version

2.10.0

### Platform

iOS

### Platform version(s)

All

### Xcode version

16.3

### Package manager

Swift Package Manager

Widcket:
Hi @963239327, thanks for reporting this.

Is this crash happening in a situation similar to this one? https://github.com/auth0/Auth0.swift/issues/827#issuecomment-1929412412

If so, please update to the latest version of the SDK. It'll return an error when attempting to create concurrent login operations: https://github.com/auth0/Auth0.swift/blob/master/Auth0/Auth0WebAuth.swift#L175-L177

963239327:
Hi @Widcket, thanks for your quick response!

[...]

PR #1008, opened 2025-09-23
Added fix for GH issue 950 

NandanPrabhu:

### 📋 Changes

This PR covers fix for a crash reported as a github issue #950. Crash is reported as a result of continuation resumed twice. The fix is a temporary fix. We aim to rewrite the ASWebAuthenticationSession layer with fully async await. The fix is tested with both success and fail cases.

### 🎯 Testing

Tested universal login flow success and failure cases

### DIFF
--- a/Auth0/Auth0WebAuth.swift
+++ b/Auth0/Auth0WebAuth.swift
@@ -355,9 +355,12 @@ extension Auth0WebAuth {
 extension Auth0WebAuth {
 
     func start() async throws -> Credentials {
+        var alreadyResumed = false
         return try await withCheckedThrowingContinuation { continuation in
             Task { @MainActor in
                 self.start { result in
+                    guard !alreadyResumed else { return }
+                    alreadyResumed = true
                     continuation.resume(with: result)
                 }
             }
--- a/Auth0Tests/WebAuthSpec.swift
+++ b/Auth0Tests/WebAuthSpec.swift
@@ -1,4 +1,5 @@
 import Quick
+import Combine
 import Nimble
 import SafariServices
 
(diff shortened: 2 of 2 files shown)


pmathew92 (dismissed):

Auth0/Auth0WebAuth.swift
```diff
@@ -355,9 +355,12 @@ extension Auth0WebAuth {
 extension Auth0WebAuth {
 
     func start() async throws -> Credentials {
+        var alreadyResumed = false
```
pmathew92, line 358: Would it make sense to make this reusable as mentioned in
https://github.com/auth0/Auth0.swift/issues/950#issuecomment-2868557222
as this can occur at all places where we are mapping a callback to async type ?

NandanPrabhu: @pmathew92 Good suggestion. right now we are adding quick fix for the crash. So, either way would work the same. For a long term fix, I am gonna raise a issue with Apple reporting this issue where in ASWebAuthenticationSession returns callback multiple times for certain scenarios. With this we will probably in this or next quarter be focusing on refactoring web authentication completely written using async await. Hence removing closure based implementation end to end

Auth0/Auth0WebAuth.swift
```diff
@@ -355,9 +355,12 @@ extension Auth0WebAuth {
...
         return try await withCheckedThrowingContinuation { continuation in
             Task { @MainActor in
                 self.start { result in
+                    guard !alreadyResumed else { return }
+                    alreadyResumed = true
```
pmathew92, line 363: Can we add a UT for checking this scenario ?

NandanPrabhu: added @pmathew92

walkerheathr24-rgb (commented):
Thank you for all your help!

PR #1027, opened 2025-11-27
Added Agents.md file

NandanPrabhu:
### 📋 Changes

-  Added AGENTS.md file to the repo 

AGENTS.md complements this by containing the extra, sometimes detailed context coding agents need: build steps, tests, and conventions that might clutter a README or aren’t relevant to human contributors.

### 📎 References
https://agents.md/

### DIFF
--- a/Auth0/BioAuthentication.swift
+++ b/Auth0/BioAuthentication.swift
@@ -23,13 +23,21 @@ struct BioAuthentication {
     var available: Bool {
         return self.authContext.canEvaluatePolicy(evaluationPolicy, error: nil)
     }
+    
+    /// Creates a new LAContext with the same settings for fresh biometric prompts.
+    private func createFreshContext() -> LAContext {
+        let newContext = LAContext()
+        newContext.localizedFallbackTitle = self.fallbackTitle
+        newContext.localizedCancelTitle = self.cancelTitle
+        return newContext
+    }
 
     init(authContext: LAContext,
          evaluationPolicy: LAPolicy,
          title: String,
          cancelTitle: String? = nil,
          fallbackTitle: String? = nil,
-         policy: BiometricPolicy = .always
+         policy: BiometricPolicy = .default
     ) {
         self.authContext = authContext
         self.evaluationPolicy = evaluationPolicy
@@ -40,7 +48,18 @@ struct BioAuthentication {
     }
 
     func validateBiometric(callback: @escaping (Error?) -> Void) {
-        self.authContext.evaluatePolicy(evaluationPolicy, localizedReason: self.title) {
+        let contextToUse: LAContext
+        
+        switch self.policy {
+        case .default:
+            // Use the existing context (preserves current behavior)
+            contextToUse = self.authContext
+        case .always, .session, .appLifecycle:
+            // Create a fresh context to ensure biometric prompt is shown
+            contextToUse = self.createFreshContext()
+        }
+        
+        contextToUse.evaluatePolicy(evaluationPolicy, localizedReason: self.title) {
             guard $1 == nil else { return callback($1) }
             callback($0 ? nil : LAError(.authenticationFailed))
         }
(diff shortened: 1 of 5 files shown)


AGENTS.md
```diff
@@ -0,0 +1,239 @@
+I understand. Since you are looking for the **Auth0.swift** (iOS/macOS/tvOS/watchOS) equivalent of the `Agents.md` file you shared, I have adapted the structure and content to reflect the specific architecture, patterns (Async/Await, Combine), and tooling (SPM, Xcode) used in the Swift SDK.
+
+Here is the **Agents.md** for **Auth0.swift**.
+
+-----
+
+# AI Agent Guidelines for Auth0.swift SDK
```
subhankarmaiti:
```suggestion
# AI Agent Guidelines for Auth0.swift SDK
```

AGENTS.md
```diff
@@ -0,0 +1,239 @@
+I understand. Since you are looking for the **Auth0.swift** (iOS/macOS/tvOS/watchOS) equivalent of the `Agents.md` file you shared, I have adapted the structure and content to reflect the specific architecture, patterns (Async/Await, Combine), and tooling (SPM, Xcode) used in the Swift SDK.
+
+Here is the **Agents.md** for **Auth0.swift**.
```
pmathew92: Please remove the above text .

subhankarmaiti (commented):
can we keep the app folder as same before?

AGENTS.md
pmathew92, line 1: I would suggest you ask vscode to create minimum Agents.md file and ask claude to refine it based on the current code base ,that would give you a better Agents.md file

App/ViewController.swift
```diff
@@ -1,49 +0,0 @@
-import UIKit
-import Auth0
-
-class ViewController: UIViewController {
```
sanchitmehtagit, line 4: q: should we  have Sample App basis of UIKit and SwiftUI both. I saw we added SwiftUI Sample app integration in this PR

pmathew92: Not needed IMO. The purpose of the sample app is for showcasing the SDK implementation so no need to maintain for both UIKit and SwiftUI

sanchitmehtagit: I thought this will serve as a reference for our end customers, who may integrate it using either SwiftUI or UIKit. I agree, though, that the main goal should be to provide clear boilerplate code for SDK integration, including APIs, callbacks, and error handling, etc and as long ad SDK doesnt provide any UI component which customer would integrate to swiftui or uikit codebase should be okay.

App/ContentView.swift
```diff
@@ -0,0 +1,49 @@
...
+            Button {
+                Auth0
+                    .webAuth()
+                    .logging(enabled: true)
+                    .start {
+                        switch $0 {
+                        case .failure:
+                            break
+                        case .success:
+                            break
```
sanchitmehtagit: nit: i believe success would return some credentials and token  could we show an alert of something ceredntials fetched just like UIKit integration

App/ContentView.swift
```diff
@@ -0,0 +1,49 @@
...
+                    }
+            } label: {
+                Text("Login")
+            }.padding(.horizontal)
+            
+            Button {
+                Auth0
+                    .webAuth()
+                    .logging(enabled: true)
+                    .clearSession(federated: false) { result in
```
sanchitmehtagit: as this sample app integration would  be share with customers i would recommend to have async await based integration for sample app as that is the preferred way as mentioned in  the md file.

pmathew92: Hi @sanchitmehta94 , this app is not something we share with customers. It is primarily used for development purpose and testing. We do have a separate sample app for customer reference
https://github.com/auth0-samples/auth0-ios-swift-sample

AGENTS.md
NandanPrabhu: done

PR #1031, opened 2025-12-04
refactor: updated the key logic for storing MRRT token

pmathew92:

### 📋 Changes

This PR updates the current key for storing token while using the MRRT flow. The current implementation maps token to each audience. This change considers scope also into the key logic as same audience can have different scope and can get different tokens issued for each scope. One example is the my-account use case.

### DIFF
--- a/Auth0.xcodeproj/project.pbxproj
+++ b/Auth0.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		2Z7FPPVSUUJRZC06V5RA8B2B /* BiometricPolicy.swift in Sources */ = {isa = PBXBuildFile; fileRef = KDG54I4VDN4TZII38JVUS4AP /* BiometricPolicy.swift */; };
 		5B0893E620F8A52100FBF962 /* CredentialsManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5BEDE1891EC21B040007300D /* CredentialsManager.swift */; };
 		5B0893E720F8A52400FBF962 /* CredentialsManagerError.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5B5E93F81EC45C22002A37F9 /* CredentialsManagerError.swift */; };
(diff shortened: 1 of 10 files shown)


Auth0/CredentialsManager.swift
```diff
@@ -694,23 +696,34 @@ public struct CredentialsManager {
...
+    private func retrieveAPICredentials(audience: String,scope: String?) -> APICredentials? {
+        let key = getAPICredentialsStorageKey(audience: audience, scope: scope)
+        guard let data = self.storage.getEntry(forKey: key) else { return nil }
         return try? APICredentials(from: data)
     }
+    
+    private func getAPICredentialsStorageKey(audience: String, scope: String?) -> String {
+        // Use audience if scope is null else use a combination of audience and scope
+        if let scope = scope {
+            return "\(audience)::\(scope.replacingOccurrences(of: " ", with: "::"))"
```
sanchitmehtagit: Suggestion
if we have two scopes like this but the order is different  ,
"user.read user.write"
"user.write user.read"

This would create duplicate keys for our cache . would recommend to do something like below
let normalisedScopes = scope
    .split(separator: " ")
    .sorted()
    .joined(separator: "::")

Auth0/CredentialsManager.swift
```diff
@@ -893,14 +906,30 @@ public struct CredentialsManager {
...
-    func hasScopeChanged(from lastScope: String?, to newScope: String?) -> Bool {
+    func hasScopeChanged(from lastScope: String?, to newScope: String?, ignoreOpenid: Bool = false) -> Bool {
+        
         if let lastScope = lastScope, let newScope = newScope {
-            let lastScopeList = lastScope.lowercased().split(separator: " ").sorted()
-            let newScopeList = newScope.lowercased().split(separator: " ").sorted()
-
-            return lastScopeList != newScopeList
+            
+            var storedScopes = Set(
```
sanchitmehtagit, line 918: hasScope change should have sorted () as in the old code
"user.read user.write"
"user.write user.read"
as this is not considered a scope change

sanchitmehtagit: pls ignore we are doing set comparison

sanchitmehta94 (approved):
LGTM

PR #1035, opened 2025-12-09
Org support for  custom token exchange

NandanPrabhu:
This is PR is about adding **organization** param support as part of the custom token exchange API

**Custom Token Exchange** (CTE) allows customers to implement custom token exchange functionality, providing an external token to the Auth0 Authentication API, which is exchanged for standard Auth0 tokens.
https://auth0.com/docs/authenticate/custom-token-exchange

**Auth0 organisations** 
The Organizations feature represents a broad update to the Auth0 platform that allows our business-to-business (B2B) customers to better manage their partners and customers, and to customize the ways that end-users access their applications and APIs. Auth0 customers can use Organizations to:

### Files changed
Authentication.swift - Authentication protocol is updated with organization param in the method signature

Auth0Authentications.swift - customTokenExchange() implementation is updated to accept organization param and the token() is updated to accept organization param in the call

### Before
  `func customTokenExchange(subjectToken: String, subjectTokenType: String, audience: String?, scope: String) -> Request<Credentials, AuthenticationError>`

### After 
 ```
    func customTokenExchange(subjectToken: String,
                             subjectTokenType: String,
                             audience: String?,
                             scope: String,
                             organization: String?) -> Request<Credentials, AuthenticationError>
```

### DIFF
--- a/.github/actions/rl-scanner/action.yml
+++ b/.github/actions/rl-scanner/action.yml
@@ -22,7 +22,7 @@ runs:
         pip install boto3 requests
 
     - name: Configure AWS credentials
-      uses: aws-actions/configure-aws-credentials@00943011d9042930efac3dcd3a170e4273319bc8 # pin@v5.1.0
+      uses: aws-actions/configure-aws-credentials@61815dcd50bd041e203e49132bacad1fd04d2708 # pin@v5.1.1
       with:
         role-to-assume: ${{ env.PRODSEC_TOOLS_ARN }}
         aws-region: us-east-1

### CODE REVIEW

File: Auth0/Authentication.swift [hunk 1]
```diff
@@ -1131,17 +1138,20 @@ public protocol Authentication: SenderConstraining, Trackable, Loggable {
...
      - [Custom Token Exchange Documentation](https://auth0.com/docs/authenticate/custom-token-exchange)
      - [RFC 8693: OAuth 2.0 Token Exchange](https://tools.ietf.org/html/rfc8693)
      */
-    func customTokenExchange(subjectToken: String, subjectTokenType: String, audience: String?, scope: String) -> Request<Credentials, AuthenticationError>
-
+    func customTokenExchange(subjectToken: String,
+                             subjectTokenType: String,
+                             audience: String?,
+                             scope: String,
+                             organization: String?) -> Request<Credentials, AuthenticationError>
```
pmathew92: Shouldn't organization have default value of nil ?

NandanPrabhu, line 1154: default value cannot be added in protocol definition it can be added in extension of the protocol or concrete implementation of the protocol. in extension its added here https://github.com/auth0/Auth0.swift/pull/1035/files#diff-ee814ae7c7ac0aa751f636732eda41d4c2a54a5e38ab05af6067ba6425aa3b08R1278

Auth0/Auth0Authentication.swift
```diff
@@ -535,14 +547,20 @@ private extension Auth0Authentication {
...
+                       organization: String? = nil,
+                       parameters: [String: Any] = [:]) -> Request<Credentials, AuthenticationError> {
         var parameters: [String: Any] = parameters
         parameters["client_id"] = self.clientId
         parameters["grant_type"] = "urn:ietf:params:oauth:grant-type:token-exchange"
         parameters["subject_token"] = subjectToken
         parameters["subject_token_type"] = subjectTokenType
         parameters["audience"] = audience
         parameters["scope"] = includeRequiredScope(in: scope)
+        parameters["organization"] = organization
```
pmathew92, line 563: This still add a nil to the parameters map if organization is not passed. don't we need any checks here ?

Auth0/Authentication.swift
```diff
@@ -380,13 +384,17 @@ public protocol Authentication: SenderConstraining, Trackable, Loggable {
        - profile:            The user profile data retrieved from Facebook.
        - audience:           API Identifier that your application is requesting access to.
        - scope:              Space-separated list of requested scope values. Defaults to `openid profile email`.
+       - organization:       Identifier of an organization the user is a member of.
```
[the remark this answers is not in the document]
pmathew92: Remove the unwanted `organization` parameter from comment

Auth0/Auth0Authentication.swift
NandanPrabhu: no it wont because dictionary we have here is [String: Any] and not [String: Any?] if the value is nil, the dictionary won't accept the value and create a corresponding key value pair

Auth0/Auth0Authentication.swift
```diff
@@ -499,11 +506,16 @@ struct Auth0Authentication: Authentication {
                        telemetry: self.telemetry)
     }
 
-    func customTokenExchange(subjectToken: String, subjectTokenType: String, audience: String?, scope: String) -> Request<Credentials, AuthenticationError> {
+    func customTokenExchange(subjectToken: String,
```
sanchitmehtagit, line 509: lets update example.md 
file too it has a section for customTokenExchange

NandanPrabhu: done

pmathew92 (approved):

PR #1048, opened 2025-12-15
Make credential renewal resilient to concurrency conditions

sanchitmehtagit:
- [ ] I have added documentation for all new/changed functionality (or N/A)

### 📋 Changes

This should fix the issues where customers are trying to create multiple instances of credential manager and calling credential APIs and getting renewal Errors, This has caused multiple ESDs in the past and also will help close open ESD and github issue on React native

This pull request improves the Auth0 Swift SDK’s credential renewal mechanism to handle concurrency more safely. Previously, simultaneous calls to credential APIs from multiple instances of CredentialManager could cause renewal errors. The updated implementation ensures:

Credential renewal operations are synchronized without blocking threads unnecessarily.

Multiple instances of CredentialManager can safely invoke credential-related APIs concurrently.

Thread safety and reliability are improved under high-concurrency scenarios. This will help reduce silent renewal Errors where clients can be logged out intermittently due to such conditions

There is one more scenarios Using the Authentication API client to renew the credentials without synchronization as mentioned [here](https://github.com/auth0/Auth0.swift/blob/adabc99ad9d4d21232de1a114cec5f3850680a74/Documentation.docc/RefreshTokens.md#using-the-authentication-api-client-to-renew-the-credentials-without-synchronization) We have not made any changes to  improve renewal API Autentication API client that could be improved too but that required deduplication of request at SDK. will take that separately
I will handle request deduplication for that


Documentation.docc/RefreshTokens.md
```diff
@@ -44,21 +44,6 @@ Rotating refresh tokens are effectively single-use. The renewed credentials will
 
 ### Common causes
 
-#### Multiple Credentials Manager instances
-
-If you are using the Credentials Manager to renew the credentials -either through the `credentials()` or the `renew()` method- you should use a single Credentials Manager instance. While these methods are thread-safe, the Credentials Manager cannot synchronize them across instances.
```
sanchitmehtagit, line 49: this is safe now even if we create multiple instances at client side there would not be any race conditions

Auth0/SynchronizationBarrier.swift
```diff
@@ -0,0 +1,80 @@
...
+    ///   any thread. Operations are executed one at a time in the order they were submitted.
+    func execute(_ operation: @escaping (@escaping () -> Void) -> Void) {
+        // If reentrant, execute immediately
+        if isInsideBarrier {
+            operation {}
+            return
+        }
+
+        queue.async { [weak self] in
+            guard let self = self else { return }
```
sanchitmehtagit, line 39: I simplified the synchronization logic. Previously, we were using DispatchGroup.wait(), which blocks the current thread. The new implementation avoids calling group.wait() and instead prevents thread blocking altogether.

App/ViewController.swift
```diff
@@ -6,6 +6,7 @@ class ViewController: UIViewController {
     @IBAction func login(_ sender: Any) {
         Auth0
             .webAuth()
+            .scope("openid profile offline_access")
             .logging(enabled: true)
```
sanchitmehtagit, line 9: added offline_Access to retrieve refresh token

Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -932,6 +932,160 @@ class CredentialsManagerSpec: QuickSpec {
...
+                        expect(results.count) == callCount
+                        
+                        for result in results {
+                            expect(result).to(haveCredentials(NewAccessToken, NewIdToken, NewRefreshToken))
+                        }
+                        
+                        done()
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
sanchitmehtagit: we are not deduplicating the requests but we are making sure all the requests go one flight at a time

Auth0/SynchronizationBarrier.swift
```diff
@@ -0,0 +1,80 @@
...
+final class SynchronizationBarrier {
+
+    /// Shared singleton instance used by both CredentialsManager and Auth0Authentication
+    static let shared = SynchronizationBarrier()
+
+    /// Serial dispatch queue for coordinating operations
+    private let queue = DispatchQueue(label: "com.auth0.synchronization.serial")
+    
+    /// Queue of pending operations waiting to be executed
+    private var pendingOperations: [(@escaping () -> Void) -> Void] = []
```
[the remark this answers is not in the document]
sanchitmehtagit: `While the current implementation appears safe because all accesses happen within queue.async`

The pendingOperations array is thread safe as we are access it from a serial queue and update it within a seria queue

Auth0/SynchronizationBarrier.swift
```diff
@@ -0,0 +1,80 @@
...
+    /// Operations are queued and executed sequentially. Each operation must call its completion
+    /// handler to signal that it has finished, allowing the next operation to proceed.
+    ///
+    /// - Parameter operation: The operation to execute. The operation will receive a completion
+    ///   handler that it **must** call when finished. The completion handler can be called from
+    ///   any thread. Operations are executed one at a time in the order they were submitted.
+    func execute(_ operation: @escaping (@escaping () -> Void) -> Void) {
+        // If reentrant, execute immediately
+        if isInsideBarrier {
+            operation {}
```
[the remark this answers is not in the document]
sanchitmehtagit: this is not needed i will remove that

App/ViewController.swift
```diff
@@ -34,6 +43,42 @@ class ViewController: UIViewController {
...
+                        print("CredentialsManager \(i): FAILED - Duration: \(String(format: "%.2f", duration))s - Error: \(error)")
+                    }
+                    group.leave()
+                }
+            }
+        }
+        
+        group.notify(queue: .main) {
+            print("=== All CredentialsManager instances completed ===\n")
+        }
```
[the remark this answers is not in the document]
pmathew92: Agree on the above comment. This is not needed.

Auth0/SynchronizationBarrier.swift
```diff
@@ -0,0 +1,80 @@
...
+            return
+        }
+
+        queue.async { [weak self] in
+            guard let self = self else { return }
+
+            self.pendingOperations.append(operation)
+            self.tryExecuteNext()
+        }
```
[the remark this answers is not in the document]
sanchitmehtagit: no need

App/ViewController.swift
```diff
@@ -34,6 +43,42 @@ class ViewController: UIViewController {
                 }
             }
     }
+    
+    func testConcurrentCredentialsManagerCalls() {
+        print("\n=== Testing concurrent CredentialsManager.credentials() across multiple instances ===")
```
NandanPrabhu: lets adds async let, async group variants as well for testing concurrent calls of credentials manager

App/ViewController.swift
```diff
@@ -34,6 +43,42 @@ class ViewController: UIViewController {
...
+                credentialsManager.credentials { result in
+                    let endTime = Date()
+                    let duration = endTime.timeIntervalSince(startTime)
+                    
+                    switch result {
+                    case .success(_):
+                        print("CredentialsManager \(i): SUCCESS - Duration: \(String(format: "%.2f", duration))s")
+                    case .failure(let error):
+                        print("CredentialsManager \(i): FAILED - Duration: \(String(format: "%.2f", duration))s - Error: \(error)")
+                    }
```
NandanPrabhu: instead of having these testing code here. Lets have it in the tests folder

Auth0/SynchronizationBarrier.swift
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    private func tryExecuteNext() {
+        guard !isExecuting, !pendingOperations.isEmpty else { return }
+
+        isExecuting = true
+        let operation = pendingOperations.removeFirst()
+
+        // Run operation
+        operation { [weak self] in
```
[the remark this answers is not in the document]
sanchitmehtagit: not needed weak self is better option here

App/ViewController.swift
sanchitmehtagit: i have added test cases too  this is for app simulations with real endpoints. wil remove this code

Auth0/SynchronizationBarrier.swift
```diff
@@ -0,0 +1,80 @@
+import Foundation
+
+/// Internal utility for synchronizing operations across CredentialsManager instances
+/// and the Authentication API client. This ensures that only one operation is
+/// in flight at a time, preventing concurrency issues when Refresh Token Rotation is enabled.
+///
+/// The barrier uses a serial queue to maintain operation ordering and a queue-based system
+/// to ensure each operation completes before the next one begins, without using dispatch groups
+/// or semaphores that would block threads.
+final class SynchronizationBarrier {
```
NandanPrabhu, line 10: let's try to unify queue barrier and synchronization barrier. Purpose of both is the same that is to prevent duplicate execution of a function. also can we have an actor instead of a class and ensure the function belonging to Auth0WebAuth or CredentialsManger runs in the global actor context

Auth0/SynchronizationBarrier.swift
```diff
@@ -0,0 +1,80 @@
...
+            self.pendingOperations.append(operation)
+            self.tryExecuteNext()
+        }
+    }
+
+    private func tryExecuteNext() {
+        guard !isExecuting, !pendingOperations.isEmpty else { return }
+
+        isExecuting = true
+        let operation = pendingOperations.removeFirst()
```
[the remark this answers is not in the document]
sanchitmehtagit: not needed serial queue prevents that as mentioned
`While the serial queue should prevent this, the logic would be clearer and safer if structured as:`

Auth0/SynchronizationBarrier.swift
sanchitmehtagit: The purpose of these two approaches is different. QueueBarrier was added to prevent ASWebAuthenticationSessionCrash. I’m not deduplicating calls; rather, I’m ensuring that all API calls execute sequentially. QueueBarrier drops other calls, but in our case, we want every API call to run, one at a time. You can check the previous implementation for reference. 

We should avoid using actors here because doing so would require adding interoperability, which introduces extra overhead. Since all CredentialManager APIs are callback-based, using actors would also require a bridging layer which is additional cost.  Also, Actors would not solve the problem we would still have to add handling for a single request at a time with actors too.

let me know if we want to discuss further

App/ViewController.swift
sanchitmehtagit: will add async wait based test cases

App/ViewController.swift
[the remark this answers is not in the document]
sanchitmehtagit: +1

Auth0/CredentialsManager.swift
```diff
@@ -744,103 +743,87 @@ public struct CredentialsManager: Sendable {
...
+                            callback(.success(newCredentials))
                         }
+                    case .failure(let error):
+                        complete()
+                        callback(.failure(CredentialsManagerError(code: .renewFailed, cause: error)))
                     }
-            }
-
-            dispatchGroup.wait()
+                }
```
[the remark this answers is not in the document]
sanchitmehtagit: complete is called before calbacks() is called in all scenarios

Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -932,6 +932,160 @@ class CredentialsManagerSpec: QuickSpec {
 
         }
 
+        describe("concurrent credentials retrieval") {
+            
+            beforeEach {
+                NetworkStub.addStub(condition: { $0.isToken(Domain) && $0.hasAtLeast(["refresh_token": RefreshToken])}, response: authResponse(accessToken: NewAccessToken, idToken: NewIdToken, refreshToken: NewRefreshToken, expiresIn: ExpiresIn))
```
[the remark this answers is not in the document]
sanchitmehtagit: we are not deduplicating . we are making the second call using new refreshtoken.

Auth0/CredentialsManager.swift
```diff
@@ -744,103 +743,87 @@ public struct CredentialsManager: Sendable {
...
-                      self.hasScopeChanged(from: credentials.scope, to: scope) else {
-                    dispatchGroup.leave()
-                    return callback(.success(credentials))
-                }
-                guard let refreshToken = credentials.refreshToken else {
-                    dispatchGroup.leave()
-                    return callback(.failure(.noRefreshToken))
-                }
+        SynchronizationBarrier.shared.execute { complete in
+            guard let credentials = self.retrieveCredentials() else {
```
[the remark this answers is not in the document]
sanchitmehtagit: complete is called before callback()

Auth0/SynchronizationBarrier.swift
```diff
@@ -0,0 +1,64 @@
...
+
+            self.pendingOperations.append(operation)
+            self.tryExecuteNext()
+        }
+    }
+
+    private func tryExecuteNext() {
+        guard !isExecuting, !pendingOperations.isEmpty else { return }
+
+        isExecuting = true
```
pmathew92, line 49: Does isExecuting needs to be atomic ?

sanchitmehtagit: isExecuting is accessed (written and read) witnin a  serial queue hence thread safe

App/ViewController.swift
sanchitmehtagit: https://github.com/auth0/Auth0.swift/pull/1048/files#diff-62f64cae1a6d3cd35233f967dcd94229b5dcc21a578b10424ba2aad3590f41a1
Added asynct group test cases too

Documentation.docc/RefreshTokens.md
sanchitmehtagit: Credentials Manaager can synchornise them across instances now

NandanPrabhu (approved):

PR #1140, opened 2026-04-09
feat: Swift 6 @Sendable closures and CredentialsManager parameter Sendability

sanchitmehtagit:
## What this PR does

Continues the Swift 6 strict concurrency migration. Focuses on two independent problems:

1. **`@Sendable` closure annotations** across the networking and authentication layers.
2. **`[String: Any]` parameter capture** inside `CredentialsManager` — the root cause required a deliberate design choice explained below.

## The `[String: Any]` Sendability Problem

`CredentialsManager` accepts `parameters: [String: Any]` on its public API (e.g., `credentials(withScope:parameters:)`, `renew(parameters:)`, etc.).  

In Swift 6 strict concurrency, capturing `[String: Any]` across a concurrent closure boundary is a hard error:

```
error: capture of non-sendable type '[String : Any]' in a '@Sendable' closure
```

`[String: Any]` cannot conform to `Sendable` because `Any` is unconstrained — the compiler cannot guarantee all values inside are safe to share across isolation boundaries.

### 1. `@unchecked Sendable` wrapper (`SendableBox<[String: Any]>`)
Wrap the dictionary in the existing `SendableBox` and suppress the error.

**Why rejected:** This is the same as opting out of `Sendable` checking entirely. It silences the compiler without fixing the underlying issue, which is exactly what Swift 6 strict concurrency is meant to catch. If a caller ever passes a non-thread-safe reference type inside the dictionary, the bug would be invisible at compile time.

### 2. Change the public API to `[String: any Sendable]`
Constrain callers to only pass `Sendable` values.

**Why rejected:** This is a breaking public API change. Every existing caller passing `[String: Any]` (including user code) would need to be updated. The migration burden is not justified for a parameter that is always JSON-serializable in practice.

### 3. Change the public API to typed parameters (`Encodable` or a concrete struct)
Replace the dictionary with a typed model.

**Why rejected:** Also a breaking public API change, and it removes flexibility that callers currently rely on (e.g., passing arbitrary custom claims or extra body parameters without a named struct).

### 4. `nonisolated(unsafe)` on the capture
Mark the variable as `nonisolated(unsafe)` at the call site.

**Why rejected:** `nonisolated(unsafe)` is for static or stored properties with a well-understood external synchronization guarantee. It cannot be applied to local `let`/`var` constants, and even if it could, it would mean asserting to the compiler "I guarantee this is safe" without any actual protection — a foot-gun.

### 5. Serialize to `Data` at the API boundary ✅ (chosen)

```swift
// Entry point — serialize once
let params = normalize(parameters)   // Data, natively Sendable

// At point of use — deserialize
.parameters(params.asParameters)
```

`Data` is `Sendable` because it is a value type backed by a contiguous buffer of bytes. The `normalize` function serializes the dictionary to JSON `Data` at the moment the public method is called (on whatever isolation context the caller is on), before anything crosses a concurrent boundary. Downstream, `params.asParameters` deserializes back to `[String: Any]` at the point it is passed to the authentication request -  which is always single-threaded within `SynchronizationBarrier.shared.execute`.

This is the cleanest approach with good DX as no change on API layer

**Why this works:**

| Property | Guarantee |
|---|---|
| `Data` is `Sendable` | Compiler accepts capture in `@Sendable` closures with no suppression |
| Serialization is eager | Happens before the async boundary; no TOCTOU risk |
| Round-trip is lossless | Auth0 API parameters are always JSON-compatible (`String`, `Int`, `Bool`, nested dicts); `JSONSerialization` round-trips them faithfully |
| Zero public API change | Callers keep passing `[String: Any]`; the conversion is internal |

### `@Sendable` closure annotations
Added `@escaping @Sendable` (and `@MainActor` where the callback is always delivered on the main actor) to closure parameters in:

- `Requestable.start(_:)` and the async `start()` override
- `TokenRequestable.start(_:)` and the async `start()` override
- `TokenRequest.start(_:)` and `verifyClaims(idToken:callback:)`
- `BioAuthentication.validateBiometric(callback:)`
- `OAuth2Grant.credentials(from:callback:)` — also added `@MainActor` and wrapped all call sites in `Task { @MainActor in }` so the callback is always invoked on the main actor

### Combine publisher fixes in `CredentialsManager`
The shorthand `Future(self.start)` passes the `Future` promise closure directly as the `start` callback. In Swift 6 this is rejected because the promise is not `@Sendable`. Fixed by boxing the promise:

```swift
// Before
Future { callback in
    self.credentials(..., callback: callback)
}

// After
Future { promise in
    let box = SendableBox(value: promise)
    self.credentials(...) { @Sendable result in box.value(result) }
}

### CODE REVIEW

**Comment by sanchitmehtagit** [comment]
These changes doesnt impact API layer- no change needed on migration guide and examples.

File: Auth0/CredentialsManager.swift [hunk 1]
```diff
@@ -466,15 +467,15 @@ public struct CredentialsManager: Sendable {
...
+                                         parameters: params.asParameters,
                                          headers: headers,
                                          forceRenewal: false,
                                          callback: mainThreadCallback)
             }
         } else {
             self.retrieveCredentials(scope: scope,
                                      minTTL: minTTL,
-                                     parameters: parameters,
+                                     parameters: params.asParameters,
```
NandanPrabhu: do we need to use normalised parameters here? Understand the rationale behind using normalised params inside sendable closure

sanchitmehtagit: Good catch will fix this

NandanPrabhu (approved):

PR #1141, opened 2026-04-12
feat: Removes three sources of mutable global state that block Swift 6 strict concurrency

sanchitmehtagit:
## What this PR does

Removes three sources of mutable global state that block Swift 6 strict concurrency. Each change is independent and targets a different file.

### 1. `Auth0Log` — dependency injection

**Before:**
```swift
enum Auth0Log {
    static var loggingService: UnifiedLogging = OSUnifiedLoggingService()
}
```

**After:**
```swift
struct Auth0Log: Sendable {
    let loggingService: any UnifiedLogging
    init(loggingService: any UnifiedLogging = OSUnifiedLoggingService()) { ... }
    static let shared = Auth0Log()
}
```

`static var loggingService` was a mutable global — a Swift 6 hard error. Replaced with a `Sendable` struct and constructor-injected dependency. Production code uses `Auth0Log.shared`; tests inject their own instance without mutating shared state. Removes the `afterEach` reset block from `LoggerSpec`.

### 2. `DPoP.auth0Nonce` → `NonceStorage` (data race fix)

**Before:**
```swift
static private(set) var auth0Nonce: String?

static func storeNonce(_ nonce: String) {
    serialQueue.sync { auth0Nonce = nonce }
}
// generateProof reads Self.auth0Nonce directly — unprotected
```

**After:**
```swift
final class NonceStorage: @unchecked Sendable {
    private let queue = DispatchQueue(label: "com.auth0.sdk.nonce")
    private var _nonce: String?
    var nonce: String? { queue.sync { _nonce } }
    func store(_ nonce: String) { queue.async { _nonce = nonce } }
    func clear() { queue.async { _nonce = nil } }
}

static let nonceStorage = NonceStorage()
```

Writes went through `serialQueue.sync` but the read in `generateProof` was unprotected — a real data race under Swift 6 strict concurrency. `NonceStorage` protects every access uniformly via its own internal queue.

### 3. `ASProvider.completionHandler` — `@Sendable` closure type

**Before:**
```swift
static let completionHandler: (_ callback: @escaping WebAuthProviderCallback) -> ASHandler = { ... }
```

**After:**
```swift
static let completionHandler: @Sendable (_ callback: @escaping WebAuthProviderCallback) -> ASHandler = { ... }
```

The stored closure was not marked `@Sendable`, causing a strict concurrency warning when captured across isolation boundaries.

## Tests

- `LoggerSpec`: removed global `Auth0Log.loggingService` mutation; each test creates a local `Auth0Log(loggingService: mock)`
- `DPoPSpec`: updated `DPoP.auth0Nonce` references to `DPoP.nonceStorage.nonce`


subhankarmaiti:
@copilot

Auth0/DPoP/DPoP.swift
```diff
@@ -1,6 +1,17 @@
...
 import CryptoKit
 
+// MARK: - Nonce Storage
+
+final class NonceStorage: @unchecked Sendable {
+    private let queue = DispatchQueue(label: "com.auth0.sdk.nonce")
+    private var _nonce: String?
+
+    var nonce: String? { queue.sync { _nonce } }
+    func store(_ nonce: String) { queue.sync { _nonce = nonce } }
```
NandanPrabhu: lets use async for store and clear

NandanPrabhu (approved):

Auth0/DPoP/DPoP.swift
sanchitmehtagit: Good catch

PR #1145, opened 2026-04-15
feat: Add Error handling for DPoP thumbprint mismatch error

NandanPrabhu:

### 📋 Changes

This PR adds DPoP key validation to the `CredentialsManager`, ensuring that DPoP-bound credentials are only used when the matching key pair is available.

**New error codes in `CredentialsManagerError`:**

| Error | Description |
|-------|-------------|
| `dpopKeyMissing` | The DPoP key pair is no longer available in the Keychain. Stored credentials are cleared automatically. |
| `dpopKeyMismatch` | The current DPoP key pair does not match the one used when the credentials were saved. Stored credentials are cleared automatically. |
| `dpopNotConfigured` | The credentials are DPoP-bound but the `Authentication` client was not configured with `.useDPoP()`. |

**DPoP thumbprint storage:**

- The `CredentialsManager` now persists a SHA-256 hash of the DPoP key thumbprint (JKT) in device storage on every credential save (including `store(credentials:)` and after renewal/exchange).
- On credential retrieval, `validateDPoPState` checks that DPoP is properly configured, the key pair exists, and the hashed thumbprint matches. Mismatches trigger automatic credential clearing.
- The raw key thumbprint is never stored; only its SHA-256 hash is persisted.

**Affected flows:**

- `init(authentication: Authentication, storeKey: String, dpopThumbprintKey: String, storage: CredentialsStorage = SimpleKeychain(), maxRetries: Int = 0)` - CredentialsManager init will have dpopThumbprintKey parameter with a default value
- `credentials(withScope:minTTL:parameters:headers:callback:)` — validates DPoP state before renewal
- `ssoCredentials(parameters:headers:callback:)` — validates DPoP state before SSO exchange
- `apiCredentials(forAudience:scope:minTTL:parameters:headers:callback:)` — validates DPoP state before API token exchange
- `store(credentials:)` — saves the DPoP thumbprint hash alongside credentials
- `clear()` — removes the stored DPoP thumbprint hash

### 🎯 Testing

- 100% unit test coverage for all three new error codes across `CredentialsManagerSpec` and `CredentialsManagerErrorSpec`.
- Tests cover: missing key pair, mismatched key pair, DPoP not configured, thumbprint persistence and clearing, error message strings, and `Equatable` conformance.


Auth0/CredentialsManager.swift
```diff
@@ -710,6 +714,66 @@ public struct CredentialsManager: Sendable {
...
+        } catch {
+            hasKeyPair = false
+        }
+
+        if hasKeyPair == false {
+            _ = self.clear()
+            return .dpopKeyMissing
+        }
+
+        // Hash the current thumbprint to compare against the stored hash
```
pmathew92, line 795: Isn't the `dpop.jkt()` call returns a SHA hashed string . Do we need to hash it again ?

Auth0/CredentialsManager.swift
```diff
@@ -710,6 +714,66 @@ public struct CredentialsManager: Sendable {
...
+        let currentThumbprint = (try? dpop.jkt()).map { thumbprint in
+            SHA256.hash(data: Data(thumbprint.utf8))
+                .map { String(format: "%02x", $0) }
+                .joined()
+        }
+        if let stored = storedThumbprint, stored != currentThumbprint {
```
pmathew92: This doesn't handle the migration scenario when the stored thumbprint is null the first time. In this case ,if the current thumbprint is not null ,need to store the current thumbprint to the `dpopThumbprintKey`

pmathew92: Might not be required here as the `saveDPoPThumbprint` method saves thumbprint when a new token is renewed

Auth0/CredentialsManager.swift
```diff
@@ -710,6 +714,66 @@ public struct CredentialsManager: Sendable {
...
+        guard dpopUsed,
+              let dpop = self.authentication.dpop else {
+            _ = self.storage.deleteEntry(forKey: self.dpopThumbprintKey)
+            return
+        }
+
+        do {
+            let thumbprint = try dpop.jkt()
+            // Store a SHA-256 hash of the thumbprint to avoid persisting the raw key thumbprint in device storage
+            let hashedThumbprint = SHA256.hash(data: Data(thumbprint.utf8))
```
pmathew92: `dpop.jkt()` already returns a SHA hashed thumbprint. Do we need to hash it again ?

Auth0Tests/CredentialsManagerErrorSpec.swift
```diff
@@ -151,6 +151,66 @@ class CredentialsManagerErrorSpec: QuickSpec {
...
+                let error = CredentialsManagerError(code: .dpopNotConfigured)
+                expect(error ~= CredentialsManagerError.dpopNotConfigured) == true
+            }
+
+            it("dpopKeyMismatch should pattern match by code") {
+                let error = CredentialsManagerError(code: .dpopKeyMismatch)
+                expect(error ~= CredentialsManagerError.dpopKeyMismatch) == true
+            }
+
+            it("DPoP errors should not equal each other") {
```
pmathew92, line 214: What exactly is this test trying to validate ?

Auth0/CredentialsManager.swift
NandanPrabhu: good catch. hashing is already done while returning jkt. So removed explicit hashing here

Auth0/CredentialsManager.swift
NandanPrabhu: yes removed it

Auth0/CredentialsManager.swift
```diff
@@ -765,6 +769,60 @@ public struct CredentialsManager: Sendable {
...
+        if hasKeyPair == false {
+            try self.clear()
+            return .dpopKeyMissing
+        }
+
+        // Hash the current thumbprint to compare against the stored hash
+        let currentThumbprint = try dpop.jkt()
+        if let stored = storedThumbprint, stored != currentThumbprint {
+            try self.clear()
+            return .dpopKeyMismatch
```
sanchitmehtagit: We’ve made `validateDPoPState(for credentials: Credentials) throws -> CredentialsManagerError?`  throwable due to the try self.clear(), but it looks like the error isn’t being handled at the call site.
Wanted to check: if self.clear() fails, do we still intend to return .dpopKeyMismatch? In the current flow, it seems like it woud not return .dpopKeyMismatch if the throw happens on try self.clear()

EXAMPLES.md
```diff
@@ -1184,6 +1185,39 @@ webView.configuration.websiteDataStore.httpCookieStore.setCookie(cookie)
 > [!IMPORTANT]
 > Make sure the cookie's domain matches the Auth0 domain your *website* is using, regardless of the one your mobile app is using. Otherwise, the `/authorize` endpoint will not receive the cookie. If your website is using the default Auth0 domain (like `example.us.auth0.com`), set the cookie's domain to this value. On the other hand, if your website is using a custom domain, use this value instead.
 
+### DPoP error handling
+
+When using DPoP with the Credentials Manager, additional validation is performed on credential retrieval to ensure the DPoP key pair is consistent. The following errors may be returned:
```
sanchitmehtagit: lets add these errors as part of migration guide in error handling section

Auth0/CredentialsManager.swift
```diff
@@ -765,6 +769,60 @@ public struct CredentialsManager: Sendable {
         return try NSKeyedUnarchiver.unarchivedObject(ofClass: Credentials.self, from: data)
     }
 
+    private func validateDPoPState(for credentials: Credentials) throws -> CredentialsManagerError? {
```
sanchitmehtagit: Do we need to make this function `throwing` if callers aren’t catching errors? Could this instead be `private func validateDPoPState(for credentials: Credentials) -> CredentialsManagerError?`

Auth0/CredentialsManager.swift
```diff
@@ -765,6 +769,60 @@ public struct CredentialsManager: Sendable {
...
+            try self.clear()
+            return .dpopKeyMismatch
+        }
+
+        return nil
+    }
+
+    private func saveDPoPThumbprint(for credentials: Credentials) throws {
+        let dpopUsed = credentials.tokenType.caseInsensitiveCompare("DPoP") == .orderedSame
+        || self.authentication.dpop != nil
```
sanchitmehtagit: `self.authentication.dpop != nil`
this check can be removed  as we already have condition let check just below

>  let dpop = self.authentication.dpop else

pmathew92:
@NandanPrabhu  Both apiCredentials and ssoCredentials  flow should save and validate the Dpop thumbprint and key

Auth0/CredentialsManager.swift
```diff
@@ -765,6 +769,60 @@ public struct CredentialsManager: Sendable {
...
+            hasKeyPair = false
+        }
+
+        if hasKeyPair == false {
+            try self.clear()
+            return .dpopKeyMissing
+        }
+
+        // Hash the current thumbprint to compare against the stored hash
+        let currentThumbprint = try dpop.jkt()
```
pmathew92, line 796: What if `dpop.jkt()` throws an error ? Wouldn't this crash since we are not catchin it here ?

Auth0/CredentialsManager.swift
```diff
@@ -149,6 +150,7 @@ public struct CredentialsManager: Sendable {
         let data = try NSKeyedArchiver.archivedData(withRootObject: credentials,
                                                     requiringSecureCoding: true)
         try self.storage.setEntry(data, forKey: self.storeKey)
+        try saveDPoPThumbprint(for: credentials)
```
pmathew92: How are we handling the scenario when `saveDPoPThumbprint` throws error ?

Auth0/CredentialsManager.swift
```diff
@@ -149,6 +150,7 @@ public struct CredentialsManager: Sendable {
         let data = try NSKeyedArchiver.archivedData(withRootObject: credentials,
                                                     requiringSecureCoding: true)
         try self.storage.setEntry(data, forKey: self.storeKey)
```
pmathew92, line 155: Should we save the DPoP thumbprint only when `self.storage.setEntry(data, forKey: self.storeKey)` succeeds ?

pmathew92:
@NandanPrabhu  How are we handling the scenario where the user passes a key for the Dpop thumbprint storage ?

Auth0/CredentialsManager.swift
NandanPrabhu: thats already handled if setEntry fails then it wont execute past try statement

Auth0/CredentialsManager.swift
NandanPrabhu: any error from keychain storage will be bubbled up back to client

Auth0/CredentialsManager.swift
NandanPrabhu: this wont crash. statements are written sequentially. if dpop.jkt throws error. then the following statement wont execute and the error thrown will be bubbled up back to the client

Auth0/CredentialsManager.swift
NandanPrabhu: instead of returning error. made this function throw error. ultimately we want this error to bubbled up back to the client. and if we see retrieveCredentials or retrieveApiCredentials closely. we follow the same pattern

Auth0/CredentialsManager.swift
NandanPrabhu: the end goal is, error thrown in any step whether its clear() or dpop.jkt, the error has to be bubbled up in the correct state back to the client

Auth0/CredentialsManager.swift
NandanPrabhu: no if clear fails. customer will be getting error corresponding to why clear function failed

Auth0Tests/CredentialsManagerErrorSpec.swift
NandanPrabhu: Conformance of CredentialsManagerError to Equatable. The two are equal only when their code and localized Description matches

EXAMPLES.md
NandanPrabhu: done

Auth0/CredentialsManager.swift
NandanPrabhu: if we remove authentication.dpop != nil, then dpopUsed can be false if tokenType is not `DPoP` . If dpopUsed is false below check for extracting authentication dpop will fail even if authentication.dpop is not nil. So we need both the checks

sanchitmehtagit: This creates a slightly inconsistent condition:
dpopUsed can be true without the token being DPoP
But the guard still requires authentication.dpop to exist anyway
lmk if i miss something

Auth0/CredentialsManager.swift
```diff
@@ -210,6 +216,7 @@ public struct CredentialsManager: Sendable {
         self.biometricSession.lastBiometricAuthTime = self.biometricSession.noSession
         self.biometricSession.lock.unlock()
         #endif
+        try DPoP.clearKeypair()
```
sanchitmehtagit, line 219: What will happen for a non DPoP scenario if keypair doesnt exist , double check or make sure this should not throw any error

Auth0/CredentialsManager.swift
```diff
@@ -897,6 +959,7 @@ public struct CredentialsManager: Sendable {
                     complete()
                     return callback(.failure(.noRefreshToken))
                 }
+                try self.validateDPoPState(for: credentials)
```
sanchitmehtagit, line 961: nit: for code clarity we can move error handling here itself and remove from  error handling from line : _987_
```
do {
try self.validateDPoPState(for: credentials) 
} catch let error as CredentialsManagerError {
                complete()
                callback(.failure(error))
}
```

NandanPrabhu: that would be code duplication right

NandanPrabhu: its a good suggestion but didn't understand how it can improve over existing implementation. We dont need specific handling for dpop errors before propagating it back to client

sanchitmehtagit: we should remove handling from here : _987_ https://github.com/auth0/Auth0.swift/pull/1145/changes#diff-32cd1807ed1b4849472f72a2735e8f287ab45daf78ae0e5a8045093d6abb75a0R987  and keep the error handling close to where the failure occur. its easier to scan  i believe

Auth0/CredentialsManager.swift
NandanPrabhu: if keypair doesnt exist it doesnt throw error. for errSecItemNotFound case no error is thrown

Auth0/CredentialsManager.swift
sanchitmehtagit: > its a good suggestion but didn't understand how it can improve over existing implementation

This makes code easier to read and more maintainable

Auth0/CredentialsManager.swift
```diff
@@ -765,6 +774,54 @@ public struct CredentialsManager: Sendable {
...
+        let isDPoPBound = credentials.tokenType.caseInsensitiveCompare("DPoP") == .orderedSame
+            || storedThumbPrintValue != nil
+
+        guard isDPoPBound else { return }
+
+        guard let dpop = self.authentication.dpop else {
+            throw CredentialsManagerError.dpopNotConfigured
+        }
+
+        var hasKeyPair = try? dpop.hasKeypair()
```
sanchitmehtagit: could we review at the warning below
> variable 'hasKeyPair' was never mutated; consider changing to 'let' constant

V3_MIGRATION_GUIDE.md
```diff
@@ -420,6 +420,52 @@ class MyCustomStorage: CredentialsStorage {
 
 In v2, storage read failures were swallowed by `try?` — for example, `revoke()` returned `.success` when no credentials were stored and `getEntry(forKey:)` threw. In v3, the underlying storage error is propagated as the `cause` of the `CredentialsManagerError`, giving callers full context to respond appropriately.
 
+**DPoP validation errors:** When DPoP-bound credentials are stored, the Credentials Manager now validates the DPoP state before attempting renewal. The following new errors can be thrown by `credentials()`, `apiCredentials()`, and `ssoCredentials()`:
```
sanchitmehtagit, line 660: Could we please add a note `[!NOTE] `here that these errors are related for you  if you are using DPoP support and you must make sure to handle these errors when migrating to V3

V3_MIGRATION_GUIDE.md
```diff
@@ -420,6 +420,52 @@ class MyCustomStorage: CredentialsStorage {
...
+```swift
+credentialsManager.credentials { result in
+    switch result {
+    case .success(let credentials):
+        // use credentials
+        break
+    case .failure(let error):
+        switch error {
+        case .dpopNotConfigured:
+            // re-initialize CredentialsManager with DPoP-configured Authentication client
```
sanchitmehtagit: coudl you please fill this with example as well

V3_MIGRATION_GUIDE.md
```diff
@@ -420,6 +420,52 @@ class MyCustomStorage: CredentialsStorage {
...
+            break
+        case .dpopKeyMissing:
+            // DPoP key was deleted from Keychain — clear credentials and re-authenticate
+            try? credentialsManager.clear()
```
sanchitmehtagit: try? credentialsManager.clear()

do we need to do this we are already doing this in SDK

V3_MIGRATION_GUIDE.md
```diff
@@ -420,6 +420,52 @@ class MyCustomStorage: CredentialsStorage {
...
+            navigateToLogin()
+        case .dpopKeyMismatch:
+            // DPoP key changed — clear credentials and re-authenticate
+            try? credentialsManager.clear()
```
sanchitmehtagit: try? credentialsManager.clear()
do we need to do this we are already doing this in SDK

EXAMPLES.md
```diff
@@ -1211,6 +1212,39 @@ webView.configuration.websiteDataStore.httpCookieStore.setCookie(cookie)
...
+
+```swift
+credentialsManager.credentials { result in
+    switch result {
+    case .success(let credentials):
+        print("Obtained credentials: \(credentials)")
+    case .failure(let error):
+        switch error {
+        case .dpopNotConfigured:
+            print("DPoP is not configured on the Authentication client")
```
sanchitmehtagit: lets fill in these with proper error handling as we have in migration guid

V3_MIGRATION_GUIDE.md
NandanPrabhu: added

sanchitmehtagit (dismissed):

sanchitmehtagit (approved):

PR #1146, opened 2026-04-15
feat: integrate credentials manager with webauth

subhankarmaiti:
Integrates the `CredentialsManager` with the `WebAuth` client so that credentials are automatically stored after login and cleared after logout, resulting in a simpler DX for most customers.

## Usage

```swift
// Use a custom CredentialsManager (reuse for login + logout)
let webAuth = Auth0.webAuth()
    .useCredentialsManager(myCredentialsManager)

### CODE REVIEW

File: Auth0Tests/WebAuthSpec.swift [hunk 1]
```diff
@@ -1092,3 +1206,32 @@ class MockUserAgent: WebAuthUserAgent {
     }
 
 }
+
+class SpyCredentialsStorage: CredentialsStorage {
```
NandanPrabhu: @subhankarmaiti can we mark this as final class and also move the class to new file

Auth0/WebAuth.swift
```diff
@@ -208,6 +208,39 @@ public protocol WebAuth: SenderConstraining, Trackable, Loggable, Sendable {
     /// - ``WebAuthProvider``
     func provider(_ provider: @escaping WebAuthProvider) -> Self
 
+    /// The ``CredentialsManager`` used by this WebAuth client to automatically store credentials after login
+    /// and clear them after logout.
+    /// Returns `nil` if the credentials manager has been disabled.
+    ///
+    /// ## See Also
+    ///
+    /// - ``useCredentialsManager(_:)-8fx8``
```
NandanPrabhu: @subhankarmaiti think there is a typo here

sanchitmehtagit (commented):
lets update the migration guide/ examples

Auth0/WebAuth.swift
```diff
@@ -208,6 +208,39 @@ public protocol WebAuth: SenderConstraining, Trackable, Loggable, Sendable {
...
+    /// Enable or disable automatic credentials management. When enabled (the default), credentials are
+    /// automatically stored after login and cleared after logout. When disabled, no credentials will be
+    /// automatically stored or cleared.
+    ///
+    /// - Parameter enabled: Whether to use credentials management.
+    /// - Returns: The same `WebAuth` instance to allow method chaining.
+    ///
+    /// ## See Also
+    ///
+    /// - ``useCredentialsManager(_:)-8fx8``
```
NandanPrabhu: here too

V3_MIGRATION_GUIDE.md
```diff
@@ -421,13 +422,94 @@ credentialsManager.revoke { result in
...
+// v3 - use a custom CredentialsManager
+Auth0
+    .webAuth()
+    .useCredentialsManager(credentialsManager)
+    .start { result in
+        // ...
+    }
+
+// v3 - opt out of automatic credentials management
+Auth0
```
sanchitmehtagit, line 523: Should this be disabled by default and only enabled when needed? The reason is that customers migrating to v3 may already be handling CredentialManager themselves, and enabling this by default in v3 could lead to duplicate handling if they don’t remove their handling  after migrating to V3

NandanPrabhu (approved):

PR #1148, opened 2026-04-17
chore: update SDK version to Swift 6 for iOS, TVOS, WATCHOS, MACOS & VISIONOS

sanchitmehtagit:

### 📋 Changes

chore: update SDK version to Swift 6 for iOS, TVOS, WATCHOS, MACOS & VISIONOS

Before beta: we may need changes for swift package too. but that can be done separately


Auth0/WebViewProvider.swift
```diff
@@ -190,13 +190,12 @@ extension WebViewUserAgent: WKURLSchemeHandler {
 /// Handling of HTTPS callbacks.
 extension WebViewUserAgent: WKNavigationDelegate {
 
-    func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction, decisionHandler: @escaping (WKNavigationActionPolicy) -> Void) {
+    func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction) async -> WKNavigationActionPolicy {
```
sanchitmehtagit, line 193: using Swift 6 compatible newer callbacks for WKNavigationDelegate .
This fixes a long standing warning as well in the code too

Auth0/WebViewProvider.swift
```diff
@@ -72,7 +72,7 @@ final class WebViewUserAgent: NSObject, WebAuthUserAgent, Sendable {
     static let customSchemeRedirectionFailureMessage = "com.auth0.webview.redirection_failure"
     let defaultSchemesSupportedByWKWebview = ["https"]
 
-    let request: URLRequest
+    nonisolated let request: URLRequest
```
NandanPrabhu, line 75: why this needs to be marked as nonisolated?

sanchitmehtagit: Without nonisolated, accessing request from a non-@MainActor context would be a compile error. Marking it nonisolated opts that specific property out of the inferred actor isolation, since it's an immutable let of a Sendable type type that doesn't need that protection.

NandanPrabhu (approved):

PR #1172, opened 2026-05-06
chore: migrate RL scanner to shared devsecops-tooling action

pmathew92:
### 📋 Changes

Migrates the RL scanner from the repo-local workflow + composite action to the shared auth0/devsecops-tooling/.github/actions/rl-scan@main action.


.github/workflows/rl-scanner.yml
```diff
@@ -14,15 +14,8 @@ permissions:
 jobs:
   rl-scanner:
     name: Run Reversing Labs Scanner
-    if: github.event_name == 'workflow_dispatch' || (github.event_name == 'pull_request' && github.event.pull_request.merged && github.event.pull_request.base.ref == 'master' && startsWith(github.event.pull_request.head.ref, 'release/'))
+    if: github.event_name == 'workflow_dispatch' || (github.event_name == 'pull_request' && github.event.pull_request.merged && startsWith(github.event.pull_request.head.ref, 'release/'))
```
NandanPrabhu, line 17: this is only for release branches? previous one had master too

pmathew92: This should be triggered whenever we make a release. Don't we make the release whenever the release branch is merged ?

NandanPrabhu: previous action was triggered when base is master. now base condition is removed. Resolving the comment

NandanPrabhu (approved):

.github/workflows/rl-scanner.yml
pmathew92: Previoulsy we are triggering all releases via master only. But now we started doing beta from other branches. To accomodate that removed the base check

PR #1175, opened 2026-05-14
feat:Add partial support for PAR auth flow

pmathew92:
### 📋 Changes

adds support for Pushed Authorization Request (PAR) flows where the backend-for-frontend (BFF) handles the /par and /token endpoints while the SDK manages the browser-based authorization.

```
// Callback
  Auth0.authorizeWithRequestUri()
      .sessionTransferToken(sst) // optional
      .start(requestUri: requestUri) { result in
          switch result {
          case .success(let authCode):
              // Send authCode.code to BFF for token exchange
          case .failure(let error):
              // Handle error
          }
      }

  // Async/await
  let authCode = try await Auth0
      .authorizeWithRequestUri()
      .start(requestUri: requestUri)

```

### DIFF
--- a/Auth0/Auth0.swift
+++ b/Auth0/Auth0.swift
@@ -44,7 +44,7 @@ public let defaultScope = "openid profile email"
  ```swift
  Auth0.authentication(clientId: "client-id", domain: "samples.us.auth0.com")
  ```
-
+ 
  - Parameters:
    - clientId: Client ID of your Auth0 application.
    - domain:   Domain of your Auth0 account, for example `samples.us.auth0.com`.
@@ -290,6 +290,48 @@ public func webAuth(session: URLSession = .shared, bundle: Bundle = Bundle.main)
  */
 public func webAuth(clientId: String, domain: String, session: URLSession = .shared) -> WebAuth {
     return Auth0WebAuth(clientId: clientId, url: .httpsURL(from: domain), session: session)
+}
+
+/**
+ PAR (Pushed Authorization Request) client for performing the browser authorization step of a PAR flow.
+
+ Uses the `ClientId` and `Domain` values from the `Auth0.plist` file in your main bundle.
+
+ ## Usage
+
+ ```swift
+ let authCode = try await Auth0
+     .authorizeWithRequestUri()
+     .start(requestUri: requestUri)
+ ```
+
+ - Parameter bundle: Bundle used to locate the `Auth0.plist` file. Defaults to `Bundle.main`.
+ - Returns: PAR Web Auth client.
+ - Warning: Calling this method without a valid `Auth0.plist` file will crash your application.
+ */
+public func authorizeWithRequestUri(bundle: Bundle = Bundle.main) -> PARWebAuth {
+    let values = plistValues(bundle: bundle)!
+    return authorizeWithRequestUri(clientId: values.clientId, domain: values.domain)
+}
+
+/**
+ PAR (Pushed Authorization Request) client for performing the browser authorization step of a PAR flow.
+
+ ## Usage
+
+ ```swift
+ let authCode = try await Auth0
+     .authorizeWithRequestUri(clientId: "client-id", domain: "samples.us.auth0.com")
+     .start(requestUri: requestUri)
+ ```
+
+ - Parameters:
+   - clientId: Client ID of your Auth0 application.
+   - domain:   Domain of your Auth0 account, for example `samples.us.auth0.com`.
+ - Returns: PAR Web Auth client.
+ */
+public func authorizeWithRequestUri(clientId: String, domain: String) -> PARWebAuth {
+    return PARWebAuth(clientId: clientId, url: .httpsURL(from: domain))
 }
 #endif
 
--- a/Auth0/Auth0.swift
+++ b/Auth0/Auth0.swift
@@ -309,7 +309,7 @@ public func webAuth(clientId: String, domain: String, session: URLSession = .sha
  - Returns: PAR Web Auth client.
  - Warning: Calling this method without a valid `Auth0.plist` file will crash your application.
  */
-public func authorizeWithRequestUri(bundle: Bundle = Bundle.main) -> PARWebAuth {
+public func authorizeWithRequestUri(bundle: Bundle = Bundle.main) -> PARAuth {
     let values = plistValues(bundle: bundle)!
     return authorizeWithRequestUri(clientId: values.clientId, domain: values.domain)
 }
@@ -330,7 +330,7 @@ public func authorizeWithRequestUri(bundle: Bundle = Bundle.main) -> PARWebAuth 
    - domain:   Domain of your Auth0 account, for example `samples.us.auth0.com`.
  - Returns: PAR Web Auth client.
  */
-public func authorizeWithRequestUri(clientId: String, domain: String) -> PARWebAuth {
+public func authorizeWithRequestUri(clientId: String, domain: String) -> PARAuth {
     return PARWebAuth(clientId: clientId, url: .httpsURL(from: domain))
 }
 #endif
--- a/Auth0/AuthorizationCode.swift
+++ b/Auth0/AuthorizationCode.swift
@@ -1,0 +1,32 @@
+#if WEB_AUTH_PLATFORM
+import Foundation
+
+/// The result of a PAR (Pushed Authorization Request) authorization flow.
+///
+/// Contains the authorization code returned by the `/authorize` endpoint, which should be sent
+/// to your backend (BFF) for token exchange.
+///
+/// ## See Also
+///
+/// - ``PARWebAuth``
+public struct AuthorizationCode {
+
+    /// The authorization code returned by the `/authorize` endpoint.
+    public let code: String
+
+    /// The state parameter returned in the redirect, if present.
+    /// This is the state that was originally sent by your backend to the `/oauth/par` endpoint.
+    public let state: String?
+
+    /// Creates a new `AuthorizationCode` instance.
+    ///
+    /// - Parameters:
+    ///   - code: The authorization code.
+    ///   - state: The state parameter, if any.
+    public init(code: String, state: String? = nil) {
+        self.code = code
+        self.state = state
+    }
+
+}
+#endif
--- a/Auth0/AuthorizationCode.swift
+++ b/Auth0/AuthorizationCode.swift
@@ -9,7 +9,7 @@ import Foundation
 /// ## See Also
 ///
 /// - ``PARWebAuth``
-public struct AuthorizationCode {
+public struct AuthorizationCode: Sendable {
 
     /// The authorization code returned by the `/authorize` endpoint.
     public let code: String
--- a/Auth0/PARAuth.swift
+++ b/Auth0/PARAuth.swift
@@ -13,7 +13,7 @@ import Combine
 /// - ``AuthorizationCode``
 /// - ``WebAuthError``
 /// - ``PARWebAuth``
-public protocol PARAuth: Trackable, Sendable {
+public protocol PARAuth: Trackable, Loggable, Sendable {
 
     /// The Auth0 Client ID.
     var clientId: String { get }
--- a/Auth0/PARAuth.swift
+++ b/Auth0/PARAuth.swift
@@ -1,0 +1,72 @@
+#if WEB_AUTH_PLATFORM
+import Foundation
+import Combine
+
+/// Web-based PAR (Pushed Authorization Request) authorization.
+///
+/// Handles the browser authorization step of a PAR flow — opens the `/authorize` endpoint
+/// with a `request_uri` obtained from your backend's PAR endpoint call, and returns the
+/// authorization code for your backend to exchange for tokens.
+///
+/// ## See Also
+///
+/// - ``AuthorizationCode``
+/// - ``WebAuthError``
+/// - ``PARWebAuth``
+public protocol PARAuth: Trackable, Sendable {
+
+    /// The Auth0 Client ID.
+    var clientId: String { get }
+
+    /// The Auth0 Domain URL.
+    var url: URL { get }
+
+    // MARK: - Builder Methods
+
+    /// Provide a session transfer token to be passed as a query parameter to the `/authorize` endpoint.
+    /// This enables web single sign-on by transferring an existing session to the browser.
+    ///
+    /// - Parameter token: The session transfer token obtained from ``Authentication/ssoExchange(refreshToken:parameters:headers:)``.
+    /// - Returns: The same instance to allow method chaining.
+    func sessionTransferToken(_ token: String) -> Self
+
+    /// Specify a custom ``WebAuthProvider`` to handle the browser session.
+    ///
+    /// - Parameter provider: A custom provider.
+    /// - Returns: The same instance to allow method chaining.
+    func provider(_ provider: @escaping WebAuthProvider) -> Self
+
+    /// Use a private browser session to avoid storing the session cookie in the shared cookie jar.
+    ///
+    /// - Returns: The same instance to allow method chaining.
+    func useEphemeralSession() -> Self
+
+    // MARK: - Start
+
+    /// Start the PAR authorization flow using a `request_uri` from a PAR response.
+    /// Opens the browser with the authorize URL and returns the authorization code
+    /// for the app to exchange via BFF.
+    ///
+    /// - Parameters:
+    ///   - requestUri: The `request_uri` obtained from the PAR endpoint (must start with `urn:ietf:params:oauth:request_uri:`).
+    ///   - callback: Callback with the authorization code result. Always called on the main thread.
+    func start(requestUri: String, callback: @escaping @Sendable @MainActor (WebAuthResult<AuthorizationCode>) -> Void)
+
+    #if canImport(_Concurrency)
+    /// Start the PAR authorization flow using async/await.
+    ///
+    /// - Parameter requestUri: The `request_uri` obtained from the PAR endpoint.
+    /// - Returns: An ``AuthorizationCode`` containing the authorization code.
+    /// - Throws: A ``WebAuthError`` if the operation fails.
+    @MainActor
+    func start(requestUri: String) async throws -> AuthorizationCode
+    #endif
+
+    /// Start the PAR authorization flow as a Combine publisher.
+    ///
+    /// - Parameter requestUri: The `request_uri` obtained from the PAR endpoint.
+    /// - Returns: A publisher that emits an ``AuthorizationCode`` or a ``WebAuthError``.
+    func start(requestUri: String) -> AnyPublisher<AuthorizationCode, WebAuthError>
+
+}
+#endif
--- a/Auth0/PARTransaction.swift
+++ b/Auth0/PARTransaction.swift
@@ -36,17 +36,20 @@ class PARTransaction: NSObject, AuthTransaction {
             result[item.name] = item.value
         }
 
-        if let error = items["error"] {
-            let description = items["error_description"] ?? "Unknown error"
+        if items["error"] != nil {
             let cause = AuthenticationError(info: items, statusCode: 302)
-            self.finishUserAgent(with: .failure(WebAuthError(code: .other, cause: cause)))
+            let error = WebAuthError(code: .other, cause: cause)
+            self.finishUserAgent(with: .failure(error))
+            self.callback(.failure(error))
         } else if let code = items["code"] {
             self.finishUserAgent(with: .success(()))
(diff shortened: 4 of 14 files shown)


Auth0/PARTransaction.swift
```diff
@@ -0,0 +1,61 @@
+#if WEB_AUTH_PLATFORM
+import Foundation
+
+/// Transaction that handles the PAR authorization flow callback.
+/// Parses the redirect URL for the authorization code and optional state.
+class PARTransaction: NSObject, AuthTransaction {
```
NandanPrabhu: Lets make PARTransaction struct. Any specific reason we want it to be a class

Auth0/PARWebAuth.swift
```diff
@@ -0,0 +1,218 @@
...
+///     .sessionTransferToken(token)
+///     .start(requestUri: requestUri)
+/// // Send authCode.code to your BFF for token exchange
+/// ```
+///
+/// ## See Also
+///
+/// - ``AuthorizationCode``
+/// - ``WebAuthError``
+public class PARWebAuth {
```
NandanPrabhu: struct instead of classes

pmathew92: Followed the same approach used in `Auth0WebAuth` which is a public final class

Auth0/PARTransaction.swift
pmathew92: The existing `LoginTransaction` is a class. Followed the same approach

Auth0/PARWebAuth.swift
NandanPrabhu: makes sense. we might need to change the implementation when merging to v3 branch

Auth0/PARTransaction.swift
NandanPrabhu: lets make it final class. This would become struct once we merge to v3 branch

Auth0/PARTransaction.swift
```diff
@@ -0,0 +1,64 @@
...
+
+    typealias FinishTransaction = (WebAuthResult<AuthorizationCode>) -> Void
+
+    private(set) var userAgent: WebAuthUserAgent?
```
NandanPrabhu, line 10:
```suggestion
    private(set) weak var userAgent: WebAuthUserAgent?
```

Auth0/PARWebAuth.swift
NandanPrabhu: plz conform PARWebAuth to WebAuth protocol

NandanPrabhu:
```suggestion
public final class PARWebAuth {
```

Auth0/PARUtils.swift
```diff
@@ -0,0 +1,44 @@
+#if WEB_AUTH_PLATFORM
+import Foundation
+
+/// Shared utilities for PAR (Pushed Authorization Request) flows.
+enum PARUtils {
```
NandanPrabhu: any specific reason we have it another enum. Suggestion would be to have it as part of PARWebAuth class extension if adding it in the original class might bloat the number of functions it contains in one file
extension PARWebAuth {

}

Auth0/PARWebAuth.swift
```diff
@@ -0,0 +1,218 @@
...
+            return callback(.failure(WebAuthError(code: .invalidRequestUri(requestUri))))
+        }
+
+        guard let redirectURL = self.redirectURL else {
+            barrier.lower()
+            return callback(.failure(WebAuthError(code: .noBundleIdentifier)))
+        }
+
+        var additionalParameters: [String: String] = [:]
+        if let sst = sessionTransferTokenValue {
```
NandanPrabhu:
```suggestion
        if let sessionTransferTokenValue {
                    additionalParameters["session_transfer_token"] = sessionTransferTokenValue
        }
```

Auth0/Auth0.swift
```diff
@@ -291,6 +291,48 @@ public func webAuth(session: URLSession = .shared, bundle: Bundle = Bundle.main)
...
+ let authCode = try await Auth0
+     .authorizeWithRequestUri()
+     .start(requestUri: requestUri)
+ ```
+
+ - Parameter bundle: Bundle used to locate the `Auth0.plist` file. Defaults to `Bundle.main`.
+ - Returns: PAR Web Auth client.
+ - Warning: Calling this method without a valid `Auth0.plist` file will crash your application.
+ */
+public func authorizeWithRequestUri(bundle: Bundle = Bundle.main) -> PARWebAuth {
```
NandanPrabhu: lets ensure this returns protocol type instead of concrete type(class, enum, struct)

Auth0/Auth0.swift
```diff
@@ -291,6 +291,48 @@ public func webAuth(session: URLSession = .shared, bundle: Bundle = Bundle.main)
...
+     .authorizeWithRequestUri(clientId: "client-id", domain: "samples.us.auth0.com")
+     .start(requestUri: requestUri)
+ ```
+
+ - Parameters:
+   - clientId: Client ID of your Auth0 application.
+   - domain:   Domain of your Auth0 account, for example `samples.us.auth0.com`.
+ - Returns: PAR Web Auth client.
+ */
+public func authorizeWithRequestUri(clientId: String, domain: String) -> PARWebAuth {
```
NandanPrabhu: same as https://github.com/auth0/Auth0.swift/pull/1175/changes#r3247874251

Auth0/PARWebAuth.swift
pmathew92: the `start` methods in `WebAuth` is defined to return instance of `Credentials` where as this returns an instance of `AuthorizationCode` . Should we create a new protocol for this ?

Auth0/PARTransaction.swift
pmathew92: Will this compile since `WebAuthUserAgent` is not bound to `AnyObject` ? `LoginTransaction` also doesn't use weak type for userAgent

Auth0/PARUtils.swift
pmathew92: Done

Auth0/PARTransaction.swift
NandanPrabhu: let's not have this suggestion applied for now. conforming a protocol to AnyObject means the protocol will be implemented only by a reference type like class or actor. For now all the user agents are class types. But yes need to evaluate whether these user agents can be migrated to value types

Auth0/PARTransaction.swift
```diff
@@ -0,0 +1,64 @@
+#if WEB_AUTH_PLATFORM
+import Foundation
+
+/// Transaction that handles the PAR authorization flow callback.
+/// Parses the redirect URL for the authorization code and optional state.
+final class PARTransaction: NSObject, AuthTransaction {
```
NandanPrabhu: @pmathew92 Lets remove NSObject inheritance and have weak var added for the useragent. Lets make WebUserAgent conform to AnyObject
```suggestion
final class PARTransaction: AuthTransaction {

    typealias FinishTransaction = (WebAuthResult<AuthorizationCode>) -> Void

    private(set) weak var userAgent: WebAuthUserAgent?

    let redirectURL: URL
    let callback: FinishTransaction

    init(redirectURL: URL,
         userAgent: WebAuthUserAgent,
         callback: @escaping FinishTransaction) {
        self.redirectURL = redirectURL
        self.userAgent = userAgent
        self.callback = callback
    }
```

Auth0/PARTransaction.swift
```diff
@@ -0,0 +1,64 @@
...
+        if items["error"] != nil {
+            let cause = AuthenticationError(info: items, statusCode: 302)
+            let error = WebAuthError(code: .other, cause: cause)
+            self.finishUserAgent(with: .failure(error))
+            self.callback(.failure(error))
+        } else if let code = items["code"] {
+            self.finishUserAgent(with: .success(()))
+            let authorizationCode = AuthorizationCode(code: code, state: items["state"])
+            self.callback(.success(authorizationCode))
+            return true
```
NandanPrabhu: this true is not needed as we are returning at the end

Auth0/PARTransaction.swift
pmathew92: > Lets make WebUserAgent conform to AnyObject

Wouldn't that be a breaking change ?

Auth0/PARUtils.swift
```diff
@@ -0,0 +1,43 @@
...
+    ///   - requestUri: The request_uri from the PAR endpoint response.
+    ///   - additionalParameters: Extra query parameters to append (e.g., session_transfer_token).
+    /// - Returns: The fully constructed authorize URL.
+    static func buildAuthorizeURL(baseURL: URL,
+                                  clientId: String,
+                                  requestUri: String,
+                                  additionalParameters: [String: String] = [:]) -> URL {
+        let authorizeURL = baseURL.appendingPathComponent("authorize")
+        var components = URLComponents(url: authorizeURL, resolvingAgainstBaseURL: true)!
+        var queryItems = [
```
NandanPrabhu:
```suggestion
        var queryItems: [String: String] = [:]
        queryItems["client_id"] =  clientId
        queryItems["request_uri"] = requestUri
        additionalParameters.forEach { (key, value) in
            queryItems[key] = value
        }
        components.queryItems = queryItems.map { URLQueryItem(name: $0.key, value: $0.value)}
```

Auth0/PARWebAuth.swift
```diff
@@ -0,0 +1,218 @@
...
+/// // Send authCode.code to your BFF for token exchange
+/// ```
+///
+/// ## See Also
+///
+/// - ``AuthorizationCode``
+/// - ``WebAuthError``
+public final class PARWebAuth {
+
+    let clientId: String
```
NandanPrabhu: can we make these properties private

pmathew92: Auth0WebAuth uses the same internal access level for clientId/url, and tests assert on these properties via @testable import. Making them private breaks tests.

Auth0/PARWebAuth.swift
```diff
@@ -0,0 +1,218 @@
...
+    ///   - clientId: The Auth0 client ID.
+    ///   - url: The Auth0 domain URL.
+    public init(clientId: String, url: URL) {
+        self.clientId = clientId
+        self.url = url
+        self.storage = TransactionStore.shared
+        self.barrier = QueueBarrier.shared
+    }
+
+    init(clientId: String,
```
NandanPrabhu, line 79: will this init be used just for internal purpose?
 as it takes more params than public init, assume it can be consumed from outside the SDK

pmathew92: This init is used only for testing purposes . It's already internal  so it's not accessible outside the SDK. This follows the same pattern used in Auth0WebAuth.

Auth0/PARWebAuth.swift
```diff
@@ -0,0 +1,218 @@
...
+///     .sessionTransferToken(token)
+///     .start(requestUri: requestUri)
+/// // Send authCode.code to your BFF for token exchange
+/// ```
+///
+/// ## See Also
+///
+/// - ``AuthorizationCode``
+/// - ``WebAuthError``
+public final class PARWebAuth {
```
sanchitmehtagit: I’d suggest keeping `PARWebAuth` as a `struct`

`Auth0WebAuth` is a class on master and `struct` on V3 for swift 6 compliance,  but if we make `PARWebAuth` a class in `master`, we’ll run into two issues:

1. We’ll need to redo `PARWebAuth` again while merging into `develop/v3`.
2. As its a new change if we merge this as class in master and then on v3 branch we will need to convert it back to struct and which will further add breaking changes for v3  migration

Auth0/AuthorizationCode.swift
```diff
@@ -0,0 +1,32 @@
...
+
+/// The result of a PAR (Pushed Authorization Request) authorization flow.
+///
+/// Contains the authorization code returned by the `/authorize` endpoint, which should be sent
+/// to your backend (BFF) for token exchange.
+///
+/// ## See Also
+///
+/// - ``PARWebAuth``
+public struct AuthorizationCode {
```
sanchitmehtagit: conform to Sendable

Auth0/PARWebAuth.swift
```diff
@@ -0,0 +1,218 @@
...
+    // MARK: - Start (Callback)
+
+    /// Start the PAR authorization flow using a `request_uri` from a PAR response.
+    /// Opens the browser with the authorize URL and returns the authorization code
+    /// for the app to exchange via BFF.
+    ///
+    /// - Parameters:
+    ///   - requestUri: The `request_uri` obtained from the PAR endpoint (must start with `urn:ietf:params:oauth:request_uri:`).
+    ///   - callback: Callback with the authorization code result.
+    public func start(requestUri: String, callback: @escaping (WebAuthResult<AuthorizationCode>) -> Void) {
```
sanchitmehtagit: https://github.com/auth0/Auth0.swift/blob/2261bedce06abbb93ad4841b60179e6cb3ea1cef/Auth0/Auth0WebAuth.swift#L201-L205

We should follow the same pattern here in Swift v3 and update the API signature so the callback is always returned on `MainActor`.

Auth0/PARWebAuth.swift
```diff
@@ -0,0 +1,218 @@
...
+
+#if canImport(_Concurrency)
+public extension PARWebAuth {
+
+    /// Start the PAR authorization flow using async/await.
+    ///
+    /// - Parameter requestUri: The `request_uri` obtained from the PAR endpoint.
+    /// - Returns: An ``AuthorizationCode`` containing the authorization code.
+    /// - Throws: A ``WebAuthError`` if the operation fails.
+    func start(requestUri: String) async throws -> AuthorizationCode {
```
sanchitmehtagit, line 224: https://github.com/auth0/Auth0.swift/blob/2261bedce06abbb93ad4841b60179e6cb3ea1cef/Auth0/Auth0WebAuth.swift#L460

Annotate API signature to MainActor same as above pattern

Auth0/PARWebAuth.swift
sanchitmehtagit: Let’s also conform this to a protocol, similar to`WebAuth`, which would in turn conform to `Sendable`.
https://github.com/auth0/Auth0.swift/blob/develop/v3.0/Auth0/WebAuth.swift

Auth0/PARWebAuth.swift
```diff
@@ -0,0 +1,218 @@
...
+
+public extension PARWebAuth {
+
+    /// Start the PAR authorization flow as a Combine publisher.
+    ///
+    /// - Parameter requestUri: The `request_uri` obtained from the PAR endpoint.
+    /// - Returns: A publisher that emits an ``AuthorizationCode`` or a ``WebAuthError``.
+    func start(requestUri: String) -> AnyPublisher<AuthorizationCode, WebAuthError> {
+        return Deferred {
+            Future { [weak self] callback in
```
sanchitmehtagit: https://github.com/auth0/Auth0.swift/blob/4a70795511f90a399bf4a1418d557195f278a26c/Auth0/Auth0WebAuth.swift#L339 

Lets use strong capture here and not [weak self] similar to above implementation. If self is nil, callback is never called and the Future never completes, leaks the subscriber

Auth0/PARWebAuth.swift
sanchitmehtagit: public protocol ParWebAuth: SenderConstraining, Trackable, Loggable, Sendable {
in case we need to wire telemtery in future we should do it now otherwise it will be a breaking change since we will  have to conform ParWebAuth protocol to Trackable later if not now

Auth0/PARTransaction.swift
```diff
@@ -0,0 +1,63 @@
+#if WEB_AUTH_PLATFORM
+import Foundation
+
+/// Transaction that handles the PAR authorization flow callback.
+/// Parses the redirect URL for the authorization code and optional state.
+final class PARTransaction: NSObject, AuthTransaction {
+
+    typealias FinishTransaction = (WebAuthResult<AuthorizationCode>) -> Void
```
sanchitmehtagit: https://github.com/auth0/Auth0.swift/blob/2261bedce06abbb93ad4841b60179e6cb3ea1cef/Auth0/LoginTransaction.swift#L6

Nit: just check whether this `typealias` can also be made similar to this one, with `@MainActor` and `Sendable` conformance. If it doesn’t compile, it’s fine.

Auth0/PARWebAuth.swift
NandanPrabhu: yes. we need to

Auth0/PARAuth.swift
```diff
@@ -0,0 +1,72 @@
...
+/// Handles the browser authorization step of a PAR flow — opens the `/authorize` endpoint
+/// with a `request_uri` obtained from your backend's PAR endpoint call, and returns the
+/// authorization code for your backend to exchange for tokens.
+///
+/// ## See Also
+///
+/// - ``AuthorizationCode``
+/// - ``WebAuthError``
+/// - ``PARWebAuth``
+public protocol PARAuth: Trackable, Sendable {
```
sanchitmehtagit: Lets add Loggable as well even if we are not using logger now we can easily add logging in future with this

sanchitmehtagit (commented):
few more changes, rest LGTM

Auth0/PARUtils.swift
```diff
@@ -0,0 +1,45 @@
...
+        let authorizeURL = baseURL.appendingPathComponent("authorize")
+        var components = URLComponents(url: authorizeURL, resolvingAgainstBaseURL: true)!
+        var queryItems: [String: String] = [:]
+        queryItems["client_id"] = clientId
+        queryItems["request_uri"] = requestUri
+        additionalParameters.forEach { (key, value) in
+            queryItems[key] = value
+        }
+        let items = queryItems.map { URLQueryItem(name: $0.key, value: $0.value) }
+        components.queryItems = telemetry.queryItemsWithTelemetry(queryItems: items)
```
sanchitmehtagit, line 40: we need to use auth0ClientInfo to call this API similar to what we have done in Auth0WebAuth

Auth0/PARWebAuth.swift
```diff
@@ -0,0 +1,232 @@
...
+/// ## See Also
+///
+/// - ``AuthorizationCode``
+/// - ``PARAuth``
+/// - ``WebAuthError``
+public struct PARWebAuth: PARAuth, @unchecked Sendable {
+
+    public let clientId: String
+    public let url: URL
+    public var telemetry = Telemetry()
```
sanchitmehtagit, line 38: lets use this var auth0ClientInfo: Auth0ClientInfo and follow same pattern as in Auth0WebAuth

Auth0/PARAuth.swift
pmathew92: Added

Auth0/PARWebAuth.swift
NandanPrabhu: we need to define a new protocol something PARWebAuthDelegate

Auth0/PARUtils.swift
pmathew92: Auth0ClientInfo exists in v3. On master, Telemetry serves the same purpose — telemetry.queryItemsWithTelemetry() appends the auth0Client query param identically. We're already
  using Telemetry here, which matches the Auth0WebAuth pattern on master

Auth0/PARWebAuth.swift
pmathew92: Auth0ClientInfo doesn't exist on master. PARWebAuth already has public var telemetry = Telemetry() which is the exact same pattern as Auth0WebAuth on master

Auth0/PARTransaction.swift
NandanPrabhu: conforming WebAuthUserAgent to AnyObject would be a breaking change. Lets keep it as it is @pmathew92

Auth0/PARTransaction.swift
NandanPrabhu: yes conforming WebAuthUserAgent to AnyObject is a breaking change for developers who have values types implementing WebAuthUserAgent protocol

NandanPrabhu: but NSObject inheritance can be removed

Auth0/PARWebAuth.swift
NandanPrabhu: A new PR for the same rebased with develop/v3.0 would be required to ensure v3 beta has PAR support @pmathew92 . I think we can address these comments related to syncing to v3 in that PR

sanchitmehtagit (approved):

NandanPrabhu (approved):

PR #1190, opened 2026-05-29
feat: CTE delegation and impersonation support

NandanPrabhu:

### 📋 Changes
Adds delegation and impersonation support for Custom Token Exchange (CTE) per RFC 8693, allowing an actor (e.g. an AI agent or support rep) to act on behalf of a user.

**Types added:**
- `ActorToken` — A struct that bundles `token` and `tokenType` together, ensuring both are always provided when performing delegation/impersonation token exchanges (per RFC 8693).
- `ActClaim` — A `final class` representing the `act` (actor) claim from an ID token. Supports recursive nesting for multi-hop delegation chains. Exposes `sub`, nested `act`, and `additionalClaims`.

**Methods changed:**
- `Authentication.customTokenExchange(...)` — Added optional `actorToken: ActorToken?` parameter. When provided, `actor_token` and `actor_token_type` are sent in the token exchange request.

**Properties added:**
- `UserInfo.act: ActClaim?` — Parses the `act` claim from the ID token JSON body. Also added `"act"` to `UserInfo.publicClaims` so it is excluded from `customClaims`.

**Documentation:**
- Updated `EXAMPLES.md` with sections on actor token delegation and reading the `act` claim.
- Updated `README.md` to mention custom token exchange in the comprehensive guides note.

### 📎 References

- [RFC 8693: OAuth 2.0 Token Exchange](https://tools.ietf.org/html/rfc8693)
- [RFC 8693 Section 4.1: `act` Claim](https://tools.ietf.org/html/rfc8693#section-4.1)
- [Auth0 Custom Token Exchange Documentation](https://auth0.com/docs/authenticate/custom-token-exchange)

### 🎯 Testing

- Unit tests added in `AuthenticationSpec` for actor token parameter in CTE requests.
- Unit tests added in `UserInfoSpec` for `act` claim parsing: nil when absent, basic parsing, nested delegation chains, and exclusion from custom claims.
- Run `swift test` to verify all tests pass.

## Summary by CodeRabbit

* **New Features**
  * Support for token-exchange delegation/impersonation via actor tokens (new ActorToken type and an Authentication API overload accepting an actor token)
  * User info now exposes and parses the OIDC "act" (actor) claim, including nested delegation chains (ActClaim)

* **Documentation**
  * Added examples and guides for delegation workflows, actor token usage, and reading the `act` claim from stored user info
  * Note: providing an actor token can affect refresh token issuance

* **Tests**
  * Added tests covering actor-token request parameters and `act` claim parsing/edge cases

### DIFF
--- a/Auth0/ActClaim.swift
+++ b/Auth0/ActClaim.swift
@@ -1,0 +1,60 @@
+import Foundation
+
+/// Represents the `act` (actor) claim from an ID token issued during a token exchange delegation flow.
+///
+/// The `act` claim identifies the acting party — the entity performing actions on behalf of the subject. It is set
+/// server-side via Auth0 Actions using the `api.authentication.setActor()` command.
+///
+/// The claim may be nested to represent delegation chains (e.g., `act.act` for multi-hop delegation).
+///
+/// ## Usage
+///
+/// ```swift
+/// if let act = credentialsManager.user?.act {
+///     print("Actor: \(act.sub)")
+///     // Check for delegation chain
+///     if let innerAct = act.act {
+///         print("Original actor: \(innerAct.sub)")
+///     }
+/// }
+/// ```
+///
+/// ## See Also
+///
+/// - [RFC 8693: OAuth 2.0 Token Exchange - act Claim](https://tools.ietf.org/html/rfc8693#section-4.1)
+/// - [Custom Token Exchange Documentation](https://auth0.com/docs/authenticate/custom-token-exchange)
+public final class ActClaim: Sendable {
+
+    /// The subject identifier of the acting party.
+    public let sub: String?
+
+    /// A nested `act` claim representing the next actor in a delegation chain.
+    public let act: ActClaim?
+
+    /// Any additional claims beyond `sub` and `act` (e.g., `org`, `role`).
+    public let additionalClaims: [String: String]
+
+    /// Creates a new `ActClaim` from a JSON dictionary.
+    ///
+    /// - Parameter json: A dictionary representing the `act` claim from a decoded JWT.
+    /// - Returns: An `ActClaim` instance, or `nil` if the dictionary is empty.
+    public init?(json: [String: Any]) {
+        guard !json.isEmpty else { return nil }
+
+        self.sub = json["sub"] as? String
+
+        if let nestedAct = json["act"] as? [String: Any] {
+            self.act = ActClaim(json: nestedAct)
+        } else {
+            self.act = nil
+        }
+
+        var additional: [String: String] = [:]
+        for (key, value) in json where key != "sub" && key != "act" {
+            if let stringValue = value as? String {
+                additional[key] = stringValue
+            }
+        }
+        self.additionalClaims = additional
+    }
+}
--- a/Auth0/ActClaim.swift
+++ b/Auth0/ActClaim.swift
@@ -23,7 +23,7 @@ import Foundation
 ///
 /// - [RFC 8693: OAuth 2.0 Token Exchange - act Claim](https://tools.ietf.org/html/rfc8693#section-4.4)
 /// - [Custom Token Exchange Documentation](https://auth0.com/docs/authenticate/custom-token-exchange)
-public final class ActClaim: @unchecked Sendable {
+public final class ActClaim: Sendable {
 
     /// The subject identifier of the acting party.
     ///
@@ -38,7 +38,7 @@ public final class ActClaim: @unchecked Sendable {
     ///
     /// Values are preserved as-is, including non-string JSON values (numbers, booleans, objects, arrays), so that
     /// custom claims set via `api.authentication.setActor()` are not lost.
-    public let additionalClaims: [String: Any]
+    public let additionalClaims: [String: any Sendable]
 
     /// Creates a new `ActClaim` from a JSON dictionary.
     ///
--- a/Auth0/ActClaim.swift
+++ b/Auth0/ActClaim.swift
@@ -21,27 +21,32 @@ import Foundation
 ///
 /// ## See Also
 ///
-/// - [RFC 8693: OAuth 2.0 Token Exchange - act Claim](https://tools.ietf.org/html/rfc8693#section-4.1)
+/// - [RFC 8693: OAuth 2.0 Token Exchange - act Claim](https://tools.ietf.org/html/rfc8693#section-4.4)
 /// - [Custom Token Exchange Documentation](https://auth0.com/docs/authenticate/custom-token-exchange)
-public final class ActClaim: Sendable {
+public final class ActClaim: @unchecked Sendable {
 
     /// The subject identifier of the acting party.
-    public let sub: String?
+    ///
+    /// Per [RFC 8693 Section 4.4](https://tools.ietf.org/html/rfc8693#section-4.4), `sub` is required within an `act`
+    /// claim. An `act` claim without a `sub` is considered invalid and will not be parsed.
+    public let sub: String
 
     /// A nested `act` claim representing the next actor in a delegation chain.
     public let act: ActClaim?
 
     /// Any additional claims beyond `sub` and `act` (e.g., `org`, `role`).
-    public let additionalClaims: [String: String]
+    ///
+    /// Values are preserved as-is, including non-string JSON values (numbers, booleans, objects, arrays), so that
+    /// custom claims set via `api.authentication.setActor()` are not lost.
+    public let additionalClaims: [String: Any]
 
     /// Creates a new `ActClaim` from a JSON dictionary.
     ///
     /// - Parameter json: A dictionary representing the `act` claim from a decoded JWT.
-    /// - Returns: An `ActClaim` instance, or `nil` if the dictionary is empty.
+    /// - Returns: An `ActClaim` instance, or `nil` if the dictionary does not contain a `sub` claim.
     public init?(json: [String: Any]) {
-        guard !json.isEmpty else { return nil }
-
-        self.sub = json["sub"] as? String
+        guard let sub = json["sub"] as? String else { return nil }
+        self.sub = sub
 
         if let nestedAct = json["act"] as? [String: Any] {
             self.act = ActClaim(json: nestedAct)
@@ -49,11 +54,9 @@ public final class ActClaim: Sendable {
             self.act = nil
         }
 
-        var additional: [String: String] = [:]
+        var additional: [String: Any] = [:]
         for (key, value) in json where key != "sub" && key != "act" {
-            if let stringValue = value as? String {
-                additional[key] = stringValue
-            }
+            additional[key] = value
         }
         self.additionalClaims = additional
     }
--- a/Auth0/Authentication.swift
+++ b/Auth0/Authentication.swift
@@ -1375,7 +1375,7 @@ public extension Authentication {
                              scope: String = defaultScope,
                              organization: String? = nil,
                              actorToken: ActorToken?,
-                             parameters: [String: Any] = [:]) -> Request<Credentials, AuthenticationError> {
+                             parameters: [String: any Sendable] = [:]) -> Request<Credentials, AuthenticationError> {
         var parameters = parameters
         if let actorToken {
             parameters["actor_token"] = actorToken.token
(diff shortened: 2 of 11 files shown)


Auth0/ActClaim.swift
```diff
@@ -0,0 +1,60 @@
...
+///         print("Original actor: \(innerAct.sub)")
+///     }
+/// }
+/// ```
+///
+/// ## See Also
+///
+/// - [RFC 8693: OAuth 2.0 Token Exchange - act Claim](https://tools.ietf.org/html/rfc8693#section-4.1)
+/// - [Custom Token Exchange Documentation](https://auth0.com/docs/authenticate/custom-token-exchange)
+public final class ActClaim: Sendable {
```
sanchitmehtagit, line 26: could this be struct instead

Auth0/ActClaim.swift
```diff
@@ -0,0 +1,60 @@
...
+///     if let innerAct = act.act {
+///         print("Original actor: \(innerAct.sub)")
+///     }
+/// }
+/// ```
+///
+/// ## See Also
+///
+/// - [RFC 8693: OAuth 2.0 Token Exchange - act Claim](https://tools.ietf.org/html/rfc8693#section-4.1)
+/// - [Custom Token Exchange Documentation](https://auth0.com/docs/authenticate/custom-token-exchange)
```
kishore7snehil, line 25: This references RFC 8693 Section 4.1, but the `act` claim is defined in **Section 4.4** ("Delegation Semantics"). Section 4.1 covers the token exchange response structure.

Suggestion: `https://tools.ietf.org/html/rfc8693#section-4.4`

Auth0/ActClaim.swift
```diff
@@ -0,0 +1,60 @@
...
+public final class ActClaim: Sendable {
+
+    /// The subject identifier of the acting party.
+    public let sub: String?
+
```
kishore7snehil, line 33: Per RFC 8693 Section 4.4: "The `sub` claim is required" within an `act` claim. The Android SDK (`ActorClaim`) makes `sub` non-optional and returns `null` for the entire object if `sub` is missing from the JSON.

Here, `sub` is `String?` which means callers always need to nil-check even though a valid `act` claim per the RFC will always have it. This creates inconsistent behavior across platforms.

Consider either:
1. Making `sub` non-optional (`String`) and returning `nil` from `init?` if `sub` is absent, or
2. Documenting why the lenient approach was chosen (e.g., forward-compatibility with potential spec changes)

Auth0/ActClaim.swift
```diff
@@ -0,0 +1,60 @@
...
+        if let nestedAct = json["act"] as? [String: Any] {
+            self.act = ActClaim(json: nestedAct)
+        } else {
+            self.act = nil
+        }
+
+        var additional: [String: String] = [:]
+        for (key, value) in json where key != "sub" && key != "act" {
+            if let stringValue = value as? String {
+                additional[key] = stringValue
```
kishore7snehil: `additionalClaims` is `[String: String]` and the initializer only keeps values that cast to `String`, silently dropping non-string JSON values (booleans, numbers, objects, arrays). While RFC 8693 `act` claims primarily use string values like `sub`, custom Auth0 Actions could set non-string values via `api.authentication.setActor()` that would be lost here.

Auth0/Authentication.swift
```diff
@@ -1174,6 +1175,7 @@ public protocol Authentication: SenderConstraining, Trackable, Loggable, Sendabl
                              audience: String?,
                              scope: String,
                              organization: String?,
+                             actorToken: ActorToken?,
                              parameters: [String: Any]) -> Request<Credentials, AuthenticationError>
```
kishore7snehil, line 1177: Adding `actorToken: ActorToken?` as a required parameter to the public `Authentication` protocol is a source-breaking change for anyone who has created a custom conformance . Their implementation will fail to compile until they add the new parameter.

Consider noting it in the changelog, or providing a default implementation in a protocol extension to avoid breaking custom conformers.

Auth0Tests/AuthenticationSpec.swift
```diff
@@ -2152,6 +2152,33 @@ class AuthenticationSpec: QuickSpec {
...
+                    auth.customTokenExchange(subjectToken: subjectToken,
+                                            subjectTokenType: subjectTokenType,
+                                            actorToken: actor)
+                        .start { result in
+                            expect(result).to(haveCredentials(AccessToken, IdToken))
+                            done()
+                        }
+                }
+            }
         }
```
kishore7snehil, line 2205: This test verifies `actor_token`/`actor_token_type` ARE sent when provided, but there's no corresponding test verifying they are NOT present in the request body when `actorToken` is nil (the default).

The `hasAllOf` matcher only checks for presence of specified keys, not absence of others. Consider adding a test like:
```swift
it("should not include actor token params when actorToken is nil") {
    // ... call customTokenExchange without actorToken
    // assert body does NOT contain "actor_token" or "actor_token_type"
}
```

The Android PR includes this test (`shouldSendCTERequestWithoutActorToken`).

kishore7snehil (changes requested):

Auth0/UserInfo.swift
```diff
@@ -133,6 +134,17 @@ public struct UserInfo: JSONObjectPayload, @unchecked Sendable {
...
+    /// The acting party in a token exchange delegation chain.
+    ///
+    /// When present, represents the entity performing actions on behalf of the subject. The `act` claim may be nested
+    /// to represent delegation chains (e.g., `act.act` for multi-hop delegation).
+    ///
+    /// ## See Also
+    ///
+    /// - ``ActClaim``
+    /// - [RFC 8693: OAuth 2.0 Token Exchange](https://tools.ietf.org/html/rfc8693#section-4.1)
+    public let act: ActClaim?
```
kishore7snehil, line 148: Same issue as in `ActClaim.swift`: this links to Section 4.1 but the `act` claim is defined in **Section 4.4** ("Delegation Semantics"). Section 4.1 covers the token exchange response structure.

Should be: `https://tools.ietf.org/html/rfc8693#section-4.4`

kishore7snehil (changes requested):

EXAMPLES.md
```diff
@@ -4689,6 +4763,45 @@ Auth0
 ```
 </details>
 
+> [!NOTE]
+> When an actor token is provided, Auth0 will not issue a refresh token regardless of whether `offline_access` is in the scope.
+
+#### Reading the `act` claim from the ID token
+
+When a token exchange involves an actor token, Auth0 may include an `act` (actor) claim in the resulting ID token. This claim identifies who is acting on behalf of the subject and may be nested to represent delegation chains.
+
```
kishore7snehil, line 4772: This says "Auth0 *may* include an `act` claim" but doesn't explain *when* it will. The `act` claim only appears when an Action is configured server-side that calls `api.authentication.setActor()`. Without that, the response will never have an `act` claim regardless of whether `actor_token` is sent.

Consider adding a note like:
> The `act` claim is set server-side via an Auth0 Action that calls `api.authentication.setActor()`. Without this Action configured, the ID token will not contain an `act` claim even when an actor token is provided in the request.

EXAMPLES.md
```diff
@@ -4689,6 +4763,45 @@ Auth0
...
+The `act` claim is exposed through the `UserInfo` type via the `ActClaim` class:
+
+```swift
+import JWTDecode
+
```
kishore7snehil, line 4780: `JWTDecode` is a separate package ([JWTDecode.swift](https://github.com/auth0/JWTDecode.swift)) that needs to be added as a dependency. Developers copying this example may not know they need to add it to their project.

Consider adding a brief note: "This example uses [JWTDecode.swift](https://github.com/auth0/JWTDecode.swift) — add it to your project if you haven't already."

kishore7snehil (changes requested):

EXAMPLES.md
NandanPrabhu: Updated the docs EXAMPLES.md to include the exact server side configuration required

Auth0/ActClaim.swift
NandanPrabhu: Updated RFC references for act claim to 4.4

Auth0/ActClaim.swift
NandanPrabhu: Done. Updated the changes. Act claim init succeeds only is sub is non nil. Created a failable init to handle this scenario

Auth0/UserInfo.swift
NandanPrabhu: Updated the RFC reference to 4.4

Auth0Tests/AuthenticationSpec.swift
NandanPrabhu: Added the required test

Auth0/ActClaim.swift
NandanPrabhu: If its converted to struct it gives an error: value type 'ActClaim' cannot have a  stored 
    property that recursively contains it. We need recursive property as per RFC

Auth0/Authentication.swift
NandanPrabhu: Added overriding of existing CTE function to just include ActorToken { acttoken and actTokenType }

Auth0/ActClaim.swift
NandanPrabhu: Updated the logic where in type of value in key value pair can be string or another types

kishore7snehil (dismissed):
LGTM!

Auth0/ActClaim.swift
```diff
@@ -0,0 +1,63 @@
...
+///         print("Original actor: \(innerAct.sub)")
+///     }
+/// }
+/// ```
+///
+/// ## See Also
+///
+/// - [RFC 8693: OAuth 2.0 Token Exchange - act Claim](https://tools.ietf.org/html/rfc8693#section-4.4)
+/// - [Custom Token Exchange Documentation](https://auth0.com/docs/authenticate/custom-token-exchange)
+public final class ActClaim: @unchecked Sendable {
```
sanchitmehtagit: Let's avoid using `@unchecked Sendable`.

If the issue is caused by `additionalClaims`, which is currently typed as `[String: Any]`, consider changing it to `[String: any Sendable]` instead.

kishore7snehil (dismissed):
LGTM!

Auth0/Authentication.swift
```diff
@@ -1324,4 +1324,67 @@ public extension Authentication {
...
+    func customTokenExchange(subjectToken: String,
+                             subjectTokenType: String,
+                             audience: String? = nil,
+                             scope: String = defaultScope,
+                             organization: String? = nil,
+                             actorToken: ActorToken?,
+                             parameters: [String: Any] = [:]) -> Request<Credentials, AuthenticationError> {
+        var parameters = parameters
+        parameters["actor_token"] = actorToken?.token
+        parameters["actor_token_type"] = actorToken?.tokenType
```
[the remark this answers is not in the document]
NandanPrabhu: done

EXAMPLES.md
```diff
@@ -4672,11 +4676,81 @@ do {
...
+        if case .failure(let error) = completion {
+            print("Failed with: \(error)")
+        }
+    }, receiveValue: { credentials in
+        print("Obtained credentials: \(credentials)")
+    })
+    .store(in: &cancellables)
+```
+</details>
+
```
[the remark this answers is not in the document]
NandanPrabhu: done

Auth0/ActClaim.swift
NandanPrabhu: done

sanchitmehtagit (dismissed):

Auth0/Authentication.swift
```diff
@@ -1324,4 +1324,69 @@ public extension Authentication {
...
+     - [Custom Token Exchange Documentation](https://auth0.com/docs/authenticate/custom-token-exchange)
+     - [RFC 8693: OAuth 2.0 Token Exchange](https://tools.ietf.org/html/rfc8693)
+     */
+    func customTokenExchange(subjectToken: String,
+                             subjectTokenType: String,
+                             audience: String? = nil,
+                             scope: String = defaultScope,
+                             organization: String? = nil,
+                             actorToken: ActorToken?,
+                             parameters: [String: any Sendable] = [:]) -> Request<Credentials, AuthenticationError> {
```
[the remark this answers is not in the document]
NandanPrabhu: The code compiled and passed all tests with [String: any Sendable]. There was no
  build error, no CI failure, no runtime bug. any Sendable forwards cleanly to the protocol's [String: Any] parameter, so it works correctly as-is.

sanchitmehtagit (approved):

PR #1208, opened 2026-06-12
feat: IPSIE session_expiry enforcement in CredentialsManager [EA]

NandanPrabhu:
### Added

- **`CredentialsManagerError.sessionExpired`** — new error case returned by `credentials()`, `ssoCredentials()`, and `apiCredentials()` when the IPSIE `session_expiry` ceiling is reached. Stored credentials are cleared automatically. Apps can match it explicitly or let an existing `noCredentials` re-login path handle it transparently.

- **`Credentials.sessionExpiresAt: Int?`** — computed property that decodes the `session_expiry` Unix timestamp from the current ID token on demand. Returns `nil` when the claim is absent. Intended for app-level display (e.g. countdown timers); `CredentialsManager` uses the Keychain-pinned value for authoritative enforcement.

- **`session_expiry` enforcement in `CredentialsManager`** — on every `credentials()`, `ssoCredentials()`, and `apiCredentials()` call, the upstream-IdP session ceiling is checked **before** the access-token validity check and before any refresh-token exchange. Once the ceiling has passed (with a **30 s clock-skew leeway**), stored credentials are cleared and `.sessionExpired` is returned immediately — no call to `/oauth/token` is made.

- **Keychain-pinned ceiling (`sessionExpiryKey`)** — on the first `store(credentials:)`, the `session_expiry` value from the ID token is written once to the Keychain under a dedicated key (default `"session_expiry"`, configurable via the new `sessionExpiryKey` init parameter). The `pinSessionExpiry(for:)` guard prevents any subsequent `store()` call from overwriting it, so the original ceiling is preserved even when a renewed ID token re-emits or omits the claim. `clear()` removes the pinned entry on logout.

- **`store(credentials:)` DocC note** — documents that the DPoP thumbprint and `session_expiry` ceiling are both session-scoped and that `clear()` must be called before storing credentials for a different user to avoid stale pinned values.

- **`parseSessionExpiry(fromIdToken:)`** — module-internal free function in `Credentials.swift` shared by both `Credentials.sessionExpiresAt` and `CredentialsManager`. Decodes the claim via `NSNumber` (truncates fractional values rather than dropping them) and rejects values outside `(0, 10_000_000_000)` to guard against millisecond timestamps that would silently disable the ceiling.

- **Sample app** — redacted `accessToken` from all `print` statements in `ContentViewModel`.

- **README and EXAMPLES** — added an [Early Access] section covering enforcement behaviour, Keychain-pinning contract, error handling with explicit pattern-match example, upgrade guidance for existing apps, and how to read `sessionExpiresAt`.

### Behaviour table

| Scenario | Behaviour |
|---|---|
| `session_expiry` absent from ID token | Non-breaking — falls through to existing behaviour |
| `session_expiry` in the future | Credentials returned normally |
| `session_expiry` within 30 s leeway | `.sessionExpired`; credentials cleared |
| `session_expiry` in the past, valid access token | `.sessionExpired`; check fires before the early-return success path |
| `session_expiry` in the past, expired access token | `.sessionExpired`; no `/oauth/token` call made |
| Value ≤ 0 or ≥ 10,000,000,000 (e.g. millisecond timestamp) | Treated as absent — fail-open |
| Fractional value (e.g. `1234567890.75`) | Truncated via `NSNumber.intValue`, not dropped |
| Renewal: new ID token re-emits a later ceiling | Keychain-pinned original value preserved; original ceiling still enforced |
| Renewal: new ID token omits the claim | Keychain-pinned value survives; ceiling still enforced |
| Session stored before this SDK version (no Keychain entry) | Falls back to live ID token claim — backward compatible |
| `ssoCredentials()` / `apiCredentials()` when ceiling reached | `.sessionExpired`; exchange not attempted |

## References

- [IPSIE SL1 profile §3.2.2](https://openid.github.io/ipsie-openid-sl1/draft-openid-ipsie-sl1-profile.html)

## Testing

- [x] `xcodebuild test -project Auth0.xcodeproj -scheme Auth0.iOS -destination 'platform=iOS Simulator,name=iPhone 17'` — all tests pass
- [x] 18 cases under `"session_expiry enforcement"` in `CredentialsManagerSpec` pass
- [x] 4 cases in `CredentialsManagerErrorSpec` pass
- [x] Log in with a Post-Login Action injecting `session_expiry = now + 120`; after ~90 s `credentials()` returns `.sessionExpired`
- [x] Sessions on connections without the option are unaffected

### DIFF
--- a/App/ContentViewModel.swift
+++ b/App/ContentViewModel.swift
@@ -25,7 +25,7 @@ final class ContentViewModel: ObservableObject {
             let stored = credentialsManager.store(credentials: credentials)
             if stored {
                 isAuthenticated = true
-                print("Access Token: \(credentials.accessToken)")
+                print("Access Token: <redacted>")
             } else {
                 errorMessage = "Failed to store credentials"
             }
@@ -56,7 +56,7 @@ final class ContentViewModel: ObservableObject {
             let stored = credentialsManager.store(credentials: credentials)
             if stored {
                 isAuthenticated = true
-                print("Access Token: \(credentials.idToken)")
+                print("Access Token: <redacted>")
             } else {
                 errorMessage = "Failed to store credentials"
             }
@@ -103,7 +103,7 @@ final class ContentViewModel: ObservableObject {
         do {
             let credentials = try await credentialsManager.credentials()
             isAuthenticated = true
-            print("Valid credentials found: \(credentials.idToken)")
+            print("Valid credentials found: <redacted>")
         } catch {
             isAuthenticated = false
             print("No valid credentials found")
--- a/App/ContentViewModel.swift
+++ b/App/ContentViewModel.swift
@@ -56,7 +56,7 @@ final class ContentViewModel: ObservableObject {
             let stored = credentialsManager.store(credentials: credentials)
             if stored {
                 isAuthenticated = true
-                print("Access Token: \(credentials.accessToken)")
+                print("Access Token: \(credentials.idToken)")
             } else {
                 errorMessage = "Failed to store credentials"
             }
@@ -103,7 +103,7 @@ final class ContentViewModel: ObservableObject {
         do {
             let credentials = try await credentialsManager.credentials()
             isAuthenticated = true
-            print("Valid credentials found: \(credentials.accessToken)")
+            print("Valid credentials found: \(credentials.idToken)")
         } catch {
             isAuthenticated = false
             print("No valid credentials found")
--- a/Auth0.xcodeproj/project.pbxproj
+++ b/Auth0.xcodeproj/project.pbxproj
@@ -4101,7 +4101,7 @@
 					"@executable_path/Frameworks",
 				);
 				MARKETING_VERSION = 1.0;
-				PRODUCT_BUNDLE_IDENTIFIER = com.auth0.OAuth2.dev;
+				PRODUCT_BUNDLE_IDENTIFIER = com.auth0.OAuth2;
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				PROVISIONING_PROFILE = "";
 				PROVISIONING_PROFILE_SPECIFIER = "";
@@ -4126,7 +4126,7 @@
 					"@executable_path/Frameworks",
 				);
 				MARKETING_VERSION = 1.0;
-				PRODUCT_BUNDLE_IDENTIFIER = com.auth0.OAuth2.dev;
+				PRODUCT_BUNDLE_IDENTIFIER = com.auth0.OAuth2;
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				PROVISIONING_PROFILE_SPECIFIER = "";
 				SWIFT_VERSION = 5.0;
--- a/Auth0.xcodeproj/project.pbxproj
+++ b/Auth0.xcodeproj/project.pbxproj
@@ -4101,7 +4101,7 @@
 					"@executable_path/Frameworks",
 				);
 				MARKETING_VERSION = 1.0;
-				PRODUCT_BUNDLE_IDENTIFIER = com.auth0.OAuth2;
+				PRODUCT_BUNDLE_IDENTIFIER = com.auth0.OAuth2.dev;
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				PROVISIONING_PROFILE = "";
 				PROVISIONING_PROFILE_SPECIFIER = "";
@@ -4126,7 +4126,7 @@
 					"@executable_path/Frameworks",
 				);
 				MARKETING_VERSION = 1.0;
-				PRODUCT_BUNDLE_IDENTIFIER = com.auth0.OAuth2;
+				PRODUCT_BUNDLE_IDENTIFIER = com.auth0.OAuth2.dev;
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				PROVISIONING_PROFILE_SPECIFIER = "";
 				SWIFT_VERSION = 5.0;
--- a/Auth0/Credentials.swift
+++ b/Auth0/Credentials.swift
@@ -11,6 +11,24 @@ private struct _A0Credentials {
     let recoveryCode: String?
 }
 
+/// Decodes the IPSIE `session_expiry` claim from a JWT string.
+///
+/// Returns the Unix-seconds timestamp as an `Int`, or `nil` when the token is absent, unparseable,
+/// does not carry the claim, or carries a value outside `(0, 10_000_000_000)`.
+/// The upper bound rejects timestamps expressed in milliseconds (13-digit values produced by
+/// `Date.now()` in JavaScript) which would silently disable the ceiling if accepted.
+/// The claim is read as `NSNumber` so a fractional value is truncated rather than dropped.
+func parseSessionExpiry(fromIdToken idToken: String?) -> Int? {
+    guard let idToken = idToken,
+          let jwt = try? decode(jwt: idToken),
+          let rawValue = jwt.body["session_expiry"] as? NSNumber else {
+        return nil
+    }
+    let value = rawValue.intValue
+    guard value > 0, value < 10_000_000_000 else { return nil }
+    return value
+}
+
 /// User's credentials obtained from Auth0.
 @objc(A0Credentials)
 public final class Credentials: NSObject, Sendable {
@@ -80,19 +98,13 @@ public final class Credentials: NSObject, Sendable {
     /// The value is decoded on demand from ``idToken``. A value that is not a plausible Unix-seconds
     /// timestamp (outside `(0, 10_000_000_000)`) is treated as "no ceiling" and returns `nil`.
     ///
-    /// - Important: The ``CredentialsManager`` enforces this ceiling using the value pinned at the
-    /// initial login. After a refresh whose new ID token omits the claim, this property returns `nil`
-    /// even though the Credentials Manager still enforces the pinned ceiling.
+    /// - Important: The ``CredentialsManager`` enforces this ceiling using the value pinned to the
+    /// Keychain at the initial login. The pinned value is never updated by a refresh-token grant, so
+    /// even if a renewal returns an ID token that omits or lowers the claim, the original ceiling
+    /// remains in effect. This property reflects the *current* ID token only — use
+    /// ``CredentialsManager`` for authoritative enforcement.
     public var sessionExpiresAt: Int? {
-        guard let jwt = try? decode(jwt: self.idToken),
-              let rawValue = jwt.body["session_expiry"] as? NSNumber else {
-            return nil
-        }
-        let sessionExpiry = rawValue.intValue
-        guard sessionExpiry > 0, sessionExpiry < 10_000_000_000 else {
-            return nil
-        }
-        return sessionExpiry
+        return parseSessionExpiry(fromIdToken: self.idToken)
     }
 
     /// Custom description that redacts the tokens with `<REDACTED>`.
--- a/Auth0/Credentials.swift
+++ b/Auth0/Credentials.swift
@@ -17,14 +17,14 @@ private struct _A0Credentials {
 /// does not carry the claim, or carries a value outside `(0, 10_000_000_000)`.
 /// The upper bound rejects timestamps expressed in milliseconds (13-digit values produced by
 /// `Date.now()` in JavaScript) which would silently disable the ceiling if accepted.
-/// The claim is read as `NSNumber` so a fractional value is truncated rather than dropped.
+/// The claim is read as `Double` so a fractional value is truncated rather than dropped.
 func parseSessionExpiry(fromIdToken idToken: String?) -> Int? {
     guard let idToken = idToken,
           let jwt = try? decode(jwt: idToken),
-          let rawValue = jwt.body["session_expiry"] as? NSNumber else {
+          let rawValue = jwt.body["session_expiry"] as? Double else {
         return nil
     }
-    let value = rawValue.intValue
+    let value = Int(rawValue)
     guard value > 0, value < 10_000_000_000 else { return nil }
     return value
 }
@@ -88,21 +88,11 @@ public final class Credentials: NSObject, Sendable {
     /// - [MFA Recovery Codes](https://auth0.com/docs/secure/multi-factor-authentication/configure-recovery-codes-for-mfa)
     public let recoveryCode: String?
 
-    /// The absolute session-expiry ceiling, in **Unix seconds**, asserted by the upstream identity
-    /// provider via the IPSIE `session_expiry` claim, or `nil` when the connection does not emit it.
+    /// Unix-seconds timestamp of the IPSIE `session_expiry` ceiling, or `nil` when the claim is absent
+    /// or outside the valid range `(0, 10_000_000_000)`.
     ///
-    /// This is a session-level ceiling that is independent of ``expiresIn`` (the access-token expiry):
-    /// it usually sits much further out and caps how long the local session may live, regardless of
-    /// access-token renewals. A `nil` value means there is no such ceiling.
-    ///
-    /// The value is decoded on demand from ``idToken``. A value that is not a plausible Unix-seconds
-    /// timestamp (outside `(0, 10_000_000_000)`) is treated as "no ceiling" and returns `nil`.
-    ///
-    /// - Important: The ``CredentialsManager`` enforces this ceiling using the value pinned to the
-    /// Keychain at the initial login. The pinned value is never updated by a refresh-token grant, so
-    /// even if a renewal returns an ID token that omits or lowers the claim, the original ceiling
-    /// remains in effect. This property reflects the *current* ID token only — use
-    /// ``CredentialsManager`` for authoritative enforcement.
+    /// - Important: Reflects the *current* ID token only. ``CredentialsManager`` enforces the ceiling
+    /// using the value pinned to the Keychain at initial login, which survives refresh-token renewals.
     public var sessionExpiresAt: Int? {
         return parseSessionExpiry(fromIdToken: self.idToken)
     }
--- a/Auth0/Credentials.swift
+++ b/Auth0/Credentials.swift
@@ -1,4 +1,5 @@
 import Foundation
+import JWTDecode
 
 private struct _A0Credentials {
     let accessToken: String
@@ -68,6 +69,31 @@ public final class Credentials: NSObject, Sendable {
     ///
     /// - [MFA Recovery Codes](https://auth0.com/docs/secure/multi-factor-authentication/configure-recovery-codes-for-mfa)
     public let recoveryCode: String?
+
+    /// The absolute session-expiry ceiling, in **Unix seconds**, asserted by the upstream identity
+    /// provider via the IPSIE `session_expiry` claim, or `nil` when the connection does not emit it.
+    ///
+    /// This is a session-level ceiling that is independent of ``expiresIn`` (the access-token expiry):
+    /// it usually sits much further out and caps how long the local session may live, regardless of
+    /// access-token renewals. A `nil` value means there is no such ceiling.
+    ///
+    /// The value is decoded on demand from ``idToken``. A value that is not a plausible Unix-seconds
+    /// timestamp (outside `(0, 10_000_000_000)`) is treated as "no ceiling" and returns `nil`.
+    ///
+    /// - Important: The ``CredentialsManager`` enforces this ceiling using the value pinned at the
+    /// initial login. After a refresh whose new ID token omits the claim, this property returns `nil`
+    /// even though the Credentials Manager still enforces the pinned ceiling.
+    public var sessionExpiresAt: Int? {
+        guard let jwt = try? decode(jwt: self.idToken),
+              let rawValue = jwt.body["session_expiry"] as? NSNumber else {
+            return nil
+        }
+        let sessionExpiry = rawValue.intValue
+        guard sessionExpiry > 0, sessionExpiry < 10_000_000_000 else {
+            return nil
+        }
+        return sessionExpiry
+    }
 
     /// Custom description that redacts the tokens with `<REDACTED>`.
     public override var description: String {
--- a/Auth0/Credentials.swift
+++ b/Auth0/Credentials.swift
@@ -11,24 +11,6 @@ private struct _A0Credentials {
     let recoveryCode: String?
 }
 
-/// Decodes the IPSIE `session_expiry` claim from a JWT string.
-///
-/// Returns the Unix-seconds timestamp as an `Int`, or `nil` when the token is absent, unparseable,
-/// does not carry the claim, or carries a value outside `(0, 10_000_000_000)`.
-/// The upper bound rejects timestamps expressed in milliseconds (13-digit values produced by
-/// `Date.now()` in JavaScript) which would silently disable the ceiling if accepted.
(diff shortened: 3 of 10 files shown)


Auth0/CredentialsManager.swift
```diff
@@ -834,7 +834,10 @@ public struct CredentialsManager: Sendable {
                 complete()
                 return callback(.failure(.noRefreshToken))
             }
-            
+            guard !self.hasSessionExpired(idToken: credentials.idToken) else {
```
kishore7snehil: This ceiling check sits *after* the early-success return a few lines up, where `credentials()` returns the cached credentials when the access token is still valid (not expired, scope unchanged, minTTL met). 

So a session that's past its `session_expiry` but still holding a valid access token keeps returning credentials, the ceiling only fires when a renewal would otherwise happen, not on every read. It's also after the `noRefreshToken` guard, so an expired-ceiling session with no refresh token gets `.noRefreshToken` instead of `.sessionExpired`.

We should move this guard to immediately after the credentials are loaded, before both the validity early-return and the refresh-token guard, so the ceiling is enforced on every read regardless of token state.

Auth0/CredentialsManager.swift
```diff
@@ -1001,6 +1004,18 @@ public struct CredentialsManager: Sendable {
         return expiresIn < Date()
     }
 
+    /// Checks whether the IPSIE `session_expiry` ceiling has been reached.
+    /// Reads `session_expiry` from the ID token claims only.
+    /// Applies a 30-second leeway to account for clock skew between the device and Auth0.
+    func hasSessionExpired(idToken: String) -> Bool {
```
kishore7snehil: The PR description says `session_expiry` is persisted to a dedicated keychain key and used as a fallback so the ceiling survives refresh-token renewals (whose new ID token doesn't carry the claim). There's no such persistence in the diff - this method reads the claim only from the current ID token. 
On renewal the new credentials adopt the renewal response's ID token (which lacks `session_expiry`), so the next check finds no claim and returns false: enforcement silently stops after the first renewal, the exact thing the fallback was meant to prevent. 

We should try implementing the persisted fallback (extract at store time, read it here when the token lacks the claim, clear it in `clear()`).

Auth0/CredentialsManager.swift
```diff
@@ -1001,6 +1004,18 @@ public struct CredentialsManager: Sendable {
...
+        guard let jwt = try? decode(jwt: idToken),
+              let sessionExpiry = jwt.body["session_expiry"] as? Int else {
```
kishore7snehil: The claim is read with `as? Int` and no range check. 

Since `session_expiry` is set by a customer-side action it's untrusted input, and the most likely mistake is emitting milliseconds (`* 1000`): a 13-digit ms value read as seconds lands tens of thousands of years out, never trips the check, and silently disables the ceiling while looking enabled . 

We should reject implausible values (`<= 0` and `>= 10_000_000_000`, the seconds-vs-ms boundary ≈ year 2286) and failing open to "no ceiling." While here, decoding permissively (accept a numeric value, not strictly `Int`) avoids a fractional NumericDate silently casting to nil.

EXAMPLES.md
```diff
@@ -658,6 +659,76 @@ let isValid = credentialsManager.isBiometricSessionValid()
...
+### IPSIE session expiry [EA]
+
+> [!NOTE]
+> This feature is currently available in [Early Access](https://auth0.com/docs/troubleshoot/product-lifecycle/product-release-stages#early-access). It requires the `id_token_session_expiry_federated` feature flag and the `id_token_session_expiry_supported` option enabled on your OIDC or Okta enterprise connection.
+
+Auth0 supports the [IPSIE SL1](https://openid.github.io/ipsie-openid-sl1/draft-openid-ipsie-sl1-profile.html) `session_expiry` claim, which lets an upstream identity provider (e.g. Okta) set a hard ceiling on how long an Auth0-issued session may live. When your connection has this option enabled, Auth0 includes a `session_expiry` Unix timestamp in the ID token it returns to your app after login.
+
+The `CredentialsManager` automatically enforces this ceiling. On every `credentials()` call it reads `session_expiry` from the stored ID token and, once the ceiling has passed (with a 30-second clock-skew leeway), returns a `CredentialsManagerError.sessionExpired` error instead of attempting a token renewal. No code changes are needed to opt in — the enforcement is transparent once the connection option is active on your tenant.
```
kishore7snehil: This says the ceiling is enforced automatically, but only `credentials()` runs the check.`ssoCredentials()` and `apiCredentials()` don't, so those surfaces return tokens past the ceiling. 

If that's intentional, scope the wording here to `credentials()`; if not, the other two retrieval paths need the same guard (and tests).

EXAMPLES.md
```diff
@@ -658,6 +659,76 @@ let isValid = credentialsManager.isBiometricSessionValid()
...
+
+#### Reading the raw `session_expiry` value
+
+You can inspect the claim directly from the stored ID token for app-level logic (e.g. a countdown timer):
+
+```swift
+import JWTDecode
+
+if let credentials = credentialsManager.user,
+   let jwt = try? decode(jwt: credentialsManager.credentials { _ in }./* ... */ ""),
```
kishore7snehil: This "reading the raw value" snippet doesn't compile — `decode(jwt: credentialsManager.credentials { _ in }./* ... */ "")` mixes an async-callback API into an inline expression and never yields a token string, and `credentialsManager.user` returns user info, not the ID token. 

We should add something that decodes the ID token from a retrieved `Credentials`: `if let jwt = try? decode(jwt: credentials.idToken), let sessionExpiry = jwt.body["session_expiry"] as? Int { ... }`.

App/ContentViewModel.swift
```diff
@@ -56,7 +56,7 @@ final class ContentViewModel: ObservableObject {
             let stored = credentialsManager.store(credentials: credentials)
             if stored {
                 isAuthenticated = true
-                print("Access Token: \(credentials.accessToken)")
+                print("Access Token: \(credentials.idToken)")
```
kishore7snehil: This (and the matching line further down) switches the debug print from the access token to the ID token, which is unrelated to the feature and logs user info to the console. Ideally it should just print an `expiresIn`

Auth0.xcodeproj/project.pbxproj
```diff
@@ -4077,7 +4077,7 @@
 					"@executable_path/Frameworks",
 				);
 				MARKETING_VERSION = 1.0;
-				PRODUCT_BUNDLE_IDENTIFIER = com.auth0.OAuth2;
+				PRODUCT_BUNDLE_IDENTIFIER = com.auth0.OAuth2.dev;
```
kishore7snehil: This bundle-identifier change (`com.auth0.OAuth2` → `com.auth0.OAuth2.dev`, in both Debug and Release) looks unrelated to the feature. Is this intentional?

Auth0Tests/CredentialsManagerSpec.swift
```diff
@@ -932,8 +954,116 @@ class CredentialsManagerSpec: QuickSpec {
...
+
+                waitUntil(timeout: Timeout) { done in
+                    credentialsManager.credentials { result in
+                        expect(result).to(beSuccessful())
+                        done()
+                    }
+                }
+            }
+
+            it("should succeed after renewal when new id token has no session_expiry claim") {
```
kishore7snehil, line 1086: 1. this case asserts `beSuccessful()` after a renewal whose new ID token has no claim - that misses the the silent-disable behaviour above as expected rather than catching it
2.  every `.sessionExpired` test forces renewal via an expired access token, so there's no test for the real hole, a *valid* access token with a past ceiling (which currently returns success)
3. no malformed-value tests (milliseconds / negative / zero / non-numeric) to pin the fail-open behaviour
4. nothing covers `ssoCredentials`/`apiCredentials`. Worth adding failure-asserting cases for each.

kishore7snehil (changes requested):

Auth0.xcodeproj/project.pbxproj
NandanPrabhu: removed the change

App/ContentViewModel.swift
NandanPrabhu: reverted this

NandanPrabhu:
@coderabbitai full review

EXAMPLES.md
NandanPrabhu: Updated it. Besides session expiry will be retrieved everytime by parsing id token as part of credentials object. So EXAMPLES.md will be have this accordingly.

EXAMPLES.md
NandanPrabhu: sso credentials and api credentials also has this check

Auth0/CredentialsManager.swift
NandanPrabhu: Add this check too

Auth0/CredentialsManager.swift
NandanPrabhu: done. Added it at the top before returning cached credentials

Auth0/CredentialsManager.swift
```diff
@@ -59,15 +60,18 @@ public struct CredentialsManager: Sendable {
     ///   - authentication: Auth0 Authentication API client.
     ///   - storeKey:       Key used to store user credentials in the Keychain. Defaults to 'credentials'.
     ///   - dpopThumprintKey: Key used to store dpop thumbprint. Defaults to 'dpop_thumbprint'.
+    ///   - sessionExpiryKey: Key used to store the IPSIE `session_expiry` ceiling. Defaults to 'session_expiry'.
     ///   - storage:        The ``CredentialsStorage`` instance used to manage credentials storage. Defaults to a standard `SimpleKeychain` instance.
     ///   - maxRetries:     Maximum number of retry attempts for credential renewal on transient errors. Defaults to 0.
     public init(authentication: Authentication,
                 storeKey: String = "credentials",
                 dpopThumbprintKey: String = "dpop_thumbprint",
+                sessionExpiryKey: String = "session_expiry",
```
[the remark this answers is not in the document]
NandanPrabhu: Persistent storage is removed to match other claims too

Auth0/CredentialsManager.swift
NandanPrabhu: removed persisting of session_expiry. Session expiry is always retrieving from parsing id token

NandanPrabhu:
@coderabbitai full review

Auth0Tests/CredentialsManagerSpec.swift
NandanPrabhu: done. added required test cases

Auth0Tests/CredentialsSpec.swift
```diff
@@ -222,5 +222,48 @@ class CredentialsSpec: QuickSpec {
...
+    }
+    let headerJSON = try! JSONSerialization.data(withJSONObject: ["typ": "JWT", "alg": "HS256"])
+    let payloadJSON = try! JSONSerialization.data(withJSONObject: payload)
+    func encode(_ data: Data) -> String {
+        return data.base64EncodedString()
+            .replacingOccurrences(of: "+", with: "-")
+            .replacingOccurrences(of: "/", with: "_")
+            .replacingOccurrences(of: "=", with: "")
     }
+    return "\(encode(headerJSON)).\(encode(payloadJSON)).fakesig"
```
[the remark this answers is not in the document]
NandanPrabhu: resolved

Auth0/CredentialsManager.swift
```diff
@@ -59,15 +60,18 @@ public struct CredentialsManager: Sendable {
...
                 storage: CredentialsStorage = SimpleKeychain(),
                 maxRetries: Int = 0) {
         self.storeKey = storeKey
         self.dpopThumbprintKey = dpopThumbprintKey
+        self.sessionExpiryKey = sessionExpiryKey
```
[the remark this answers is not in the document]
NandanPrabhu: we have similar keys approach for other data too. Hence not resolving for now

Auth0/CredentialsManager.swift
```diff
@@ -1001,6 +1030,46 @@ public struct CredentialsManager: Sendable {
...
+    func sessionExpiry(fromIdToken idToken: String?) -> Int? {
+        return parseSessionExpiry(fromIdToken: idToken)
+    }
+
+    /// Writes the `session_expiry` ceiling to the Keychain on the first login.
+    ///
+    /// Only writes when no ceiling is already pinned, so calling `store(credentials:)` on a renewal
+    /// grant is safe — the existing pinned value is preserved.
+    private func pinSessionExpiry(for credentials: Credentials) {
+        guard pinnedSessionExpiry() == nil,
```
kishore7snehil, line 1058: The pin is keyed to the Keychain key rather than to the user, and it only resets inside `clear()`. So if an app logs in a different user with `store(credentials:)` without calling `clear()` first, which is an account switch that skips logout, the previous user's ceiling stays in force for the new session. If the old ceiling sits further out than the new user's, the new session ends up running past its own ceiling, and if it sits nearer, the new user gets bounced early.

The documented logout path calls `clear()`, so the normal flow is fine and this only affects a direct re-store for a different subject. Shall we reset the pin when `store()` sees a different `sub`, or is relying on `clear()` the intended contract here? WDYT?

EXAMPLES.md
```diff
@@ -658,6 +659,79 @@ let isValid = credentialsManager.isBiometricSessionValid()
...
+
+> [!NOTE]
+> This feature is currently available in [Early Access](https://auth0.com/docs/troubleshoot/product-lifecycle/product-release-stages#early-access). It requires the `id_token_session_expiry_federated` feature flag and the `id_token_session_expiry_supported` option enabled on your OIDC or Okta enterprise connection.
+
+Auth0 supports the [IPSIE SL1](https://openid.github.io/ipsie-openid-sl1/draft-openid-ipsie-sl1-profile.html) `session_expiry` claim, which lets an upstream identity provider (e.g. Okta) set a hard ceiling on how long an Auth0-issued session may live. When your connection has this option enabled, Auth0 includes a `session_expiry` Unix timestamp in the ID token it returns to your app after login.
+
+The `CredentialsManager` enforces this ceiling on every retrieval — `credentials()`, `ssoCredentials()`, and `apiCredentials()`. Once the ceiling has passed (with a 30-second clock-skew leeway), the call clears the stored credentials and returns a `CredentialsManagerError.sessionExpired` error instead of attempting a token renewal. No code changes are needed to opt in — the enforcement is transparent once the connection option is active on your tenant.
```
kishore7snehil, line 669: This covers the error case well, but there's nothing here for apps that are upgrading. With the feature active, `credentials()` can start returning `.sessionExpired` for a user who was previously logged in once the ceiling passes, so code that assumed a stored session stays usable until the access token expires now has a new failure path to handle.

Could we add a short upgrading line here noting that existing sessions and connections without the option carry no ceiling and behave exactly as before, but once the option is on the re-login path needs to handle `.sessionExpired`?

EXAMPLES.md
```diff
@@ -658,6 +659,79 @@ let isValid = credentialsManager.isBiometricSessionValid()
...
+
+The ceiling is **pinned at the initial login**: the `session_expiry` value from the first ID token is persisted to the Keychain and never overwritten by a refresh-token grant. This means a renewal whose ID token re-emits the claim cannot raise the ceiling past the original value. `clear()` removes the pinned value on logout so the next login pins a fresh ceiling.
```
kishore7snehil, line 671: The Early Access callout above names the feature flags - that's internal to us. So we should avoid sharing this information

kishore7snehil (changes requested):

EXAMPLES.md
NandanPrabhu: Removed reference of FF. Doc mentions Use ID Token for Session Expiry from tenant dashboard settings though

EXAMPLES.md
NandanPrabhu: Added a note for the same

NandanPrabhu: https://github.com/auth0/Auth0.swift/pull/1208/changes#diff-30bf75f8f19e56642c61ecaa902d4d6a0a0126a234b800d76b79a7f1faab8be2R715

Auth0/CredentialsManager.swift
NandanPrabhu: Added a note for the same in CredentialsManager. swift for the developer to call clear first before storing credentials

README.md
```diff
@@ -357,6 +357,28 @@ do {
 ```
 </details>
 
+### IPSIE session expiry [EA]
+
+> [!NOTE]
+> This feature is currently available in [Early Access](https://auth0.com/docs/troubleshoot/product-lifecycle/product-release-stages#early-access). It requires the **Use ID Token for Session Expiry** option enabled on your OIDC or Okta enterprise connection.
+
```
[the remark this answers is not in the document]
NandanPrabhu: done. only generic requirements for the OIDC or Okta enterprise connection is described

Auth0/CredentialsManager.swift
utkrishtsahu, line 424: The session_expiry check happens inside retrieveCredentials → retrieveCredentialsWithRetry, which only runs after bioAuth.validateBiometric (line 433). That means a user whose session ceiling has already passed still gets prompted for Face ID / Touch ID, and only after they authenticate do we return .sessionExpired. We should short-circuit before the biometric prompt, like Android does in SecureCredentialsManager.getCredentials (if (isSessionExpired(null)) { clearCredentials(); … return } ahead of the prompt). Since the ceiling is pinned in the Keychain, we can check it here without loading the credentials:


if self.hasSessionExpired(idToken: nil) {
    _ = self.clear()
    return callback(.failure(.sessionExpired))
}
Suggest adding this at the very top of credentials(...), before the if let bioAuth block, so it also covers the non-biometric branch.

Auth0/CredentialsManager.swift
```diff
@@ -939,6 +965,15 @@ public struct CredentialsManager: Sendable {
                                         headers: [String: String],
                                         callback: @escaping (CredentialsManagerResult<APICredentials>) -> Void) {
         SynchronizationBarrier.shared.execute { complete in
+            // IPSIE session_expiry: enforce the upstream-IdP session ceiling before serving cached
+            // API credentials or exchanging the refresh token, so the session is never extended past
+            // it. The ceiling is read from the value pinned at login (with the stored ID token as a
+            // fallback); past it we clear and surface the dedicated error.
+            if self.hasSessionExpired(idToken: self.retrieveCredentials()?.idToken) {
```
utkrishtsahu, line 982: When the ceiling is hit here we call clear(), which removes the main credentials, DPoP thumbprint, and the pinned session_expiry key — but not the API credentials stored under the audience key. On a subsequent apiCredentials() call, retrieveCredentials() is now nil and the pinned ceiling is gone, so hasSessionExpired returns false, and the still-cached API credentials get served if their own TTL hasn't lapsed. Edge case (needs the API token to still be within its own expiry), but to be strict we should also clear the audience-scoped API entry when surfacing .sessionExpired. Android has the same shape, so a follow-up is fine if you'd rather not expand scope here.

utkrishtsahu (changes requested):

Auth0/CredentialsManager.swift
NandanPrabhu: resolved. session expiry is validated before invoking bio auth

Auth0/CredentialsManager.swift
NandanPrabhu: This case will be fixed with clearAll added in v3 but for now since clear cannot delete audience based creds, have added a note specifying this

utkrishtsahu (dismissed):

kishore7snehil (dismissed):
LGTM!

Auth0/Credentials.swift
```diff
@@ -10,6 +11,24 @@ private struct _A0Credentials {
...
 }
 
+/// Decodes the IPSIE `session_expiry` claim from a JWT string.
+///
+/// Returns the Unix-seconds timestamp as an `Int`, or `nil` when the token is absent, unparseable,
+/// does not carry the claim, or carries a value outside `(0, 10_000_000_000)`.
+/// The upper bound rejects timestamps expressed in milliseconds (13-digit values produced by
+/// `Date.now()` in JavaScript) which would silently disable the ceiling if accepted.
+/// The claim is read as `Double` so a fractional value is truncated rather than dropped.
+func parseSessionExpiry(fromIdToken idToken: String?) -> Int? {
```
sanchitmehtagit: Given Credentials.swift already imports JWTDecode, I would make it  a static method on Credentials as it a ntaural fit there
 // Credentials.swift
`static func parseSessionExpiry(fromIdToken idToken: String?) -> Int? { ... }`

This also lets you delete the thin sessionExpiry(fromIdToken:) wrapper on CredentialsManager - callers just reference 
Credentials.parseSessionExpiry(fromIdToken:) directly.

Auth0/CredentialsManager.swift
```diff
@@ -1001,6 +1046,46 @@ public struct CredentialsManager: Sendable {
         return expiresIn < Date()
     }
 
+    /// Negative clock-skew leeway (seconds) applied when checking the `session_expiry` ceiling, so the
+    /// session is treated as expired slightly *before* the wall-clock ceiling, never after.
+    private static let sessionExpiryLeeway: TimeInterval = 30
+
+    func sessionExpiry(fromIdToken idToken: String?) -> Int? {
+        return parseSessionExpiry(fromIdToken: idToken)
```
sanchitmehtagit: parseSessionExpiry I would make this a static function on Credentials and call that directly and remove this wrapper

EXAMPLES.md
```diff
@@ -658,6 +659,82 @@ let isValid = credentialsManager.isBiometricSessionValid()
...
+
+#### Reading the `session_expiry` value
+
+`Credentials` exposes the ceiling via the `sessionExpiresAt` property (Unix seconds, or `nil` when the connection does not emit the claim), which you can use for app-level logic such as a countdown timer:
+
+```swift
+credentialsManager.credentials { result in
+    guard case .success(let credentials) = result,
+          let sessionExpiry = credentials.sessionExpiresAt else { return }
+    let expiresAt = Date(timeIntervalSince1970: TimeInterval(sessionExpiry))
```
sanchitmehtagit: I would change sessionExpiresAt to return Date so that caller do not explicitly have to convert it to date same as expiresIn

Auth0/CredentialsManager.swift
```diff
@@ -1001,6 +1046,46 @@ public struct CredentialsManager: Sendable {
         return expiresIn < Date()
     }
 
+    /// Negative clock-skew leeway (seconds) applied when checking the `session_expiry` ceiling, so the
```
sanchitmehtagit: The value is positive and is added to the current time - the effect is that expiry fires 30 seconds early. I would suggest it to update it to  

> 
>   /// Grace period (seconds): the session is treated as expired this many seconds before the
>   /// wall-clock ceiling, ensuring it is never extended past the upstream IdP boundary.

Auth0/Credentials.swift
NandanPrabhu: thanks for the suggestion. Updated it

Auth0/CredentialsManager.swift
NandanPrabhu: resolved

Auth0/CredentialsManager.swift
NandanPrabhu: modified the comment

EXAMPLES.md
NandanPrabhu: done

sanchitmehtagit (approved):

PR #1214, opened 2026-06-22
chore: add versioned API documentation deployment

subhankarmaiti:
## Changes

- Adds CI to build versioned DocC documentation and deploy it to GitHub Pages with a version switcher. The default version is served at the canonical version-less URL, with each version also available under `/v<version>/`.
- Stops tracking the generated `docs/` directory (now published to the `gh-pages` branch instead).

No changes to the SDK or its public API.


.github/workflows/release.yml
```diff
@@ -0,0 +1,106 @@
...
+permissions:
+  contents: write
+
+env:
+  ruby: '3.3.1'
+  xcode: '16.1'
+
+jobs:
+  release:
+    name: Tag, publish to CocoaPods, and create GitHub Release
```
sanchitmehtagit: Lets raise a separate PR for Cocoapods publish

.github/workflows/docs.yml
```diff
@@ -0,0 +1,92 @@
...
+          ruby: ${{ env.ruby }}
+          xcode: ${{ env.xcode }}
+
+      - name: Bootstrap Carthage dependencies
+        run: carthage bootstrap --use-xcframeworks --no-use-binaries --cache-builds
+
+      # Produces the freshly built static site into docs_build/ with a
+      # version-scoped hosting base path (/Auth0.swift/v<version>/).
+      - name: Build documentation
+        run: bundle exec fastlane build_docs
```
sanchitmehtagit, line 48: since docs build step is automated Lets remove it from ship command that we run locally here https://github.com/auth0/Auth0.swift/blob/5d778270b0ec93bf0b33bc973d9813f28ec96965/.shiprc#L8

sanchitmehtagit (approved):

PR #1230, opened 2026-07-01
fix: update retention policy documentation and tests for version management

subhankarmaiti:
Rewrites the `DocsVersions` retention policy so a live next-major preview (e.g. `3.0.0-beta.2`) is no longer dropped when a lower-major stable ships.
                                                                                                                                                         
  **Policy (two major lines kept):**                                             
  - **Latest major** → its latest release. A stable and its own beta never coexist, so promoting a beta to stable drops the beta; a major with only betas keeps its latest beta.                                                                                                                                         
  - **Previous major** → its latest stable.                                                                                                              
  - Old versions of the current major are always dropped on release.                                                                                     
  - Prerelease releases additionally keep the latest stable below them, so the stable users rely on stays live during a preview.                            
                                                                                                                                                                                               
  ## References                                                                                                                                          
                                                                                 
  Regression from the versioned-docs rollout: the `2.24.0` stable deploy pruned the
  live `v3.0.0-beta.2` docs because the old stable branch discarded all prereleases.

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

* **Documentation**
  * Clarified the release retention policy for documentation builds, with more explicit rules for stable and prerelease versions.

* **Bug Fixes**
  * Updated version selection so the current release is always kept, along with the previous major’s latest stable release.
  * Improved prerelease handling to retain the latest stable version below the current preview, ensuring older-major content remains available.
  * Adjusted stable-release behavior to prefer the newest stable version within a major, while preserving the most relevant version from the next most recent major.

<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/scripts/DocsVersions/Tests/DocsVersionsTests/DocsVersionsTests.swift
+++ b/scripts/DocsVersions/Tests/DocsVersionsTests/DocsVersionsTests.swift
@@ -49,28 +49,71 @@ final class RetentionTests: XCTestCase {
         raw.compactMap(SemVer.init)
     }
 
-    func testStableKeepsTwoMostRecentMajors() {
+    func testStableKeepsPreviousMajorStable() {
+        // Current major (3) → current release; previous major (2) → its latest
+        // stable. Older majors (1) fall outside the two-line window.
         let keep = Retention.versionsToKeep(
-            current: SemVer("3.1.0")!,
-            existing: versions("3.0.0", "2.22.0", "2.21.2", "1.36.0")
+            current: SemVer("3.0.0")!,
+            existing: versions("2.22.0", "2.21.2", "1.36.0")
         )
-        XCTAssertEqual(keep.map { $0.description }, ["3.1.0", "2.22.0"])
+        XCTAssertEqual(keep.map { $0.description }, ["3.0.0", "2.22.0"])
     }
 
-    func testStableUsesHighestPatchPerMajor() {
+    func testStableDropsOldCurrentMajorVersions() {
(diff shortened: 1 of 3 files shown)


scripts/DocsVersions/Tests/DocsVersionsTests/DocsVersionsTests.swift
```diff
@@ -49,28 +49,71 @@ final class RetentionTests: XCTestCase {
...
         )
-        XCTAssertEqual(keep.map { $0.description }, ["2.22.0", "1.36.0"])
+        XCTAssertEqual(keep.map { $0.description }, ["1.0.0"])
     }
 
     func testPrereleaseKeepsCurrentPlusEarlierStableMajor() {
         let keep = Retention.versionsToKeep(
             current: SemVer("3.0.0-beta.2")!,
-            existing: versions("3.0.0-beta.1", "2.22.0", "2.21.2")
+            existing: versions("3.0.0-beta.1", "2.23.0", "2.22.0")
```
NandanPrabhu, line 114: since the latest is 2.24.0 should we update this?

sanchitmehtagit (approved):

scripts/DocsVersions/Tests/DocsVersionsTests/DocsVersionsTests.swift
subhankarmaiti: its a test to validate the scenario. and needs not to be true with the latest version

PR #1237, opened 2026-07-03
fix: make ActClaim conform to Sendable instead of @unchecked Sendable

NandanPrabhu:
## 📋 Changes

`ActClaim` was declared `@unchecked Sendable` because its `additionalClaims` property was typed `[String: Any]`, and `Any` isn't `Sendable`. This PR removes the escape hatch by typing `additionalClaims` (and the `init?(json:)` parameter) as `[String: any Sendable]`, so `ActClaim` gets real, compiler-checked `Sendable` conformance instead of an unchecked one.

- `Auth0/ActClaim.swift`: `ActClaim` now conforms to `Sendable` (not `@unchecked Sendable`); `additionalClaims` and `init?(json:)` now use `[String: any Sendable]`.
- `Auth0/UserProfile.swift`: updated the `act` claim cast at the `ActClaim(json:)` call site to `[String: any Sendable]` to match the new initializer signature.

`UserProfile.customClaims` remains `[String: Any]?` (and `UserProfile` remains `@unchecked Sendable`) — out of scope for this fix.

## 📎 References

N/A

## 🎯 Testing

- `swift build` succeeds with no errors.
- `swift test` — full suite passes, no failures.
- `swiftlint lint` — no new violations introduced by this change.

### DIFF
--- a/Auth0/ActClaim.swift
+++ b/Auth0/ActClaim.swift
@@ -23,7 +23,7 @@ import Foundation
 ///
 /// - [RFC 8693: OAuth 2.0 Token Exchange - act Claim](https://tools.ietf.org/html/rfc8693#section-4.4)
 /// - [Custom Token Exchange Documentation](https://auth0.com/docs/authenticate/custom-token-exchange)
-public final class ActClaim: @unchecked Sendable {
+public final class ActClaim: Sendable {
 
     /// The subject identifier of the acting party.
     ///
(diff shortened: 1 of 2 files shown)


NandanPrabhu:
Closing this PR as we need to relook on mark few types to Sendable instead of @unchecked Sendable. This PR covers only the ActClaim but we have other types too

PR #1249, opened 2026-07-13
fix: use clearAll() instead of clear() in CredentialsManager error paths

NandanPrabhu:
## Summary
- Replaces `clear()` with `clearAll()` in `CredentialsManager` error/cleanup paths (DPoP key missing, DPoP key mismatch, and expired-session handling) so all stored credentials are removed, not just the default set.

## Test plan
- [x] `swift test`
- [x] `swiftlint lint`

## Summary by CodeRabbit

* **Bug Fixes**
  * Authentication data is now fully cleared when stored credentials are invalid, DPoP keys are missing or mismatched, or the session has expired.
  * Authentication renewal and credential exchanges now report more specific errors, making failures easier to identify.
  * Improved handling prevents stale authentication state from interfering with future sign-ins.

### DIFF
--- a/Auth0.xcodeproj/project.pbxproj
+++ b/Auth0.xcodeproj/project.pbxproj
@@ -633,61 +633,11 @@
 		D41DED3B2DCA09B200F5B1A4 /* SimpleKeychain.xcframework in Frameworks */ = {isa = PBXBuildFile; fileRef = 5CD9FC8526FE30EB009C2B27 /* SimpleKeychain.xcframework */; };
 		D41DED3D2DCA09B400F5B1A4 /* JWTDecode.xcframework in Frameworks */ = {isa = PBXBuildFile; fileRef = 5CD9FC8426FE30EB009C2B27 /* JWTDecode.xcframework */; };
 		D41DED402DCA09B500F5B1A4 /* SimpleKeychain.xcframework in Frameworks */ = {isa = PBXBuildFile; fileRef = 5CD9FC8526FE30EB009C2B27 /* SimpleKeychain.xcframework */; };
-		D424ED5F2F15FEFD00052186 /* Auth0MFAClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = D424ED5C2F15FEFD00052186 /* Auth0MFAClient.swift */; };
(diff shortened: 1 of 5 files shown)


Auth0/CredentialsManager.swift
```diff
@@ -801,15 +801,15 @@ public struct CredentialsManager: Sendable {
         let hasKeyPair = try? dpop.hasKeypair()
 
         guard hasKeyPair == true else {
-            try self.clear()
+            try self.clearAll()
```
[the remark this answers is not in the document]
NandanPrabhu: we want to capture error from all paths including clearAll

Auth0.xcodeproj/project.pbxproj
```diff
@@ -2881,7 +2749,6 @@
 				5C354C05276CE1A500ADBC86 /* PasswordlessType.swift in Sources */,
 				5CFB82512D5BF324009FD237 /* APICredentials.swift in Sources */,
 				5C3D88262DC1509800AACC34 /* LoginPasskey.swift in Sources */,
-				D424ED6D2F15FF5200052186 /* Authenticator.swift in Sources */,
```
NandanPrabhu, line 2884: Files under MFA folder had double references. Hence remove the second reference

subhankarmaiti (dismissed):

pmathew92 (approved):

Auth0/CredentialsManager.swift
```diff
@@ -1011,7 +1011,7 @@ public struct CredentialsManager: Sendable {
                 return callback(.failure(error))
             } catch {
                 complete()
-                return callback(.failure(CredentialsManagerError(code: .noCredentials, cause: error)))
```
sanchitmehtagit, line 1014: why do we need this change?

sanchitmehtagit (approved):
LGTM

Auth0/CredentialsManager.swift
NandanPrabhu: There are few calls in this function and few others which can throw error on failure and the error is caught inside the catch block. If we return the nocredentials error with a cause which is the error caught by in the catch block. It wouldn't make sense because .nocredentials error shouldn't have underlying cause as per documentation. Hence changing it to the error expected to be returned by the method. so operation error with cause capturing why exactly the operation failed is what we are trying to do.

PR #1250, opened 2026-07-14
fix: CredentialsManager error paths/DPoP cleanup, MFA file duplication, and safariProvider() stuck transaction

NandanPrabhu:
## Summary
- **CredentialsManager clearAll() in error paths:** DPoP key mismatch/missing and expired-session cleanup now call `clearAll()` instead of `clear()`, so all stored credentials are removed, not just the default set.
- **Fixes double reference of MFA support files** in the Xcode project (duplicate build-file entries in `project.pbxproj`).
- **Dedicated CredentialsManager error codes:** `retrieveCredentialsWithRetry`, `retrieveSSOCredentials`, and `retrieveAPICredentials` now map failures to `.renewFailed`/`.ssoExchangeFailed`/`.apiExchangeFailed` instead of the generic `.noCredentials`; DPoP cleanup uses `try?` so a failed `clearAll()` doesn't mask the original DPoP error.
- **DPoPSpyStorage test double** now implements `deleteAllEntries()` — without it, `CredentialsManagerDPoPSpec` crashed the test process instead of failing the individual test (the default protocol extension implementation is an `assertionFailure()`).
- **`safariProvider()` stuck-transaction fix:** `SafariUserAgent.finish(with:)` deferred its dismiss/callback logic into a `Task` with a weak `self` capture, but `LoginTransaction` drops its only strong reference to the agent immediately after calling `finish(with:)` and before that `Task` runs — so the agent is deallocated and the callback that lowers the Web Auth barrier never fires, permanently stuck at `WebAuthError.transactionActiveAlready` after one login/dismiss cycle. Fixed by removing the weak capture so the `Task` keeps the agent alive until the callback actually runs.

The `V3_MIGRATION_GUIDE.md` corrections originally included here have been moved to a separate PR.

## Test plan
- [x] Full `Auth0.iOS` unit test suite (1242 tests, 0 failures), including `SafariProviderSpec` and `CredentialsManagerSpec`/`CredentialsManagerDPoPSpec`
- [x] `swiftlint lint` — no new violations
- [x] On-device repro on a physical iPhone — repeated login → dismiss → immediate re-login via `safariProvider()` now succeeds every time, with no `transactionActiveAlready`

### DIFF
--- a/Auth0.xcodeproj/project.pbxproj
+++ b/Auth0.xcodeproj/project.pbxproj
@@ -633,61 +633,11 @@
 		D41DED3B2DCA09B200F5B1A4 /* SimpleKeychain.xcframework in Frameworks */ = {isa = PBXBuildFile; fileRef = 5CD9FC8526FE30EB009C2B27 /* SimpleKeychain.xcframework */; };
 		D41DED3D2DCA09B400F5B1A4 /* JWTDecode.xcframework in Frameworks */ = {isa = PBXBuildFile; fileRef = 5CD9FC8426FE30EB009C2B27 /* JWTDecode.xcframework */; };
 		D41DED402DCA09B500F5B1A4 /* SimpleKeychain.xcframework in Frameworks */ = {isa = PBXBuildFile; fileRef = 5CD9FC8526FE30EB009C2B27 /* SimpleKeychain.xcframework */; };
-		D424ED5F2F15FEFD00052186 /* Auth0MFAClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = D424ED5C2F15FEFD00052186 /* Auth0MFAClient.swift */; };
-		D424ED602F15FEFD00052186 /* MFAClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = D424ED5D2F15FEFD00052186 /* MFAClient.swift */; };
-		D424ED612F15FEFD00052186 /* Auth0MFAClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = D424ED5C2F15FEFD00052186 /* Auth0MFAClient.swift */; };
-		D424ED622F15FEFD00052186 /* MFAClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = D424ED5D2F15FEFD00052186 /* MFAClient.swift */; };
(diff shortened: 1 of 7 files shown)


Auth0.xcodeproj/project.pbxproj
```diff
@@ -633,61 +633,11 @@
 		D41DED3B2DCA09B200F5B1A4 /* SimpleKeychain.xcframework in Frameworks */ = {isa = PBXBuildFile; fileRef = 5CD9FC8526FE30EB009C2B27 /* SimpleKeychain.xcframework */; };
 		D41DED3D2DCA09B400F5B1A4 /* JWTDecode.xcframework in Frameworks */ = {isa = PBXBuildFile; fileRef = 5CD9FC8426FE30EB009C2B27 /* JWTDecode.xcframework */; };
 		D41DED402DCA09B500F5B1A4 /* SimpleKeychain.xcframework in Frameworks */ = {isa = PBXBuildFile; fileRef = 5CD9FC8526FE30EB009C2B27 /* SimpleKeychain.xcframework */; };
-		D424ED5F2F15FEFD00052186 /* Auth0MFAClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = D424ED5C2F15FEFD00052186 /* Auth0MFAClient.swift */; };
```
sanchitmehtagit, line 636: Just wondering - how did these files get duplicated? Was it during the merge to master?

V3_MIGRATION_GUIDE.md
```diff
@@ -42,6 +42,7 @@ Auth0.swift v3 is a Swift 6-ready release with improved error handling, predicta
   + [WebAuthError cases](#webautherror-cases)
   + [Renamed APIs](#renamed-apis)
   + [New required methods on the Authentication protocol](#new-required-methods-on-the-authentication-protocol)
+  + [New required methods on the MyAccountAuthenticationMethods protocol](#new-required-methods-on-the-myaccountauthenticationmethods-protocol)
```
sanchitmehtagit: nit: Lets have separate PR for migration guide updates. I'd like to take a look at it separately

Auth0.xcodeproj/project.pbxproj
NandanPrabhu: Could be. Files are not duplicated but references

sanchitmehtagit (dismissed):
LGTM

sanchitmehtagit (approved):

PR #1285, opened 2026-09-15
feat: add EmbeddedAuth discovery client for GET /e/discovery [EA]

NandanPrabhu:
- [x] All new/changed/fixed functionality is covered by tests
- [x] I have added documentation for all new/changed functionality

### 📋 Changes

Adds an **Embedded Login Discovery** client that wraps `GET /e/discovery`, returning the live set of login alternatives available for an application so integrators can render the right options at runtime instead of hardcoding them. This flag is in Early Access (EA).

**New public API (`Auth0/EmbeddedAuth/`):**

- `Auth0.embeddedAuth(clientId:domain:session:)` and `Auth0.embeddedAuth(session:bundle:)` factories (the latter reads `Auth0.plist`).
- `EmbeddedAuth` protocol with `discover(connection:)` and a `discover()` convenience, returning `Request<DiscoveryResult, EmbeddedAuthError>` — so it gets the callback, `async throws`, and Combine surfaces for free.
- `DiscoveryResult` with `options: [LoginOption]` plus projections (`types`, `passwordRealms`, `passkeyConnections`, `otpOptions`, `socialProviders`, `supports(_:)`).
- `LoginOption` enum (password, passwordRealm, passkey, passwordlessOtp, nativeSocial, embeddedAuthorize, and a forward-compatible `unknown`) with supporting `GrantType`, `PasswordlessIdentifier`, `OTPFlowType`, `OTPOption`.
- `EmbeddedAuthError` conforming to `Auth0APIError`, with `isFeatureDisabled` / `isInvalidRequest` / `isInvalidClient` helpers.
- `EmbeddedAuthResult<T>` type alias.

The request carries the `Auth0-Client` telemetry header and honours the `tracking(enabled:)` opt-out via the existing `Request` / `Auth0ClientInfo` mechanism.

This PR also registers the `EmbeddedAuth` sources in the Xcode project (framework targets) and documents the feature in `README.md` and `EXAMPLES.md`.

### 🎯 Testing

- `Auth0Tests/EmbeddedAuth/EmbeddedAuthTests.swift` (27 cases) and `EmbeddedAuthErrorTests.swift` (13 cases) cover decoding of every grant type, the `unknown` forward-compat path, query-param construction (`client_id`, optional `connection`), result projections, and error mapping for 400/401/500/bare-404 — all via `NetworkStub`.
- The Auth0 target compiles cleanly with the new files. Note: the full `swift test` suite currently aborts locally at an unrelated, pre-existing force-unwrap in `WebAuthSpec.swift:361` (`Bundle.main.bundleIdentifier!` is `nil` under the SPM test runner); this is independent of these changes.


Auth0/EmbeddedAuth/Auth0EmbeddedAuth.swift
```diff
@@ -0,0 +1,115 @@
...
+    var loginOption: LoginOption {
+        switch grantType {
+        case "authorization_code":
+            guard let conn = connection else { return .unknown(rawGrantType: grantType, connection: nil) }
+            return .embeddedAuthorize(connection: conn)
+        case "urn:ietf:params:oauth:grant-type:token-exchange":
+            return .nativeSocial(subjectTokenType: subjectTokenType ?? "")
+        case "password":
+            return .password
+        case "urn:okta:params:oauth:grant-type:webauthn":
```
sanchitmehtagit, line 94: "urn:okta:params:oauth:grant-type:webauthn", "http://auth0.com/oauth/grant-type/passwordless/otp"
 I would avoid raw string in switch cases

Auth0/EmbeddedAuth/Auth0EmbeddedAuth.swift
```diff
@@ -0,0 +1,115 @@
...
+        case realm
+    }
+
+    var loginOption: LoginOption {
+        switch grantType {
+        case "authorization_code":
+            guard let conn = connection else { return .unknown(rawGrantType: grantType, connection: nil) }
+            return .embeddedAuthorize(connection: conn)
+        case "urn:ietf:params:oauth:grant-type:token-exchange":
+            return .nativeSocial(subjectTokenType: subjectTokenType ?? "")
```
sanchitmehtagit, line 91: I believe we may need to  add fallback handling when subjectTokenType is nil instead of passing "" empty string here.

Auth0/EmbeddedAuth/Auth0EmbeddedAuth.swift
```diff
@@ -0,0 +1,115 @@
...
+                case "email":        return .email
+                case "phone_number": return .phoneNumber
+                default:             return nil
+                }
+            }
+            return .passwordlessOtp(connection: connection ?? "",
+                                    identifiers: identifiers,
+                                    type: type == "auth0" ? .auth0 : .legacy)
+        case "http://auth0.com/oauth/grant-type/password-realm":
+            return .passwordRealm(realm: realm ?? "")
```
sanchitmehtagit, line 109: if realm is nil should we pass empty string or add some handling for it

Auth0/EmbeddedAuth/Auth0EmbeddedAuth.swift
```diff
@@ -0,0 +1,115 @@
...
+        } catch {
+            callback(.failure(EmbeddedAuthError(from: response)))
+        }
+    }
+}
+
+// MARK: - Wire types (Decodable)
+
+private struct DiscoveryResponse: Decodable {
+    let alternatives: [DiscoveryAlternativePayload]
```
pmathew92: The naming is different from what is proposed in the design doc.  Lets keep the naming uniform across all the platforms

Auth0/EmbeddedAuth/Auth0EmbeddedAuth.swift
```diff
@@ -0,0 +1,115 @@
...
+        case identifierTypes = "identifier_types"
+        case realm
+    }
+
+    var loginOption: LoginOption {
+        switch grantType {
+        case "authorization_code":
+            guard let conn = connection else { return .unknown(rawGrantType: grantType, connection: nil) }
+            return .embeddedAuthorize(connection: conn)
+        case "urn:ietf:params:oauth:grant-type:token-exchange":
```
pmathew92, line 90: subjectTokenType is a mandatory value returned for `token-exchange` grant type. So no need to put a fallback

Auth0/EmbeddedAuth/Auth0EmbeddedAuth.swift
pmathew92, line 109: This will have a mandatory value for realm. So fallback value is not required

Auth0/EmbeddedAuth/DiscoveryResult.swift
```diff
@@ -0,0 +1,157 @@
...
+        case .embeddedAuthorize:   return .authorizationCode
+        case .unknown:             return .unknown
+        }
+    }
+
+    /// The connection name associated with this alternative, if any.
+    public var connection: String? {
+        switch self {
+        case .password, .nativeSocial:                          return nil
+        case .passwordRealm:                                    return nil
```
pmathew92, line 110: .passwordRealm will have the connection value. You can return the realm of this grant type for this

Auth0/EmbeddedAuth/DiscoveryResult.swift
```diff
@@ -0,0 +1,157 @@
...
+    case passkey(connection: String)
+
+    /// Passwordless OTP — either `legacy` (via `/passwordless/start`) or `auth0` (via `/otp/challenge`).
+    case passwordlessOtp(connection: String, identifiers: [PasswordlessIdentifier], type: OTPFlowType)
+
+    /// Native social login via token exchange (Google, Apple, Facebook).
+    case nativeSocial(subjectTokenType: String)
+
+    /// Interactive embedded authorization via `POST /e/authorize`.
+    case embeddedAuthorize(connection: String)
```
pmathew92, line 86: Lets name it authorizationCode since thats what the grant type is . In future we will have different types coming in this grant_type other than `embedded_authorize`

Auth0/EmbeddedAuth/DiscoveryResult.swift
```diff
@@ -0,0 +1,157 @@
...
+    case email
+    /// Phone number.
+    case phoneNumber
+}
+
+/// Which passwordless flow a connection uses, determining the follow-up API call.
+///
+/// - `legacy`: Use `POST /passwordless/start` → verify with realm + otp.
+/// - `auth0`:  Use `POST /otp/challenge` → verify with auth_session + otp.
+public enum OTPFlowType: Sendable, Equatable {
```
pmathew92, line 144: Lets name this PasswordlessType  to align with what we have in Android / Design doc

Auth0/EmbeddedAuth/DiscoveryResult.swift
```diff
@@ -0,0 +1,157 @@
...
+            guard case .nativeSocial(let stt) = $0 else { return nil }
+            return stt
+        }
+    }
+
+    /// Returns `true` if any alternative matches the given grant type.
+    public func supports(_ grantType: GrantType) -> Bool {
+        options.contains { $0.grantType == grantType }
+    }
+
```
pmathew92, line 62: Lets also add a helper called `hasEmbeddedAuthorization` which checks if embedded authorization is enabled or not. Check the design doc

pmathew92 (changes requested):

Auth0/EmbeddedAuth/Auth0EmbeddedAuth.swift
NandanPrabhu: made it similar to android. if connection not present or any other check, by default accessing loginOption property will return into nil value

Auth0/EmbeddedAuth/DiscoveryResult.swift
NandanPrabhu: added helper below

Auth0/EmbeddedAuth/Auth0EmbeddedAuth.swift
NandanPrabhu: handled it same as android

Auth0/EmbeddedAuth/Auth0EmbeddedAuth.swift
NandanPrabhu: handled same as android

NandanPrabhu: login option would now return nil aligned with android

Auth0/EmbeddedAuth/DiscoveryResult.swift
NandanPrabhu: done

Auth0/EmbeddedAuth/DiscoveryResult.swift
NandanPrabhu: done

Auth0/EmbeddedAuth/DiscoveryResult.swift
NandanPrabhu: addressed

Auth0/EmbeddedAuth/Auth0EmbeddedAuth.swift
```diff
@@ -0,0 +1,101 @@
...
+                       parameters: params,
+                       logger: logger,
+                       auth0ClientInfo: auth0ClientInfo)
+    }
+
+}
+
+// MARK: - Wire types (Decodable)
+
+struct DiscoveryResponse: Decodable {
```
pmathew92, line 40: should this be marked as private ?

Auth0/EmbeddedAuth/Auth0EmbeddedAuth.swift
```diff
@@ -0,0 +1,101 @@
...
+    let alternatives: [DiscoveryAlternativePayload]
+}
+
+struct DiscoveryAlternativePayload: Decodable {
```
pmathew92, line 44: Should this be marked as private ?

NandanPrabhu: by default this is internal. So, host app/sdk won't have access to this code. marking it private would result into error because other files inside the SDK wont be able to access it

Auth0/EmbeddedAuth/Auth0EmbeddedAuth.swift
NandanPrabhu: by default this is internal. So, host app/sdk won't have access to this code. marking it private would result into error because other files inside the SDK wont be able to access it

pmathew92 (approved):

